PR #6, opened 2017-12-20
Counter badge

rokasmik:


lbelinsk:
Wait, I am reconsidering this one. Why did we call it CounterBadge?
In order to answer this, we would probably need a product to tell us whether we can have in the settings panel some slider that can affect the border-radius.
For now, we will leave this component.
@shirlev-wix question: Do you know if we have a case where we would control the Badge border-radius?

shirlev-wix:
I don't know... I never had to use a badge as a TPA.

PR #15, opened 2018-04-11
Add dropdown styles; fix chevron

ihork:


src/components/Autocomplete/Autocomplete.st.css
```diff
@@ -8,6 +8,11 @@
     -st-default: CoreAutocomplete;
 }
 
+:import {
```
shlomitc: this is not needed

src/components/Autocomplete/Autocomplete.st.css
```diff
@@ -46,6 +51,19 @@
     align-items: center;
 }
 
+DropdownContent::dropdownOption {
```
shlomitc: replace with `.root::dropdownContent`

why it works:
`.root` extends `CoreAutocomplete`
`CoreAutocomplete` extends `InputWithOptions` 
`InputWithOptions` extends `Dropdown`
`Dropdown` exposes `dropdownContent` class which overrides the dropdown content

why your solution also work:
it's like targeting a tag name. it works because there is a dropdown, but theoretically, the implementation can change and not be the dropdown from core, but something else. maintaining the CoreAutocomplete API preserves the behavior.

src/components/Autocomplete/Autocomplete.st.css
```diff
@@ -46,6 +51,19 @@
...
+    height: 42px;
```
shlomitc: Item height should be 39px according to up-to-date design

src/components/Autocomplete/Autocomplete.st.css
```diff
@@ -46,6 +51,19 @@
...
+    box-sizing: border-box;
+    padding: 10px 14px;
+    font-size: 16px;
```
shlomitc: This should be mapped to a predefined font - `Body-M`.

src/components/Autocomplete/Autocomplete.st.css
```diff
@@ -46,6 +51,19 @@
...
+    color: #212121;
+}
+
+DropdownContent::dropdownOption:hovered {
+    background-color: #f4f4f4;
```
shlomitc: hovering should apply opactity of 60%

src/components/Autocomplete/Autocomplete.st.css
```diff
@@ -46,6 +51,19 @@
     align-items: center;
 }
 
+DropdownContent::dropdownOption {
+    height: 42px;
+    box-sizing: border-box;
+    padding: 10px 14px;
+    font-size: 16px;
+    color: #212121;
```
shlomitc: color should also be mapped like Input color

src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -43,7 +43,12 @@ export const Autocomplete: AutocompleteType = ((props: AutocompleteProps) => {
           empty={errorProps.empty}
           suffix={
             <span className={style.suffix}>
-            {<ChevronDown className={style.arrowIcon}/>}
+            {<ChevronDown
```
shlomitc: the icon color should be different when hovered/selected and when not

shlomitc:
image

There is a grey border that is left there when the input is focused.
It's probably the input's focus css that needs to be overridden by the autocomplete focus state

ihork:
@shlomitc I think your screenshot is not up-to-date
screen shot 2018-04-13 at 10 32 48 am

src/components/Autocomplete/Autocomplete.tsx
ihork: I know, that in Zeplin it visually looks different, but actually it uses the same color states as text inside of input. And according to latest changes, text keeps it's color on hover state. Anyway I sent the question to Doron to be sure

PR #27, opened 2018-11-04
changed all dist import drivers to the new index file

RashelleKosib:


src/components/Input/Input.protractor.driver.ts
```diff
@@ -1 +1 @@
-export * from 'wix-ui-core/dist/src/components/Input/Input.protractor.driver';
+export * from 'wix-ui-core/drivers/protractor';
```
shlomitc: This looks incorrect. you should export only the Input's content

RashelleKosib: fixed

PR #35, opened 2019-01-14
Text Component with override capabilities

liorMar:


src/components/Text/Text.driver.ts
```diff
@@ -0,0 +1,13 @@
+import {StylableDOMUtil} from '@stylable/dom-test-kit';
+import style from './Text.st.css';
+
+export const textDriverFactory = ({element}) => {
+  const stylableDOMUtil = new StylableDOMUtil(style);
+
+  return {
+    exists: () => !!element,
+    getTagName: () => element.tagName.toLowerCase(),
+    getContent: () => element.innerHTML,
```
shlomitc: I think `textContent` is better

liorMar: @shlomitc done

PR #41, opened 2019-01-16
Button component

liorMar:
added overridable button component

### DIFF
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -2,7 +2,6 @@ import * as React from 'react';
 import style from './Button.st.css';
 import {ButtonNext} from 'wix-ui-core/button-next';
 import {ButtonProps as ButtonNextProps} from 'wix-ui-core/dist/src/components/button-next/button-next';
-// import { withEllipsedTooltip } from 'wix-ui-core/dist/src/hocs/EllipsedTooltip';
 
 export enum PRIORITY {
   basic = 'basic',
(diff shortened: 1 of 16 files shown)


src/components/Button/Button.tsx
```diff
@@ -0,0 +1,24 @@
...
+import {ButtonProps as ButtonNextProps} from 'wix-ui-core/dist/src/components/button-next/button-next';
+// import { withEllipsedTooltip } from 'wix-ui-core/dist/src/hocs/EllipsedTooltip';
+
+export enum PRIORITY {
+  primary = 'primary',
+  secondary = 'secondary'
+}
+
+export interface ButtonProps extends ButtonNextProps {
+  dataHook?: string;
```
liorMar: because of ButtonNext and it's UniDriver i had to add the dataHook
is there any better solution?

shlomitc: why is that? this should not be a mandatory property

liorMar: https://github.com/wix/wix-ui/blob/master/packages/wix-ui-test-utils/src/vanilla/vanilla.tsx#L59
because of this line i get props error in Button

liorMar: @shlomitc after new install of node modules i stopped getting the error. i'll delete it

src/components/Button/Button.tsx
```diff
@@ -0,0 +1,32 @@
...
+  medium = 'medium',
+  large = 'large'
+}
+
+export interface ButtonProps extends ButtonNextProps {
+  priority?: PRIORITY;
+  size?: SIZE;
+}
+
+const CoreButton = ({priority, size, dataHook, ...rest}) =>
```
liorMar: @shlomitc i can catch the dataHook prop without declaring it on the api but i don't like this solution.
because the next dev' that will look at it will be puzzled

PR #50, opened 2019-03-20
added text typography - running text 2

poseneror:
Hey, please review and merge, shouldn't be any different than other typos

### DIFF
--- a/src/components/Text/Text.spec.tsx
+++ b/src/components/Text/Text.spec.tsx
@@ -43,9 +43,9 @@ describe('Text', () => {
     expectTextToHaveAttributes(expectedTag, expectedTypography);
   });
 
-  it('should render runningText2', () => {
+  it('should render listText', () => {
     const expectedTag = 'p';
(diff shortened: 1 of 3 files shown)


poseneror:
Heading M is Paragraph 2, if you think something else should be used I could manage that :)
We have a different task for mobile styles, so it will probably be defined in a later stage.

johannaasseraf:
Lets connect the style of Heading M to H2 of the text themes. The default design should override the text size though

poseneror:
sorry, I meant that body M is connected to Paragraph 2.
Heading-M is Heading-5 from settings

johannaasseraf:
Can we connect it to H2 (just not get the h2 tag for a11y)

poseneror:
Yeah, I think it should be set to 'Page-title'

johannaasseraf:
Page title isn't H1 in the template text themes?
image

poseneror:
There's another preset 'title' which is the H1 preset

PR #55, opened 2019-04-04
Card component

liorMar:


liorMar:
e2e failing, converting to unidriver

src/components/Card/Card.driver.ts
```diff
@@ -0,0 +1,22 @@
...
+  const stylableDOMUtil = new StylableDOMUtil(style);
+
+  const getImageContainerElement = () =>
+    element.querySelector(stylableDOMUtil.scopeSelector('.mediaContainer'));
+  const getInfoContainerElement = () =>
+    element.querySelector(stylableDOMUtil.scopeSelector('.infoContainer'));
+
+  return {
+    exists: () => !!element,
+    getMediaContent: () => getImageContainerElement().innerHTML,
```
shlomitc: instead of returning the html - return the dom node.
if you're migrating to unidriver - it should return the `getNative` element

stories/Card.story.tsx
```diff
@@ -0,0 +1,77 @@
...
+//         { title: 'API', sections: [api()] },
+//         { title: 'TestKit', sections: [testkit()] },
+//         { title: 'Playground', sections: [playground()] },
+//       ].map(tab),
+//     ]),
+//     mdx({ content: MDX }),
+//   ]
+// };
+
+import * as React from 'react';
```
shlomitc: let's move all of the docs to `src/components/Card/docs` folder

src/components/Card/Card.st.css
```diff
@@ -0,0 +1,127 @@
...
+:vars {
+    /*Constance*/
+    DefaultCardBG: color-1;
+    DefaultBorderColor: color-5;
+    DefaultBorderWidth: 1;
+}
+
+:vars {
+    /*Overrides*/
+    MainBorderWidth: --overidable;
```
shlomitc: overridable

src/components/Card/Card.tsx
```diff
@@ -0,0 +1,42 @@
...
+  RATIO_40_60 = 'ratio-40',
+  RATIO_30_70 = 'ratio-30',
+}
+
+export interface CardProps {
+  media?: React.ReactNode;
+  info?: React.ReactNode;
+  ratio?: CardRatioOptions;
+  flippedRatio?: boolean;
+  invertInfoPosition?: boolean;
```
shlomitc: what's the difference in the terminology of `flip` and `invert`? if they do the same, I would just pick one

src/components/Card/index.tsx
```diff
@@ -0,0 +1,2 @@
+export { Card, CardProps, CardRatioOptions } from './Card';
+export { OverlappingCard } from './OverlappingCard';
```
shlomitc: we don't expose more than one component per folder.
please create a separate folder and docs for it

src/components/Card/mixins.st.css
```diff
@@ -0,0 +1,16 @@
...
+}
+
+.withBorder {
+    border-style: solid;
+    border-color: value(BorderColor);
+    border-width: value(BorderWidth);
+}
+
+.withDynamicPadding {
+    padding: calc(value(PaddingSize) - value(BorderWidth));
```
shlomitc: just be aware that this is a runtime `calc` function and not a static one.
if you do want a static one, you need to use a mixin like this
https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/mixins/calc/index.js

src/components/Card/README.md
```diff
@@ -0,0 +1,41 @@
...
+            MainMediaColor '"--imageColor"',
+            MainBorderColor '"--borderColor"'
+            MainBorderWidth '"--borderWidth"'
+      );
+    }
+
+    `
+
+2. Create a component that uses it
+    ` javascript
```
shlomitc: it should be a triple backtick <code>```</code>

stories/Card.story.tsx
```diff
@@ -0,0 +1,77 @@
...
+import {Examples} from './Card';
+import {Card} from '../src/components/Card';
+
```
shlomitc:
```suggestion
const mediaExamples = [
  { label: 'image', value: <img src="something"/>},
  { label: 'video', value: <video src=""/> },
];
```

stories/Card.story.tsx
```diff
@@ -0,0 +1,77 @@
...
+export default {
+  category: 'Components',
+  storyName: 'Card',
+  component: Card,
+  componentPath: '../src/components/Card/Card.tsx',
+  componentProps: () => ({
+    'data-hook': 'storybook-Card',
+  }),
```
shlomitc:
```suggestion
  }),
  exampleProps: {
    media: mediaExamples,
  }
```

stories/Card.story.tsx
```diff
@@ -0,0 +1,77 @@
...
+import * as React from 'react';
+import {Examples} from './Card';
+import {Card} from '../src/components/Card';
+
+export default {
+  category: 'Components',
+  storyName: 'Card',
+  component: Card,
+  componentPath: '../src/components/Card/Card.tsx',
+  componentProps: () => ({
```
shlomitc:
```suggestion
  componentProps: () => ({
  media: mediaExamples[0].value
```

stories/Card/examples.ts
```diff
@@ -0,0 +1,93 @@
...
+        <Text typography={TYPOGRAPHY.runningText}>Lorem ipsum dolor sit amet.</Text>
+      </div>
+      <div style={{marginTop: '20px'}}>
+        <Button >Book</Button>
+      </div>
+    </div>
+  }/>
+`;
+
+export const overlapping = `
```
shlomitc: when splitting folder of components, split this example as well

stories/Card.story.tsx
shlomitc: same should be done for the `info` prop

src/components/Card/Card.tsx
liorMar: flippedRatio for example switch between 30%-70% to 70%-30%
invertInfoPosition switch the position of info and image positions

src/components/Card/mixins.st.css
liorMar: yeah i'm a ware of it i need the padding to be responsive to the border width.
the border width is dynamically changes in the settings panel and the should
correspond to it

src/components/Card/README.md
liorMar: yeah, i'm trying to convert to mdx for the section api with no success so far

src/components/Card/index.tsx
liorMar: @shlomitc i want to simplify the import of *.st.css files from another .st.css files
is exporting the style object from here is ok
so devs could just use `:import {
        -st-from: "wix-ui-tpa/Card";
        -st-named: CardStyle;
    }`
instead of `:import {
  -st-from: "wix-ui-tpa/src/components/Button/Button.st.css";
  -st-default: TPAButton;
}`

shlomitc: I'll make it possible in a different PR, but if there is more than one exported component - it should have a separate folder entry

shlomitc: https://github.com/wix/wix-ui-tpa/pull/56

PR #64, opened 2019-04-24
button improvement

liorMar:
until now the border grow outside the button area and made the button dimensions bigger than the design system specs
added support to full width (required for mobile view)

### DIFF
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -19,19 +19,22 @@ export enum SIZE {
 export interface ButtonProps extends ButtonNextProps {
   priority?: PRIORITY;
   size?: SIZE;
+  fullWidth?: boolean;
 }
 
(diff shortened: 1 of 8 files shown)


johannaasseraf:
great 👏🏻

src/components/Button/Button.tsx
```diff
@@ -19,19 +19,22 @@ export enum SIZE {
...
+    {...style('root', { priority, size, fullWidth }, rest)}
+  />
+);
 
 Button.displayName = 'Button';
 
 Button.defaultProps = {
   priority: PRIORITY.basic,
   size: SIZE.medium,
+  fullWidth: false,
```
shlomitc: add a simulation in the story to `fullWidth: false`, so it will raise this flag from `Misc. Props
` to `Primary Props`

PR #66, opened 2019-04-29
rtl ignore to tabs alignment

noakorkus:
Hi Shlomi,
We want to disable the alignment if we are in rtl , since the alignment is controlled by the user from settings.


jonathanadler:
So just to understand a bit more, can you elaborate more on the use case?
I'm guessing the user has an option to define the layout's alignment: `ltr`/`center`/`rtl`.
And do you pass the `alignment` prop to the component?

PR #69, opened 2019-05-02
mobile design for Button, Tabs, Text, Card

liorMar:


jonathanadler:
@liorMar I've noticed the styling of the text in the `<Button />` and in the `<Card />` don't appear as one of the `<Text />` component's presets, and I was wondering if maybe they should be, and then they would use the `<Text />` component internally, and that way we won't need to define mobile text in different components. What do you think?

src/components/Card/docs/index.story.tsx
```diff
@@ -77,6 +81,7 @@ export default {
           ...[
             { title: 'Side By Side', source: examples.sideBySide },
             { title: 'Stacked', source: examples.stacked },
+            { title: 'Stacked On Mobile', source: examples.stackedMobile },
             { title: 'Ratios', source: examples.ratio },
             { title: 'Flipped Ratios', source: examples.flippedRatio },
             {
```
jonathanadler: Do we need an example for `invertInfoPosition` and `mobile` maybe?

src/components/Text/Text.st.css
```diff
@@ -11,6 +11,10 @@
     RunningTextFont: "{theme: 'Body-M', size: '16px', lineHeight: '1.5em'}";
     ListTextFont: "{theme: 'Body-M', size: '16px', lineHeight: '2em'}";
     LargeTitleFont: "{theme: 'Heading-M', size: '32px', lineHeight: '1.25em'}";
+    MobileSmallTitleFont: "{theme: 'Page-title', size: '20px', lineHeight: '1.4em'}";
```
jonathanadler: Are these variables necessary? I think it might add unnecessary complexity. Users could still override the font using the `:mobile` state in their `.st.css` files.

src/components/Button/Button.st.css
```diff
@@ -21,6 +21,7 @@
     SmallFont: "{theme: 'Body-M', size: '16px', lineHeight: '1.5em'}";
     MediumFont: "{theme: 'Body-M', size: '16px', lineHeight: '1.5em'}";
     LargeFont: "{theme: 'Body-M', size: '20px', lineHeight: '1.5em'}";
+    MobileMediumFont: "{theme: 'Body-M', size: '14px', lineHeight: '1.44em'}";
```
jonathanadler: See [`Text.st.css`](https://github.com/wix/wix-ui-tpa/pull/69/files#diff-0fd7d140afb61156fd3669c414db9edaR14)

src/components/Tabs/Tabs.st.css
```diff
@@ -7,6 +7,7 @@
     /*Constance*/
     defaultTabColor: color-5;
     defaultTabFont: "{theme: 'Body-M', size: '16px'}";
+    defaultMobileTabFont: "{theme: 'Body-M', size: '14px'}";
```
jonathanadler: See [`Text.st.css`](https://github.com/wix/wix-ui-tpa/pull/69/files#diff-0fd7d140afb61156fd3669c414db9edaR14)

src/components/Card/docs/index.story.tsx
liorMar: by the design system we should not allowed invertInfoPosition on mobile

src/components/Text/Text.st.css
liorMar: this variables represent the design system specs. users can override everything in there `.st.css` with or without the `mixin`

liorMar:
> @liorMar I've noticed the styling of the text in the `<Button />` and in the `<Card />` don't appear as one of the `<Text />` component's presets, and I was wondering if maybe they should be, and then they would use the `<Text />` component internally, and that way we won't need to define mobile text in different components. What do you think?
my problem with that is it leaves an open for users to use other presets for `<Text/>` that are not intent. for example use the `button.st.css` `TinyFont` preset in `<Tabs/>`

liorMar:
> In general, I think we prefer adding a `<Provider />` component which will accept `isMobile` and `isRtl` props (or `mobile` and `rtl` props). Then each component that will need to use it will just consume the props. This way users of the library will define it once in the provider in the app's root.

@jonathanadler 
what do you think about deviceType('desktop' | 'mobile' | 'tablet') instead of mobile? it is more aligned
with the ooi props

src/components/Text/Text.st.css
jonathanadler: The way I see it, practically speaking, the style variables give the developers a way to change the binding of Wix's styleParams to their own settings panel. Mobile sites don't have a settings panel at the moment, and so most of the developers won't need to use these variable.
Theoretically speaking, the text is the same text element, and so the variable should stay the same. 
If we'll see that it's missing for developers in the future I guess we could add it, but for now it seems redundant.
It might seem a bit petty, but I think it will give a clearer api for developers.
@shlomitc I'd love to hear what you think

jonathanadler:
> > @liorMar I've noticed the styling of the text in the `<Button />` and in the `<Card />` don't appear as one of the `<Text />` component's presets, and I was wondering if maybe they should be, and then they would use the `<Text />` component internally, and that way we won't need to define mobile text in different components. What do you think?
> 
> my problem with that is it leaves an open for users to use other presets for `<Text/>` that are not intent. for example use the `button.st.css` `TinyFont` preset in there own text.
> we can still do something like use `<Text/>` in `<Button/>` and do a `mixin` override with fallback
> ` -st-mixin: TPAText( MainTextColor '"fallback(userOverrideTextColor, mobilePresetColor)"', MainTextFont '"fallback(userOverrideTextFont, mobilePresetFont)"' );` in `button.st.css`

So yeah, my question is: should there be really special text styles for these components? Similar to the `API first` paradigm.
If the answer is yes, there should be a special case for them, then I guess it's ok

jonathanadler:
> > In general, I think we prefer adding a `<Provider />` component which will accept `isMobile` and `isRtl` props (or `mobile` and `rtl` props). Then each component that will need to use it will just consume the props. This way users of the library will define it once in the provider in the app's root.
> 
> @jonathanadler
> what do you think about deviceType('desktop' | 'mobile' | 'tablet') instead of mobile? it is more aligned
> with the ooi props

Good question 🤔 
So first, `Tablet` isn't supported by wix sites, so that will leave us with `desktop | mobile`. Unless you use your own logic for this, which I think is a specific use case.
Second, we're thinking of aligning to the future `responsive` editor.
In this editor we wont have any distinction between device types, and all tpa's and components will have one endpoint. In this case, future components will never have a state of `mobile:true` and everything will work out of the box, or at least most will work out of the box.

src/components/WixUiTpaConfig/WixUiTpaConfig.tsx
```diff
@@ -0,0 +1,44 @@
+import * as React from 'react';
```
shlomitc: This is how I see the wanted api:

```jsx
import {Provider} from 'wix-ui-tpa/Provider';
import {Card} from 'wix-ui-tpa/Card';

const App = ({WixSdk}) => (
  <Provider mobile={WixSdk.isMobile()}>
     <Card/>
  </Provider>
)
```

The `mobile` prop should propagate internally to the `Card` component and shouldn't be a public API.

src/components/WixUiTpaConfig/WixUiTpaConfig.tsx
```diff
@@ -0,0 +1,44 @@
...
+          const deviceTypeProps = {
+            mobile: context.mobile,
+          };
+          return <WrappedComponent {...props} {...deviceTypeProps} />;
+        }}
+      </WixUiTpaConfigConsumer>
+    );
+  };
+
+export const WixUiTpaConfigWrapper = ({mobile = false}) => {
```
shlomitc: This is the responsibility of the consumer, not the library.
if it's for test purposes, put it in the tests folder

src/components/WixUiTpaConfig/WixUiTpaConfig.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+export interface WixUiTpaConfig {
+  mobile: boolean;
+}
+
+export interface WixUiTpaConfigProps {
+  mobile?: boolean;
+}
+
+const wixUiTpaConfig = React.createContext<WixUiTpaConfig>({
```
shlomitc: why name it config and not context?

src/components/WixUiTpaConfig/WixUiTpaConfig.tsx
```diff
@@ -0,0 +1,44 @@
...
+  mobile: false,
+});
+
+export const WixUiTpaConfigProvider = wixUiTpaConfig.Provider;
+
+export const WixUiTpaConfigConsumer = wixUiTpaConfig.Consumer;
+
+export const withWixUiTpaConfig = <TProps extends object>(
```
shlomitc: I don't like the extra wrapper. I prefer components just to consume the context consumer

src/components/WixUiTpaConfig/WixUiTpaConfig.tsx
liorMar: @shlomitc  in 'src/test/utils.tsx' ?

shlomitc: yes, can be a good start

src/components/WixUiTpaConfig/WixUiTpaConfig.tsx
liorMar: @shlomitc the use of the Consumer Should be like `class Button extends React.Component<ButtonProps> {
  render() {
    const {
      priority,
      size,
      fullWidth,
      ...rest
    } = this.props;
    return (<WixUiTpaContextConsumer>
      {({mobile}) => <ButtonNext
        {...rest}
        {...style('root', { priority, size, fullWidth, mobile }, rest)}
      />}
    </WixUiTpaContextConsumer>)
  }
}`?

PR #71, opened 2019-05-06
<Tooltip /> - Adding TPA styles tooltip

giladartzi:


src/components/Tooltip/Arrows.spec.tsx
```diff
@@ -0,0 +1,24 @@
+import { top, bottom, left, right, getArrowByPlacement } from './Arrows';
+
+describe('Tooltip - Custom arrows', () => {
+  test.each`
```
jonathanadler: Nice!

jonathanadler:
We're only missing an `eyes.it` e2e test.
And on the way, we'll need to externalize the `wix-ui-core/tooltip` driver for this component.

PR #74, opened 2019-05-07
disable mobile tap effect on Chrome and Safari

liorMar:


liorMar:
it was required by Avital
https://jira.wixpress.com/browse/SCHED-9881
but wait if you feel more comfortable

jonathanadler:
This is fixed in #82

PR #76, opened 2019-05-08
Generator update

liorMar:
update the generator according to Card component


generator/templates/src/components/Component/Component.spec.tsx
```diff
@@ -1,40 +1,50 @@
...
   });
 
   describe('enzyme testkit', () => {
-    it('should exist', () => {
+    it('should exist', async () => {
       expect(
-        isEnzymeTestkitExists(<{%ComponentName%} />, enzyme{%ComponentName%}TestkitFactory, mount, {
-          dataHookPropName: 'data-hook',
-        }),
+        await isUniEnzymeTestkitExists(
```
argshook: just fyi, eventually `should exist` tests will also be automated with wuf. so for now it's ok to add them to each component but the goal is to automate it

it is already automated in wsr, need to migrate and generalize the script

so it's just fyi, it'll happen some time :)

generator/templates/src/components/Component/docs/index.story.tsx
```diff
@@ -0,0 +1,59 @@
...
+  divider,
+  importExample,
+  playground,
+  tab,
+  code as baseCode,
+  tabs,
+  testkit,
+  title,
+} from 'wix-storybook-utils/Sections';
+import { baseScope } from '../../../../stories/utils/LiveCodeExample';
```
argshook:
```suggestion
import { LiveScope } from '../../../../stories/utils/allComponents';
```
no reason to proxy it through LiveCodeExample. it was an ad-hoc implementation in wsr

also, fyi, in future `code` section will not require to set `components` each time. it will be setup just once.

PR #78, opened 2019-05-12
Text color of runningText & listText to color 5

liorMar:


shlomitc:
@liorMar - why was this change required?
I want us to learn from this case and others because that API changes like these will not be easily possible in the future. It's a breaking change and can break user apps

PR #77, opened 2019-05-12
fix unknown baseScope in story generator

liorMar:
fix unknown baseScope

PR #79, opened 2019-05-13
Chore/livecodeexample remove

argshook:
in response to https://github.com/wix/wix-ui-tpa/pull/77

---

* remove LiveCodeExample
* use `allComponents` instead of `LiveScope` from LiveCodeExample
* fix generator template to use `allComponents`

---

Before `code` section there was LiveCodeExample. `code` section is the successor, hence, LiveCodeExample is no longer needed.

wix-ui-tpa includes a copy-paste of LiveCodeExample from wix-style-react. Eventually we'll get rid of LiveCodeExample from wix-style-react too.

---

in addition, the `allComponents` and `baseCode` things will eventually disappear. `code` section should just work, but until then, we still need to pass reference of all components to `code` section.


argshook:
@liorMar merging, change is harmless. let me know if you think otherwise :)

PR #63, opened 2019-04-21
Stacked Card layout

liorMar:
Stacked Card layout

PR #81, opened 2019-05-14
Strip Card Component

liorMar:


shlomitc:
@jonathanadler - please review it when you're back from Guild Week

liorMar:
@jonathanadler do not merge.

src/components/StripCard/StripCard.spec.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const createDriver = createUniDriverFactory(stripCardDriverFactory);
+
+  it('should render', async () => {
+    const driver = createDriver(<StripCard />);
+    expect(await driver.exists()).toBe(true);
+  });
+
+  it('should render info', async () => {
+    const valueAsString = '<div>info!</div>';
+    const value = <div dangerouslySetInnerHTML={{ __html: valueAsString }} />;
```
jonathanadler: Why use `dangerouslySetInnerHTML`? I'm not sure I understand

src/components/StripCard/StripCard.driver.ts
```diff
@@ -0,0 +1,31 @@
+import {
+  BaseUniDriver,
+  baseUniDriverFactory,
+} from 'wix-ui-test-utils/base-driver';
+import { StylableUnidriverUtil, UniDriver } from 'wix-ui-test-utils/unidriver';
+import style from './StripCard.st.css';
+
+export interface stripCardDriver extends BaseUniDriver {
```
jonathanadler: Super picky, I know, but lets stay with the convention of `PascalCase` for an `interface`'s name.
Meaning `StripCardDriver` instead of `stripCardDriver`

src/components/StripCard/StripCard.tsx
```diff
@@ -0,0 +1,44 @@
...
+export interface StripCardProps {
+  media?: React.ReactNode;
+  info?: React.ReactNode;
+  roundMedia?: boolean;
+  withoutSidePadding?: boolean;
+}
+
+interface DefaultProps {
+  roundMedia: boolean;
+  withoutSidePadding: boolean;
```
jonathanadler: I think it's better to avoid double negatives, and have this prop be `sidePadding`, and have a default value of `true`. I believe It's a bit more intuitive.

src/components/StripCard/StripCard.spec.tsx
```diff
@@ -0,0 +1,69 @@
...
+    const driver = createDriver(<StripCard info={value} />);
+    expect((await driver.getInfoContent()).innerHTML).toEqual(valueAsString);
+  });
+
+  it('should render image', async () => {
+    const valueAsString = '<div>image!</div>';
+    const value = <div dangerouslySetInnerHTML={{ __html: valueAsString }} />;
```
jonathanadler: > Why use dangerouslySetInnerHTML? I'm not sure I understand

Same here

src/components/StripCard/StripCard.tsx
```diff
@@ -0,0 +1,44 @@
...
+}
+
+export class StripCard extends React.Component<StripCardProps> {
+  static displayName = 'StripCard';
+  static defaultProps: DefaultProps = {
+    roundMedia: false,
+    withoutSidePadding: false,
```
jonathanadler: > I think it's better to avoid double negatives, and have this prop be sidePadding, and have a default value of true. I believe It's a bit more intuitive.

`sidePadding: true`

src/components/StripCard/StripCard.tsx
```diff
@@ -0,0 +1,44 @@
...
+      <TPAComponentsConsumer>
+        {({ mobile }) => (
+          <div
+            {...styles(
+              'root',
+              { mobile, roundMedia, withoutSidePadding },
+              rest,
+            )}
+          >
+            {media && <div className={styles.mediaContainer}>{media}</div>}
```
jonathanadler: I prefer using a ternary operator:
```
{media ? <div className={styles.mediaContainer}>{media}</div> : null}
```
because in case `media` is faulty (`0` for instance) `{media && <div className={styles.mediaContainer}>{media}</div>}` will return `0` and not `<div class="mediaContainer">0</div>`

src/components/StripCard/docs/index.story.tsx
```diff
@@ -0,0 +1,95 @@
...
+        style={{
+          background:
+            'url("https://static.wixstatic.com/media/dd7e03deffdd4267b75ef0a60d8510c7.jpg/v1/fill/w_100,h_100,al_c,q_80,usm_0.66_1.00_0.01/Gift%20Wrapped.jpg") no-repeat center center',
+          height: '100%',
+          backgroundSize: 'cover',
+        }}
+      />
+    ),
+  },
+  { label: 'video', value: <video src="" /> },
```
jonathanadler: How about using a real video here?
https://interactive-examples.mdn.mozilla.net/media/examples/flower.webm for instance?

src/components/StripCard/docs/index.story.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+          divider(),
+
+          title('Examples'),
+
+          ...[
+            { title: 'Strip', source: examples.strip },
+            { title: 'Minimum Height', source: examples.minHeight },
+            { title: 'Without Media', source: examples.stripOnlyInfo },
+            { title: 'Media Not Available', source: examples.noImageLoaded },
```
jonathanadler: I'm not sure what's the purpose of the `noImageLoaded` example

src/components/StripCard/StripCard.spec.tsx
liorMar: Shlomi didn't want us to have innerHTML in the driver and it was the only solution we found
https://github.com/wix/wix-ui-tpa/pull/63/files#diff-f3d7262b9300d4ce27455808474b8da2L27

src/components/StripCard/StripCard.driver.ts
liorMar: it is auto generated so i'm updating the generator too

src/components/StripCard/StripCard.tsx
liorMar: this ternary is still vulnerable to `0`

src/components/StripCard/StripCard.st.css
```diff
@@ -0,0 +1,86 @@
...
+            BorderWidth value(borderWidth)
+    );
+    display: flex;
+    background-color: value(cardBGColor);
+    min-height: value(CardMinHeight);
+    min-width: value(MinWidth);
+    box-sizing: border-box;
+}
+
+.root:mobile {
```
jonathanadler: If there isn't a special design for mobile, then this can be discarded along with the `TPAComponentsConsumer`

src/components/StripCard/StripCard.tsx
jonathanadler: Yes, my example wasn't that good.
But the difference is that with the ternary operator you decide what to return when it is falsy, and it is more readable.

liorMar:
> @shlomitc ?
I see that in line 35 you do use:
>
>`const driver = createDriver(
  <StripCard roundMedia media={<div>image</div>} />,
);`

@jonathanadler because on that test we are not checking the media content but if it has roundMedia attribute.

src/components/StripCard/docs/index.story.tsx
liorMar: to see that instead of the media or while the media is loaded we display the Media Background color

jonathanadler:
Looks great!

Just one small thing. Look at the examples here:
https://wix-wix-ui-tpa-pr-81.surge.sh/?activeTab=Usage&path=%2Fstory%2Fcomponents--stripcard

I think we should maybe add a wrapping div with a background color, or something like that, to the examples, because it's hard to understand where the component ends. 

What do you think?

@shlomitc I think it's ready to merge, unless you want to go over it

PR #84, opened 2019-05-21
don't add "px" to border radius

eranshabi:
I bumped major because this is a breaking change for current button users.
we already have the `px` part saved in the styleParams so we can't have it added in the library too which will result invalid css like `2pxpx`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "private": false,
   "name": "wix-ui-tpa",
-  "version": "1.0.0",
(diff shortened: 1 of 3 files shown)


jonathanadler:
I think you also need to fix the examples. 
Currently it's using a number value for the border radius, and the example breaks.
See here (at the bottom):
https://wix-wix-ui-tpa-pr-84.surge.sh/?path=/story/components--button

PR #85, opened 2019-05-22
make buttonNext button element ref exposed

eranshabi:


eranshabi:
fixed

eranshabi:
> @eranshabi
> What's the use case for exposing the ref? what do you need it for?
> Do you really need the reference to the element, or is there maybe an api missing?

I want to focus on the button

jonathanadler:
> > @eranshabi
> > What's the use case for exposing the ref? what do you need it for?
> > Do you really need the reference to the element, or is there maybe an api missing?
> 
> I want to focus on the button

Is `autoFocus` enough or do you need a special case?

eranshabi:
> > > @eranshabi
> > > What's the use case for exposing the ref? what do you need it for?
> > > Do you really need the reference to the element, or is there maybe an api missing?
> > 
> > 
> > I want to focus on the button
> 
> Is `autoFocus` enough or do you need a special case?

I need to manually focus for accessibility reasons (a popup opens and when it closes the button should be focused)

jonathanadler:
> > > > @eranshabi
> > > > What's the use case for exposing the ref? what do you need it for?
> > > > Do you really need the reference to the element, or is there maybe an api missing?
> > > 
> > > 
> > > I want to focus on the button
> > 
> > 
> > Is `autoFocus` enough or do you need a special case?
> 
> I need to manually focus for accessibility reasons (a popup opens and when it closes the button should be focused)

I see. So Lets use [React.forwardRef](https://reactjs.org/docs/forwarding-refs.html), instead of having `this.buttonRef.buttonRef.focus()` we can have `this.buttonRef.focus()`

jonathanadler:
@eranshabi 
If you don’t want to add the forward ref, you can also just pass a ref from the parent to the child Button:

```
render() {
    const { buttonRef } = this.props;

    return <Button ref={buttonRef} 
      ...
}
```

eranshabi:
fixed (used forwardRef)

src/components/Button/Button.spec.tsx
```diff
@@ -30,6 +30,12 @@ describe('Button', () => {
     expect(await driver.isMobile()).toEqual(true);
   });
 
+  it('should let focus on inner button using ref', async () => {
+    const ref = React.createRef<HTMLButtonElement>();
+    const driver = createDriver(<Button fullWidth ref={ref} />);
+    expect(ref.current.focus).toBeInstanceOf(Function);
```
jonathanadler: instead of checking the `focus` function, lets see that the `ref.current` is an instance of the button-next

eranshabi: > @eranshabi instead of checking the `focus` function, lets see that the `ref.current` is an instance of the button-next

good comment, I changed the test and found out that the `focus` method that I get is actually of `focusableHOC`.
I did PR to fix this in wix-ui-core:
https://github.com/wix/wix-ui/pull/1321

PR #86, opened 2019-05-23
button: don't add "px" to border width

eranshabi:
Like this PR but for `border-width`:
https://github.com/wix/wix-ui-tpa/pull/84

### DIFF
--- a/src/components/Button/Button.st.css
+++ b/src/components/Button/Button.st.css
@@ -36,7 +36,7 @@
 
 :vars {
     PrimaryTextColor: wixApply(color, wixApply(fallback, value(MainTextColor), value(SecondaryColor)));
-    PrimaryBorderWidth: wixApply(unit, wixApply(fallback, value(MainBorderWidth), 0), px);
+    PrimaryBorderWidth: wixApply(fallback, value(MainBorderWidth), 0);
 }
 
 .root {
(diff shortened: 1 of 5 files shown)


src/components/Button/Button.st.css
```diff
@@ -110,7 +110,7 @@
     color: value(SecondaryTextColor);
     background-color: wixApply(color, wixApply(fallback, value(MainBackgroundColor), transparent));
     border-color: value(SecondaryBorderColor);
-    border-width: wixApply(unit, wixApply(fallback, wixApply(zeroAsTrue, value(MainBorderWidth)), 1), px);
+    border-width: wixApply(fallback, wixApply(zeroAsTrue, value(MainBorderWidth)), 1);
```
jonathanadler: the `1` should be with units: `1px` or `wixApply(unit, 1, px)`

stories/Button/ButtonExtendedWithStyleParamsExample.st.css
```diff
@@ -0,0 +1,20 @@
...
+  -st-default: TPAButton;
+}
+
+.root {
+  -st-extends: TPAButton;
+  -st-mixin: TPAButton(
+          MainTextColor '"--buttonTextColor"',
+          MainBackgroundColor '"--buttonBackgroundColor"',
+          MainTextFont '"--buttonTextFont"',
+          MainBorderWidth 'unit(fallback(--borderWidth, 1), px)',
```
jonathanadler: I think the default value should be 0, like in the slider

eranshabi:
fixed

jonathanadler:
> Are there any other places that have units that need to be removed? maybe we can close them all with one pr

@eranshabi 
Lets see if there are any other places we need to remove the `px` unit, and try to combine them all to one pr

eranshabi:
> > Are there any other places that have units that need to be removed? maybe we can close them all with one pr
> 
> @eranshabi
> Lets see if there are any other places we need to remove the `px` unit, and try to combine them all to one pr

that would be all for button

PR #87, opened 2019-05-28
fix StripCard border color always apply opacity on TPA overrides

liorMar:
should not break eyes, default border width is 0

### DIFF
--- a/src/components/StripCard/StripCard.st.css
+++ b/src/components/StripCard/StripCard.st.css
@@ -34,7 +34,7 @@
 :vars {
     cardBGColor: wixApply(color, wixApply(fallback, value(MainBGColor), value(DefaultCardBG)));
     borderWidth: wixApply(unit, wixApply(fallback, wixApply(zeroAsTrue, value(MainBorderWidth)), wixApply(zeroAsTrue, value(DefaultBorderWidth))), px);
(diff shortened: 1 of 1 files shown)


src/components/StripCard/StripCard.st.css
```diff
@@ -34,7 +34,7 @@
 :vars {
     cardBGColor: wixApply(color, wixApply(fallback, value(MainBGColor), value(DefaultCardBG)));
     borderWidth: wixApply(unit, wixApply(fallback, wixApply(zeroAsTrue, value(MainBorderWidth)), wixApply(zeroAsTrue, value(DefaultBorderWidth))), px);
-    borderColor: wixApply(opacity, wixApply(color, wixApply(fallback, value(MainBorderColor), value(DefaultBorderColor))), 0.2);
+    borderColor: wixApply(color, wixApply(fallback, value(MainBorderColor), wixApply(opacity, wixApply(color, value(DefaultBorderColor)), 0.2)));
```
jonathanadler: Just to be sure, it's not a problem to wrap `color` around another `color`, right? I mean, it's recursive

liorMar: no problems

PR #88, opened 2019-05-28
<Grid /> Component to align items in grid layout

liorMar:


src/components/Grid/Grid.tsx
```diff
@@ -0,0 +1,169 @@
...
+    dividerWidth: '1px',
+  };
+
+  render() {
+    return (
+      <TPAComponentsConsumer>
+        {({ mobile }) => {
+          const { withDivider, width, items, ...rest } = this.props;
+          const rowGap =
+            this.props.rowGap || (mobile ? MOBILE_ROW_GAP : ROW_GAP);
```
jonathanadler: Correct me if i'm wrong, but this means I can't set a `rowGap` of `0`, right?

src/components/Grid/Grid.tsx
```diff
@@ -0,0 +1,169 @@
...
+          const columnGap =
+            this.props.columnGap || (mobile ? MOBILE_COLUMN_GAP : COLUMN_GAP);
```
jonathanadler: Same as above

src/components/Grid/GridUtils.tsx
```diff
@@ -0,0 +1,61 @@
...
+  let mediaQueries = '';
+  while (maxColumns > 0) {
+    const minWidth = maxColumns * minColumnWidth + columnGap * (maxColumns - 1);
+    mediaQueries = `@media (min-width: ${minWidth}px) {#${gridId} .${ListItemClass} {-ms-grid-columns: repeat(${maxColumns}, minmax(${minColumnWidth}px, 100vw));grid-template-columns: repeat(${maxColumns}, minmax(${minColumnWidth}px, 100vw));}}${mediaQueries}`;
+    maxColumns--;
+  }
+  return <style dangerouslySetInnerHTML={{ __html: mediaQueries }} />;
+}
+
+export function getMinWidthByCardType(CardComponent): number {
```
jonathanadler: Is this really necessary? 
I think it adds some "noise" to this component. 
You will obviously know which items you're passing to this component, so you can use the `minColumnWidth` prop.
And other users might want to use cards with different min width, so it seems very specific.

src/components/Grid/Grid.tsx
```diff
@@ -0,0 +1,169 @@
...
+            styles.$cssStates({ dividers: true }),
+          )[0];
+          return (
+            <div
+              {...styles('root', { dividers: withDivider }, rest)}
+              id={gridId}
+            >
+              <style
+                dangerouslySetInnerHTML={{
+                  __html: `
```
jonathanadler: I would also move this to `GridUtils` or maybe a `gridStyleTemplates` file.
It's a bit too much information for one render function IMO.

src/components/Grid/Grid.e2e.ts
```diff
@@ -0,0 +1,25 @@
...
+import { browser } from 'protractor';
+import {
+  createStoryUrl,
+  waitForVisibilityOf,
+} from 'wix-ui-test-utils/protractor';
+import { gridTestkitFactory } from '../../testkit/protractor';
+
+describe('grid', () => {
+  const storyUrl = createStoryUrl({
+    kind: 'Components',
```
jonathanadler: We want this to be under a new category of `Bookings` components.
So I guess you can change the `kind` to `Bookings/Components` or just `Bookings`.
We're not really familiar with the entire spec of this component, and we're not sure this should be part of the library. However, we don't want to obstruct your development, and we prefer having the component here, so others might want to use it, but under a different category.

src/components/Grid/Grid.tsx
liorMar: correct

src/components/Grid/Grid.e2e.ts
```diff
@@ -0,0 +1,25 @@
...
+    story: 'Grid',
+    withExamples: true,
+  });
+  const dataHook = 'storybook-Grid';
+
+  beforeEach(() => browser.get(storyUrl));
+
+  eyes.it('should render', async () => {
+    const driver = gridTestkitFactory({ dataHook });
+    debugger;
```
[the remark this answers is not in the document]
liorMar: removing

src/components/Grid/GridUtils.tsx
liorMar: we removed the min-width from the cards. we still need to enforce the the min-width.
is exporting the min-width from each card and ask users to enforce it in the MD is better solution?

liorMar:
> Regarding the Card's styling change, i'm not really sure why the changes, but it seems it didn't break the eyes tests, so I guess it's not a problem.

when stacked Card is part of grid with different size info the the image is taking more space to grow in height to fit the size of other cards. the change is simple telling the media to not grow more than you need but let the info do the growing

> Also, I took a look at the storybook, I wasn't really sure how responsiveness should work. The responsive example isn't quite responsive.
In addition, i'm not sure how it should work inside the editor. What if users want to stretch and shrink a widget with this component? How will it behave?

true, the props was incorrect but ignore the weird preview width: 50%
 if the user not setting the widget as fullWidth the it is not responsive by editor definition.

src/components/Grid/Grid.spec.tsx
```diff
@@ -0,0 +1,301 @@
...
+  });
+
+  it('should use 1 default amount of items ', async () => {
+    const driver = createDriver(<Grid items={generateItems()} width={1} />);
+
+    expect(await driver.isItemsPerRow(1)).toBe(true);
+  });
+
+  it('should use default max item width', async () => {
+    const expectedItemMaxWidth = '100vw';
```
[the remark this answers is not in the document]
liorMar: 100vw of the iframe or the window getting the same result because we still limited by the container
but after checking the result of 100vw it is taking as much space that it can instead of what it need. so changing to 100%

src/components/Grid/Grid.e2e.ts
liorMar: putting under Bookings

src/components/Grid/Grid.tsx
liorMar: created `getGridStyle` function in `GridUtils` 
and did a little refactoring of merging style tags to one tag

src/components/Grid/GridUtils.tsx
jonathanadler: But this is logic that can be inside of your app. I'm not sure this should be part of the component.
The component can accept a `minWidth` prop, and have the default `DEFAULT_MIN_WIDTH`.

liorMar: removed

src/components/Grid/Grid.driver.ts
```diff
@@ -0,0 +1,62 @@
...
+}
+
+export const gridDriverFactory = (base: UniDriver): gridDriver => {
+  const stylableUtil = new StylableUnidriverUtil(style);
+
+  return {
+    ...baseUniDriverFactory(base),
+    isItemsPerRow: async itemsPerRow => {
+      return (await base.$('ul').attr('style')).includes(
+        `repeat(${itemsPerRow},`,
```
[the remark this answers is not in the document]
liorMar: @jonathanadler changed

src/components/Grid/Grid.spec.tsx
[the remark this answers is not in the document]
carmelc: We checked, it, 100% does not provide the desired behavior, as it will not really stretch to consume all the space available, as it is just 100% of the container which will not necessarily get the full width of the window/iframe, so we will leave it as 100vw

src/components/Grid/GridUtils.tsx
carmelc: We cannot remove it, there is minimum width per card defined in the design system,
A grid cannot "take" the minimum from a cell (and maintain all the other logic that we need), so if the min for the card is higher than 1fr it will just overlap on the next grid column, which is not what we want, so this is a must,
considering that the api allows to set it by the tpa, we can have the logic in Bookings, but we already discussed it with DS team, and we agreed that the grid can support non-card elements, but optimized for them, so we do need this logic in the library.

src/components/Grid/Grid.tsx
```diff
@@ -0,0 +1,226 @@
...
+
+export interface GridItem {
+  item: React.ReactNode;
+  key?: string | number;
+  colSpan?: number;
+  rowSpan?: number;
+}
+
+export interface GridProps {
+  withDivider: boolean;
```
jonathanadler: change to `showRowDivider`

src/components/Grid/Grid.tsx
```diff
@@ -0,0 +1,226 @@
...
+  autoRowHeight: boolean;
```
jonathanadler: What do you think about `fixedRowHeight` for this prop?

src/components/Grid/GridUtils.spec.tsx
```diff
@@ -0,0 +1,209 @@
...
+
+  describe('generateKey', () => {
+    it('should add prefix to the id', () => {
+      const expectedPrefix = 'list_';
+
+      expect(generateKey(expectedPrefix)).toContain(expectedPrefix);
+    });
+  });
+
+  describe('getMinWidthByCardType', () => {
```
jonathanadler: So this can be removed, right?

src/components/Grid/GridUtils.tsx
```diff
@@ -0,0 +1,104 @@
...
+  if (rowWidth !== 0) {
+    while (rowWidth < maxColumns * itemWidth + columnGap * (maxColumns - 1)) {
+      maxColumns--;
+    }
+  }
+
+  return maxColumns || 1;
+}
+
+export function getMinWidthByCardType(CardComponent): number {
```
jonathanadler: This can be removed, right?

src/components/Grid/README.md
```diff
@@ -0,0 +1,68 @@
...
+|------------|----------|--------------|-------------|
+| children  | Grid.Item[]   | [] | items to fill the Grid |
+| showRowDivider  | boolean   | false | displays divider between the rows |
+| maxColumns  | number   | 1 | max of columns per row |
+| minColumnWidth  | number   | 0 | minimum width of each column in `px` * |
+| width  | number   | 0 | the width of the grid in `px` ** |
+| rowGap  | number   | 32 | the gap between each row |
+| columnGap  | number   | 32 | the gap between each column |
+| dividerWidth | string \| number   | `1px` | the width of the divider can accept number in `px` |
+
```
jonathanadler: `uniformRowHeight` should be added

src/components/StripCard/StripCard.st.css
```diff
@@ -19,7 +19,6 @@
     DefaultCardBG: color-1;
     DefaultMediaBG: color-5;
     DefaultBorderColor: color-5;
-    MinWidth: 740px;
```
jonathanadler: Should this be added back?

src/components/OverlappingCard/OverlappingCard.st.css
```diff
@@ -48,6 +48,7 @@
     box-sizing: border-box;
     display: inline-grid;
     width: 100%;
+    height: 100%;
     grid-template-rows: 1fr 8fr 1fr; /*1fr = 10%*/
     grid-template-areas: "media margin-top" "media info" "media margin-bottom";
```
jonathanadler: should there be a `min-width` here?

liorMar: @jonathanadler fixed all the above

PR #97, opened 2019-06-06
(Card | Strip layout) allowing info side padding to be override

liorMar:
https://zeroheight.com/7sjjzhgo2/p/19342a/t/55ee2c

### DIFF
--- a/src/components/Card/README.md
+++ b/src/components/Card/README.md
@@ -12,6 +12,7 @@ By default the Card will pull the theme from the site (site colors),
 | MainMediaColor  | string   | color-5 20% opacity | The color of the image container instead of the default color |
 | MainBorderColor  | string   | color-5 | The color of the border instead of the default color |
 | MainBorderWidth  | string   | 1px | The width of the border instead of the default width |
+| MainInfoSidePadding  | string   | 32px | The width of Stack layout Side padding |
 
 
 In order to override the theme, use Wix Stylable extends capabilities and wrap with stylable HOC:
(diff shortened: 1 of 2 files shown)


jonathanadler:
@liorMar 
I don't really have a problem with adding this, but just to be sure, is this coming from design?
Is this from here:
image
?

liorMar:
> @liorMar
> I don't really have a problem with adding this, but just to be sure, is this coming from design?
> Is this from here:
> image
> ?

yes this is where it coms from

jonathanadler:
@johannaasseraf @liorMar 
Should this accept an amount or a `true/false`?

liorMar:
> @johannaasseraf @liorMar
> Should this accept an amount or a `true/false`?

after confirming with @johannaasseraf it should accept an amount

src/components/Card/README.md
```diff
@@ -12,6 +12,7 @@ By default the Card will pull the theme from the site (site colors),
 | MainMediaColor  | string   | color-5 20% opacity | The color of the image container instead of the default color |
 | MainBorderColor  | string   | color-5 | The color of the border instead of the default color |
 | MainBorderWidth  | string   | 1px | The width of the border instead of the default width |
+| MainInfoSidePadding  | string   | 32px | The width of Stack layout Side padding |
```
jonathanadler: Maybe we should call it `StackedInfoSidePadding`?
Because it might confuse users that it isn't affecting horizontal cards.
Or maybe this should apply on both types?

liorMar: @jonathanadler renamed

PR #98, opened 2019-06-06
(POC) adding SettingsPanel to storybook sections

liorMar:
adding SettingsPanel as an extra Tab


jonathanadler:
Applitools tests are failing because some of them contain the color/font pickers and sliders in the test itself.
It's something should probably move out of the test itself.
I can add it to this PR, or separate to a different PR, wdyt?

jonathanadler:
In addition, maybe I can sit with our UXers and have them design a settings panel for us?
And maybe this can be integrated into the `storybook-utils`?

stories/helperComponents/MockSettings.scss
```diff
@@ -1,57 +1,82 @@
...
-  overflow-y: auto;
-  position: absolute;
-  right: -150px;
-  top: 0;
-  bottom: 0;
-}
+      .colorPickerContainer {
+        max-width: ($colorPickerWidth + ($colorPickerMargin * 2)) * 3;
+
+        .pickerList {
```
argshook: not critical but whenever i see i must comment :)
try to avoid nesting as much as possible, css should not reflect html hierarchy. with css modules nesting is almost never required

src/components/Card/docs/index.story.tsx
```diff
@@ -100,6 +103,50 @@ export default {
...
+            settingsPanel({
+              example: <CardExtendedExample />,
+              rawSource: ExtendedRawSource,
+              rawCSSSource: ExtendedCSSRawSource,
+              params: {
+                numbers: [
+                  {
+                    label: 'Border Width',
+                    wixParam: 'borderWidth',
+                    defaultNumber: 1,
```
argshook:
```suggestion
                    default: 1,
```
it can be implied that it's number, cause it's part of numbers array. same for colors

unless, maybe it's too much work to change now?

stories/helperComponents/MockSettings.tsx
```diff
@@ -315,86 +317,106 @@ export class MockSettings extends React.PureComponent<
...
+                </li>
+              ))}
+            </ul>
+          </div>
+          <div className={styles.palettePickerContainer}>
+            <div className={styles.palettePicker}>
+              <label>Palette Picker</label>
+              <UI.ColorPickerPalettePicker
+                value={this.state.selectedPalette}
+                onChange={palette => this.onPaletteChange(palette)}
```
argshook:
```suggestion
                onChange={this.onPaletteChange}
```

argshook:
> And maybe this can be integrated into the `storybook-utils`?

@jonathanadler 	as this is specific to wix-ui-tpa only, i wouldn't want
to add it to wix-storybook-utils, it should be generic.

what we can think of is some kind of plugin thing, so that `settingsPanel` section is easy to add in
other component stories

shlomitc:
@argshook @NitayRabi @jonathanadler - adding to `wix-storybook-util` sounds great, but I would like to see a technical design of it before any development starts.
In the meanwhile, adding something specific to `wix-ui-tpa` as a POC is ok

stories/helperComponents/MockSettings.scss
jonathanadler: I respectfully disagree 😄 
In any case, this is just the style for the storybook color picker elements, so it really isn't critical.

src/components/Card/docs/index.story.tsx
jonathanadler: This is actually the APi of the `BaseUI` component that we import from `wix-base-ui`

stories/helperComponents/MockSettings.tsx
jonathanadler: totally

stories/helperComponents/MockSettings.scss
argshook: we can have a beer about it some day.

i have had a looot of wasted time on css like this, and it never proved to be worthy to deeply nest things. moreover, nesting creates overly specific selectors which are not always desired and could be a problem later on

jonathanadler:
> > And maybe this can be integrated into the `storybook-utils`?
> 
> @jonathanadler as this is specific to wix-ui-tpa only, i wouldn't want
> to add it to wix-storybook-utils, it should be generic.
> 
> what we can think of is some kind of plugin thing, so that `settingsPanel` section is easy to add in
> other component stories

This is actually something that Shlomi and I talked about.
I think it would be awesome to be able to add this as a plugin to `wix-storybook-utils`, since it's really specific, but also might be useful for @NitayRabi (and maybe others?).
I think it will also be some kind of "proof" for `wix-storybook-utils`'s maturity and effectiveness.

stories/helperComponents/MockSettings.scss
jonathanadler: 🍻

PR #99, opened 2019-06-07
(Card | OverlappingCard) Media container with multiple background colors

liorMar:
added multiple background colors to Media container

https://zeroheight.com/7sjjzhgo2/p/19342a/b/247036

### DIFF
--- a/src/components/Card/Card.st.css
+++ b/src/components/Card/Card.st.css
@@ -10,6 +10,7 @@
 
 :vars {
     /*Constance*/
+    DefaultCardBG: color-1;
     DefaultInfoCardBG: color-1;
     DefaultMediaCardBG: color-5;
     DefaultBorderColor: color-5;
@@ -22,11 +23,13 @@
     MainBorderColor: --overridable;
     MainInfoColor: --overridable;
     MainMediaColor: --overridable;
+    MainCardColor: --overridable;
 }
 
 :vars {
     borderWidth: wixApply(unit, wixApply(fallback, wixApply(zeroAsTrue, value(MainBorderWidth)), value(DefaultBorderWidth)), px);
     borderColor: wixApply(color, wixApply(fallback, value(MainBorderColor), value(DefaultBorderColor)));
+    cardBGColor: wixApply(color, wixApply(fallback, value(MainCardColor), value(DefaultCardBG)));
     infoBGColor: wixApply(color, wixApply(fallback, value(MainInfoColor), value(DefaultInfoCardBG)));
     mediaBGColor: wixApply(color, wixApply(fallback, value(MainMediaColor), wixApply(opacity, wixApply(color, value(DefaultMediaCardBG)), 0.2)));
 }
@@ -45,7 +48,7 @@
     box-sizing: border-box;
     width: 100%;
     height: 100%;
-    background-color: value(infoBGColor);
+    background-color: value(cardBGColor);
     -st-mixin: withBorder(
             BorderColor value(borderColor),
             BorderWidth value(borderWidth)
@@ -132,7 +135,7 @@
 
 .root > .infoContainer {
     box-sizing: border-box;
-    background-color: transparent;
+    background-color: value(infoBGColor);
     -st-mixin: withDynamicPadding(
             PaddingSize value(infoPadding),
             BorderWidth value(borderWidth)
--- a/src/components/Card/Card.st.css
+++ b/src/components/Card/Card.st.css
@@ -45,6 +45,7 @@
     box-sizing: border-box;
     width: 100%;
     height: 100%;
+    background-color: value(infoBGColor);
     -st-mixin: withBorder(
             BorderColor value(borderColor),
             BorderWidth value(borderWidth)
@@ -131,7 +132,7 @@
 
 .root > .infoContainer {
     box-sizing: border-box;
-    background-color: value(infoBGColor);
+    background-color: transparent;
     -st-mixin: withDynamicPadding(
             PaddingSize value(infoPadding),
             BorderWidth value(borderWidth)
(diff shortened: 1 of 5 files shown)


jonathanadler:
@liorMar 
Can you elaborate on this a bit more?
I'm not sure I understand this change.

johannaasseraf:
Each container needs it’s own background color. According to this:
https://zeroheight.com/7sjjzhgo2/p/19342a/b/247036


On 10 June 2019 at 12:07:17, Jonathan Adler (notifications@github.com)
wrote:

@liorMar <https://github.com/liorMar>
Can you elaborate on this a bit more?
I'm not sure I understand this change.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/wix/wix-ui-tpa/pull/99?email_source=notifications&email_token=AI3UXWB2GLCOVADNK23ZYRLPZYKUBA5CNFSM4HVVP4V2YY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGODXJK5TI#issuecomment-500346573>,
or mute the thread

src/components/Card/Card.st.css
```diff
@@ -131,7 +132,7 @@
 
 .root > .infoContainer {
     box-sizing: border-box;
-    background-color: value(infoBGColor);
+    background-color: transparent;
```
jonathanadler: Why was this moved to the `root`?

src/components/OverlappingCard/OverlappingCard.st.css
```diff
@@ -152,3 +146,9 @@
             BorderWidth value(borderWidth)
     );
 }
+
+.root > .mediaContainer {
+    background: linear-gradient(value(mediaBGColor), value(mediaBGColor)), linear-gradient(value(infoBGColor), value(infoBGColor));
```
jonathanadler: Where did the `linear-gradient`'s come from?

src/components/Card/Card.st.css
liorMar: @jonathanadler for example if the user  wants to have transparent media background then we need to show the info background and not whats behind the Card so the easiest solution was to move the info background to root and make the info transparent

src/components/OverlappingCard/OverlappingCard.st.css
liorMar: @jonathanadler in order to have the same behaviour as in `Card` we can't use the same solution. because of the height different. so `linear-gradient` lets us use multiple layers of background Color

liorMar:
> @liorMar
> Can you elaborate on this a bit more?
> I'm not sure I understand this change.

the main goal is to have media container base Color Background the same as of Info container

src/components/Card/Card.st.css
jonathanadler: So the info background color is the base background for the whole component?

johannaasseraf:
There’s a background color for the whole card, and a background color for
the info container but bookings will set their info container background to
0% opacity


On 10 June 2019 at 18:10:04, Jonathan Adler (notifications@github.com)
wrote:

*@jonathanadler* commented on this pull request.
------------------------------

In src/components/Card/Card.st.css
<https://github.com/wix/wix-ui-tpa/pull/99#discussion_r292053266>:

> @@ -131,7 +132,7 @@

 .root > .infoContainer {
     box-sizing: border-box;
-    background-color: value(infoBGColor);
+    background-color: transparent;

So the info background color is the base background for the whole component?

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/wix/wix-ui-tpa/pull/99?email_source=notifications&email_token=AI3UXWGEAJRYO5H2ZUNKW3TPZZVERA5CNFSM4HVVP4V2YY3PNVWWK3TUL52HS4DFWFIHK3DMKJSXC5LFON2FEZLWNFSXPKTDN5WW2ZLOORPWSZGOB3BSH6Q#discussion_r292053266>,
or mute the thread

src/components/Card/Card.st.css
liorMar: > So the info background color is the base background for the whole component?

yes, but after Johanna's comment i think that i'll add a third css override for the Card BG with default the same as Info

src/components/OverlappingCard/OverlappingCard.st.css
jonathanadler: I'm not familiar with this syntax.
Why are there 2 `linear-gradients`?
And what's the difference between this and `background-color: value(mediaBGColor);`

liorMar: you can't apply multiple layers of `background-color`  or have `background-color: value(mediaBGColor); ` and `background: linear-gradient(value(mediaBGColor), value(mediaBGColor));` at the same time.
`background` is stronger and will be shown on top of `bacgrounf-color` regardless the css order

liorMar: but `linear-gradient` after `linear-gradient` (only work with `background`) can let you apply multiple layers with order of how is on top.

jonathanadler: Why not use:
```
.media {
    background-color: fallback(value(mediaBgColor), value(cardBgColor));
}

.info {
    background-color: fallback(value(infoBgColor), value(cardBgColor));
}
```

liorMar: Because if the user selecting a Color with opacity the the `cardBgColor` will not be shown

jonathanadler: Thank you 🙏 
That was what I was missing.

PR #105, opened 2019-06-11
add StatesButton component

eranshabi:


package.json
```diff
@@ -22,7 +22,7 @@
     "!protractor.conf.js"
   ],
   "scripts": {
-    "build": "yoshi build && npm run copy-styleable-ext && npm run compile-base-ui && npm run generate-stylable-index && npm run storybook:build",
+    "build": "npm run create-icons && yoshi build && npm run copy-styleable-ext && npm run compile-base-ui && npm run generate-stylable-index && npm run storybook:build",
```
deanshub: if it's a yoshi4 project you can just import svgs as a component (using https://github.com/smooth-code/svgr)

eranshabi: it's not 😞

src/components/StatesButton/StatesButton.tsx
```diff
@@ -0,0 +1,72 @@
...
+    this.props.onClick();
+  };
+
+  public render() {
+    const { text, disabled, className } = this.props;
+    const { success } = this.state;
+    return (
+      <Button
+        data-hook="states-button"
+        className={className ? className : null}
```
jonathanadler: I think stylable deals with the className, so no need to add it.
But I might be wrong, so lets check

src/components/StatesButton/StatesButton.tsx
```diff
@@ -0,0 +1,72 @@
...
+  public state = {
+    success: false,
+  };
+
+  public focus = () => {
+    this.buttonRef.current.focus();
+  };
+
+  public async onActionSuccess() {
+    this.setState({ success: true });
```
jonathanadler: We'll have to think of a different approach.
This won't work with the "loading..." state.

src/components/StatesButton/StatesButton.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  public render() {
+    const { text, disabled, className } = this.props;
+    const { success } = this.state;
+    return (
+      <Button
+        data-hook="states-button"
+        className={className ? className : null}
+        disabled={disabled}
+        size={SIZE.large}
```
jonathanadler: Is this component always large?
I can't see this in the spec

src/components/StatesButton/StatesButton.tsx
```diff
@@ -0,0 +1,72 @@
...
+  disabled: boolean;
+  onClick: Function;
+  text: string;
+}
+
+interface StatesButtonState {
+  success: boolean;
+}
+
+export class StatesButton extends React.Component<
```
jonathanadler: We'll need to find a different name. Maybe `SuccessButton`.
I'll get back to you with this tomorrow

src/components/StatesButton/StatesButton.tsx
```diff
@@ -0,0 +1,72 @@
...
+        data-hook="states-button"
+        className={className ? className : null}
+        disabled={disabled}
+        size={SIZE.large}
+        onClick={this.onClick}
+        ref={this.buttonRef}
+        {...style('root', {}, this.props)}
+      >
+        <div
+          className={classNames(style.text, {
```
jonathanadler: We can implement this with stylable, which is preferred.
Add a `success` state to the `root`

src/components/StatesButton/StatesButton.tsx
```diff
@@ -0,0 +1,72 @@
+import * as React from 'react';
+import * as classNames from 'classnames';
+import style from './StatesButton.st.css';
+import { Button, SIZE } from '../Button';
+import { Check } from '../../icons/dist';
```
jonathanadler: I think we'll work on icons in a different PR.
Let's start by using a local adjacent file, that won't be public.
So we can also remove the svg build process.

src/components/StatesButton/StatesButton.tsx
```diff
@@ -0,0 +1,72 @@
...
+  private readonly onClick = () => {
+    this.props.onClick();
+  };
+
+  public render() {
+    const { text, disabled, className } = this.props;
+    const { success } = this.state;
+    return (
+      <Button
+        data-hook="states-button"
```
jonathanadler: The data hook should be received as a prop (this means we might actually not need a `dataHooks` file).

src/components/StatesButton/StatesButton.tsx
```diff
@@ -0,0 +1,72 @@
+import * as React from 'react';
+import * as classNames from 'classnames';
+import style from './StatesButton.st.css';
+import { Button, SIZE } from '../Button';
+import { Check } from '../../icons/dist';
+
+const delay = time => new Promise(resolve => setTimeout(resolve, time));
+
+export interface StatesButtonProps {
```
jonathanadler: Should extend the `ButtonProps`

src/components/StatesButton/StatesButton.tsx
```diff
@@ -0,0 +1,72 @@
...
+    return (
+      <Button
+        data-hook="states-button"
+        className={className ? className : null}
+        disabled={disabled}
+        size={SIZE.large}
+        onClick={this.onClick}
+        ref={this.buttonRef}
+        {...style('root', {}, this.props)}
+      >
```
jonathanadler: From what I understand from Johanna, this should work with any type of button, not just a `Large` button, and also with any type of priority.
So you should probably pass the props to the `Button` component (`{...this.props}`), and also manage the `success` state inside the `st.css` file, to override the button's background and such.

src/icons/raw/Check.svg
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jonathanadler: I recommend starting with a React element, and not an svg file (you can see the tooltip pr for reference).
That way we won't need to deal with the whole svg build phase.

jonathanadler:
My proposal for this component's API:

```jsx
const onButtonClick = async () => {
    this.setState({currentState: ProgressButtonStates.LOADING});
    await doSomethingThatTakesTime();
    this.setState({currentState: ProgressButtonStates.SUCCESS});
}

<ProgressButton currentState={this.state.currentState}
                              onClick={onButtonClick}
                              onProgressEnd={() => this.setState({currentState: ProgressButtonStates.IDLE})}>
    <ProgressButton.Idle>Click Me!</ProgressButton.Idle>
    <ProgressButton.Loading>Loading... 🤔</ProgressButton.Loading>
    {/* In addition <ProgressButton.LoadingIcon /> (but I think you won't have to develop this) */}
    <ProgressButton.Success>Success 😃</ProgressButton.Success>
    {/* In addition <ProgressButton.SuccessIcon /> */}
    {/* We'll probably also need to add 
        <ProgressButton.Error>Error 😢</ProgressButton.Error>
        and
        <ProgressButton.ErrorIcon /> 
    */}
</ProgressButton>
```

@shlomitc wdyt?

shlomitc:
@jonathanadler when `onProgressEnd` is triggered?

jonathanadler:
> @jonathanadler when `onProgressEnd` is triggered?

This is triggered 2 seconds after success is triggered.
It lets the developers the chance to switch back to `idle` state.
I know It's not that declarative, but I couldn't think of another way to enforce the 2 second behaviour.
It's just a way to guide the developers on how to use this component.

The problem with this component, is that we want it to be controlled, in order for the developer to tell it when it should be in `loading` state, but we also want it to be uncontrolled in regard to the `success` timeout.

We can also just expose a `ProgressButton.SUCCESS_TIMEOUT = 2000;` static const as a guideline, and let the developers do what they want.

src/components/StatesButton/StatesButton.tsx
eranshabi: are you sure? @shlomitc told be to use `svg2react-icon`

src/icons/raw/Check.svg
eranshabi: are you sure? @shlomitc told be to use `svg2react-icon`

src/components/StatesButton/StatesButton.spec.tsx
```diff
@@ -0,0 +1,84 @@
+import { createUniDriverFactory } from 'wix-ui-test-utils/uni-driver-factory';
+import * as React from 'react';
+import { StatesButton } from './StatesButton';
+import { statesButtonDriverFactory } from './StatesButton.driver';
+
+jest.useFakeTimers();
```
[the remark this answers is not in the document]
eranshabi: unidriver's wait didn't work 🤷‍♂

src/components/StatesButton/StatesButton.tsx
eranshabi: @shlomitc suggested `StatesButton` so please decide with him what name you want
(`ProgressButton` sounds better imo)

src/components/StatesButton/StatesButton.tsx
jonathanadler: yeah, it's just a bit of a complication for this pr.
We can do this on another branch.
And maybe we'll be able to upgrade to Yoshi 4, and use the svg import feature.

src/components/StatesButton/StatesButton.tsx
jonathanadler: @shlomitc  what do you think?

src/components/StatesButton/StatesButton.tsx
jonathanadler: What I meant was that instead of adding classes, we can add a `success` state in the `st.css` file, and give a different styling for the this state:
in `StatesButton.tsx`:
```jsx
<Button
    ...
    {...style('root', { success }, this.props )}
```
in `StatesButton.st.css`:
```
.root {
    ...
    -st-states: success;
    ...
}

.root:success {
    // some styling here...
}
```

NitayRabi:
@shlomitc - I Agree with @jonathanadler's proposal - 
1. I would rename the callback `onProgressReset` instead of `onProgressEnd` (It's more descriptive) 
2. In general passing the control to the container component is a better approach for such

Regarding the icons discussion - 
1. Adding a "framework" type PR should be done separately (considering readability/icon-overriding options etc, re-usage of core icons etc..)

jonathanadler:
> @shlomitc - I Agree with @jonathanadler's proposal -
> 
> 1. I would rename the callback `onProgressReset` instead of `onProgressEnd` (It's more descriptive)
> 2. In general passing the control to the container component is a better approach for such
> 
> Regarding the icons discussion -
> 
> 1. Adding a "framework" type PR should be done separately (considering readability/icon-overriding options etc, re-usage of core icons etc..)

👍 
I like the `onProgressReset` suggestion

src/icons/raw/Check.svg
jonathanadler: @eranshabi 
Yeah, lets start with working on the Button itself, and will add the icon support later

src/components/StatesButton/StatesButton.tsx
jonathanadler: I got @shlomitc's approval for the name 😄

src/components/StatesButton/StatesButton.tsx
```diff
@@ -0,0 +1,65 @@
...
+    await delay(2000);
+    this.setState({ success: false });
+  }
+
+  public render() {
+    const { text, disabled, dataHook, ...rest } = this.props;
+    const { success } = this.state;
+    return (
+      <Button
+        data-hook={dataHook}
```
shlomitc: This is not needed. We decided to ditch the camelCase datHook in those new wix-ui lobs

src/icons/raw/Check.svg
shlomitc: In general - this icons looks like a mess. A lot of unnecessary meta data and wrong icon structure. You should follow these guidelines 
https://github.com/wix/wix-ui/blob/master/packages/wix-ui-icons-common/README.md#guidelines

eranshabi: Hey I don't think this needs a fix because it's only the raw svg the ux lady gave me. You can see in the `dist` folder after you build the project that all the unnecessary meta data will not get into production.

src/components/StatesButton/docs/index.story.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { MockSettings } from '../../../../stories/helperComponents/MockSettings';
+import { SIZE } from '../../Button';
+
+const ref: React.RefObject<StatesButton> = React.createRef();
+
+const defaultProps = {
+  disabled: false,
+  onClick: () => {
+    // tslint:disable-next-line:no-floating-promises
+    ref.current.onActionSuccess();
```
jonathanadler: btw, note that the build is failing because `onActionSuccess` does not exist

PR #108, opened 2019-06-17
Feat: yoshi upgrade

medasiaudvytyte:


stories/helperComponents/MockSettings.tsx
```diff
@@ -261,13 +261,11 @@ export class MockSettings extends React.PureComponent<
...
             style: { bold: false, italic: false, underline: false },
-            value: `font:normal normal normal 30px/1.4em ${
-              font.cssFontFamily
-            };`,
+            value: `font:normal normal normal 30px/1.4em ${font.cssFontFamily};`,
           };
           return obj;
         },
-        {},
+        { textFont: undefined },
```
jonathanadler: Why was this added? What does this affect?

mantaspa:
> Looks very good!
> 
> I'd be happy if you could write here, a short explanation on why the changes in `external-types.d.ts` and `tsconfig.json` files were made

I have reverted changes in these files, since they break other stuff. Only type declarations is added to tsconfig.json (including SVG).

PR #110, opened 2019-06-18
move icons to assets + add es module support

jonathanadler:


package.json
```diff
@@ -120,6 +122,7 @@
   },
   "yoshi": {
     "exports": "wix-ui-tpa",
-    "tpaStyle": true
+    "tpaStyle": true,
+    "entry": "./index.js"
```
shlomitc: not sure this is required

src/index.js
```diff
@@ -0,0 +1,16 @@
+import * as Icons from './assets/icons';
+export { Icons };
```
shlomitc: so do you suggest using this or what's in line 1?

jonathanadler: It seems that using:
```
export * as Icons from './assets/icons';
```
Is experimental and not supported.
The solution is to split it to two:
```
import * as Icons from './assets/icons';
export { Icons };
```

package.json
jonathanadler: I guess that's true

src/index.js
jonathanadler: @shlomitc was this what you meant?

PR #112, opened 2019-06-20
<Pagination /> component

mbutsykin:


src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { ChevronLeftLarge, ChevronRightLarge } from 'wix-ui-icons-common';
+
+import { TPAComponentsConsumer } from '../TPAComponentsConfig';
+
+import { DoubleChevronLeftLarge, DoubleChevronRightLarge } from './icons';
+import styles from './Pagination.st.css';
+
+export interface PaginationProps extends CorePaginationProps {
+  /** Justify content */
+  alignment?: 'left' | 'center' | 'right';
```
sdoomz: no need to redefine all properties from `CorePaginationProps` just extend core props with `alignment` and `mobileView`

src/components/Pagination/icons.tsx
```diff
@@ -0,0 +1,16 @@
+import * as React from 'react';
+import { ChevronRightLarge, ChevronLeftLarge } from 'wix-ui-icons-common';
+
+export const DoubleChevronLeftLarge = () => (
```
sdoomz: @jonathanadler can you help us to get double chevron icon from designers, so we can update it here https://wix-wix-ui-icons-common.surge.sh/?selectedKind=Icons%2FGeneral&selectedStory=Small&full=0&addons=0&stories=1&panelRight=0

src/components/Pagination/Pagination.tsx
sdoomz: `alignment` not optional anymore, at least as long as you are defining it in `defaultProps`

src/components/Pagination/icons.tsx
jonathanadler: Sure.
In the meantime you can use the icons from here: https://zeroheight.com/7sjjzhgo2/p/884c1f/t/b83d (click on the icon, and in the top right corner you'll have a knife icon, that will lead you to the asset download button).
I'll check with the designers that this is updated.

But note that our icons are not the same as in wix-ui-icons-common.
We have our own icons folder which we export and also use internally (We now use Yoshi 4, so we can just import SVG's).

src/components/Pagination/Pagination.st.css
```diff
@@ -0,0 +1,263 @@
...
+  /* TEXT NAV BUTTONS */
+  navButtonTextSize: 15px;
+  navButtonTextColor: #000;
+  navButtonDisabledTextColor: #b1b1b1;
+
+  /* @type FONT */
+  fnt: font_8;
+  bRadius: 0;
+  shadow: none;
+}
```
sdoomz: we should also specify `--overridable` styles

src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,85 @@
...
+  /** Show / hide last page */
+  showLastPage?: boolean;
+  /** Show total pages in input mode */
+  showInputModeTotalPages?: boolean;
+  /** Is parent width influence pagination appearance */
+  responsive?: boolean;
+  /** How much pages will be visible in pagination */
+  maxPagesToShow?: number;
+  /** Disable pagination */
+  disabled?: boolean;
```
sdoomz: `data-hook` prop is missing

src/components/Pagination/icons.tsx
jonathanadler: So I checked with the designers.
The icons I sent you are up to date.
So if you managed to use them, that's perfect.
Just note that you might need to clean them from `id`'s `defs`'s etc.

src/components/Pagination/Pagination.spec.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+describe('Pagination', () => {
+  const createDriver = createDriverFactory(paginationDriverFactory);
+
+  it('should render', async () => {
+    const driver = createDriver(<Pagination totalPages={10} />);
+    expect(driver.exists()).toBe(true);
+  });
+
+  it('should use mobile design', async () => {
```
sdoomz: we need test to verify data-hook is set

src/components/Pagination/Pagination.st.css
```diff
@@ -0,0 +1,264 @@
...
+
+.root:disabled::pageButton {
+  cursor: default;
+}
+
+.root::navButton {
+  min-width: 1em;
+  min-height: 1em;
+  padding: calc(0.75em - value(navButtonBorderWidth));
```
jonathanadler: I think either have `navButtonBorderWidth: 3px`, or `padding: calc(0.75em - value(navButtonBorderWidth)px);`
Or am I missing something?

jonathanadler: Same for all places we need a `px`-ed value

src/components/Pagination/Pagination.st.css
```diff
@@ -0,0 +1,264 @@
...
+  color: value(navButtonTextColor);
+  line-height: 1em;
+}
+
+.root::navButton:disabled span, .root[data-preview="disabled"]::navButton span {
+  color: value(navButtonDisabledTextColor);
+}
+
+.root::navButtonFirst {
+  margin-right: 1em;
```
jonathanadler: Again, `rtl` might be a problem here:
Either we use `margin-right` but the first `navButton` will always appear on the left side (meaning not using `flex`), or using `flex`, and having an empty `div` with `width: 1em` or something like that.

src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { ChevronLeft } from '../../assets/icons/ChevronLeft';
+import { ChevronRight } from '../../assets/icons/ChevronRight';
+
+import { TPAComponentsConsumer } from '../TPAComponentsConfig';
+
+import styles from './Pagination.st.css';
+
+export interface PaginationProps extends CorePaginationProps {
+  /** Justify content */
+  alignment?: 'left' | 'center' | 'right';
```
jonathanadler: We should have an `enum` for the `alignment`.
```
export enum Alignment {
    LEFT = 'left',
    CENTER = 'center',
    RIGHT = 'Right',
}
```

src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,57 @@
...
+};
+
+Pagination.displayName = 'Pagination';
+
+Pagination.defaultProps = {
+  maxPagesToShow: 5,
+  totalPages: 10,
+  alignment: 'center',
+  onChange: () =>
+    console.warn('wix-ui-tpa/Pagination: onChange is not implemented'),
```
jonathanadler: I think we can remove the log.
Do we want the `onChange` to be required maybe?

src/components/Pagination/README.md
```diff
@@ -0,0 +1,36 @@
...
+An implementation of Pagination for TPAs.
+The TPA implementation provides few extendable theme properties with the ability to override the base properties as well,
+
+## Theme properties (All are optional)
+
+| propName   | propType | defaultValue | description |
+|------------|----------|--------------|-------------|
+| TextFont | string   | Palette font_8 | The font of the pagination pages |
+| TextFontColor  | string   | #000 | The color of the pagination pages |
+| SelectedTextFontColor  | string   | #ed24d9 | The color of the selected page |
```
jonathanadler: As stated before, these should be template colors.
I'm checking with the designers which colors they want them to be.

src/components/Pagination/docs/index.story.tsx
```diff
@@ -0,0 +1,23 @@
...
+    currentPage: 1,
+    onChange: ({ page }) => setState({ currentPage: page }),
+  }),
+
+  exampleProps: {
+    alignment: ['left', 'center', 'right'],
+    paginationMode: ['pages', 'input'],
+    onChange: ({ page }) => `Selected page: ${page}`,
+  },
+};
```
jonathanadler: Please add sections to the story.
You can see an example for it here:
https://github.com/wix/wix-ui-tpa/blob/master/src/components/LikeButton/docs/index.story.tsx

It will gives us a very readable story.

src/components/Pagination/docs/examples.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export const mobileExample = `
+<TPAComponentsProvider value={{ mobile: true }}>
+  <Pagination 
+    maxPagesToShow={5}
+    totalPages={10}
+    currentPage={1}
+  />
+</TPAComponentsProvider>
+`;
```
jonathanadler: We should probably add the `MockSettings` panel, as we have in other stories (see https://github.com/wix/wix-ui-tpa/blob/master/src/components/LikeButton/docs/extendedExamples.tsx for an example).
This will let us see that the style wiring works the way we want it to.

src/components/Pagination/Pagination.st.css
```diff
@@ -0,0 +1,264 @@
...
+
+  /* TEXT NAV BUTTONS */
+  navButtonTextSize: 15px;
+  navButtonTextColor: #000;
+  navButtonDisabledTextColor: #b1b1b1;
+
+  /* @type FONT */
+  fnt: font_8;
+  bRadius: 0;
+  shadow: none;
```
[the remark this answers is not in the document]
sdoomz: We also need to prefix all default properties with `default` keyword, ex. `defaultFont`, `defaultSelectedPageTextColor`etc...

src/components/Pagination/Pagination.st.css
```diff
@@ -0,0 +1,264 @@
...
+  arrowsColor: #000;
+  arrowsDisabledColor: #b1b1b1;
+
+  /* TEXT NAV BUTTONS */
+  navButtonTextSize: 15px;
+  navButtonTextColor: #000;
+  navButtonDisabledTextColor: #b1b1b1;
+
+  /* @type FONT */
+  fnt: font_8;
```
sdoomz: `font_8` => `{theme: 'Body-M', size: '16px'}`

src/components/Pagination/Pagination.st.css
```diff
@@ -0,0 +1,264 @@
...
+
+.root:alignment(left) {
+  justify-content: flex-start;
+}
+
+.root:alignment(center) {
+  justify-content: center;
+}
+.root:alignment(right) {
+  justify-content: flex-end;
```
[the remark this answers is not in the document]
sdoomz: @jonathanadler yes, alignment is something that we have in our internal design spec. But it sounds logical to have this in `wix-ui-tpa` as well,  for example `Tabs` have them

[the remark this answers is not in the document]
jonathanadler: @sdoomz So I talked to the designers about it, and they decided that the component shouldn't support alignment.
It should be an `inline-block` element, which will take its content's width.
The difference from the `Tabs` component, is that the `Tabs` has a full width design, where its bottom border extends beyond the tabs' widths.

src/components/Pagination/Pagination.spec.tsx
```diff
@@ -0,0 +1,57 @@
...
+    const driver = createDriver(
+      TPAComponentsWrapper({ mobile: true })(<Pagination totalPages={10} />),
+    );
+    expect(await driver.isMobile()).toBe(true);
+  });
+
```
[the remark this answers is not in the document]
mbutsykin: these cases are fully covered in `wix-ui-core` there is no need to retest them.
im overriding few props so i guess i could write tests for them
```
showFirstLastNavButtons={props.totalPages > 5}
paginationMode={mobileView ? 'input' : 'pages'}
```

src/components/Pagination/Pagination.e2e.ts
```diff
@@ -0,0 +1,27 @@
+import { browser } from 'protractor';
+import {
+  createStoryUrl,
+  waitForVisibilityOf,
+} from 'wix-ui-test-utils/protractor';
+import { paginationTestkitFactory } from '../../testkit/protractor';
+
+/**
+ * For tests containing interactions.
+ * Can be removed if not used.
```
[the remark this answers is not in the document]
mbutsykin: you are right this file contains only one test
```
await driver.element().isDisplayed()
``` 
which basically tests if element contains `data-hook` attr. 
but i would prefer to stay with that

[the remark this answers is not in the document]
jonathanadler: Sure no problem.
It's just that this is also tested with enzyme.

src/components/Pagination/Pagination.st.css
```diff
@@ -0,0 +1,205 @@
...
+  max-width: 4ch;
+  min-width: 1ch;
+  padding-left: 0.25em;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  color: wixApply(color, wixApply(fallback, value(TextFontColor), value(defaultTextFontColor)));
+  align-self: center;
+}
+
+@media all and (-ms-high-contrast: none), (-ms-high-contrast: active) {
```
jonathanadler: 🤔 what does this do?
[Here](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/-ms-high-contrast) it says its pretty much deprecated

src/components/Pagination/Pagination.st.css
```diff
@@ -0,0 +1,205 @@
...
+
+.root::navButton svg {
+  width: 1em;
+  min-width: 16px;
+  height: 1em;
+  min-height: 16px;
+  fill: wixApply(color, wixApply(fallback, value(TextFontColor), value(defaultTextFontColor)));
+}
+
+.root[dir="rtl"]::navButton svg {
```
jonathanadler: So regarding the `rtl` prop:
I would prefer having the style **not** be aware of `rtl`.
The `rtl` prop is more for the logic in the `tsx` files, for accessibility for instance.

However, if we have to use `rtl` prop here, then I think it should be used as a stylable state and not using `dir=rtl` (which by the way, I didn't see was added to the element)

src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,46 @@
...
+/** Pagination component */
+export const Pagination: React.FunctionComponent<PaginationProps> = props => {
+  return (
+    <TPAComponentsConsumer>
+      {({ mobile: mobileView, rtl }) => {
+        return (
+          <CorePagination
+            {...styles('root', { mobileView, rtl }, props)}
+            previousLabel={rtl ? ChevronLeft : ChevronRight}
+            nextLabel={!rtl ? ChevronLeft : ChevronRight}
```
jonathanadler: I think this is flipped, no?

jonathanadler: Oh, btw, I wasn't aware of the `rtl` prop that the `core` component has.
I think that it means we don't need to add anything special in our code, because I think it automatically flips the prev/next buttons.

src/components/Pagination/Pagination.st.css
mbutsykin: have no idea =]
will remove it

src/components/Pagination/Pagination.tsx
mbutsykin: yup. it's swaps prev/next buttons - so icons looks weird (i.e. prev button with next icon etc)

jonathanadler: So maybe this `rtl ? ChevronLeft : ChevronRight` isn't needed?

mbutsykin: w/o this fix:
Screen Shot 2019-06-26 at 6 14 21 PM
with:
Screen Shot 2019-06-26 at 6 14 41 PM

mbutsykin: you are right)
rechecked - corrected

jonathanadler:
Oh, another thing I forgot to mention, is that we are missing accessibility for this component.
However, if i'm not mistaken, I think we need to add this in `wix-ui-core`.

mbutsykin:
@jonathanadler 
please recheck RTL related stuff, i removed JS icons swap logic & reverted `[dir=rtl]` styles back
here is how it looks now
Screen Shot 2019-06-26 at 6 45 54 PM

Main consequences for RTL are
- no native svg icons mirror
- spacing issue for first / last nav buttons

so those stiles fixing issues described above

src/components/Pagination/Pagination.st.css
mbutsykin: it's added to the element in [wix-ui-core](https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/pagination/Pagination.tsx#L307)

src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,46 @@
...
+          <CorePagination
+            {...styles('root', { mobileView, rtl }, props)}
+            previousLabel={ChevronLeft}
+            nextLabel={ChevronRight}
+            firstLabel={DoubleChevronLeft}
+            lastLabel={DoubleChevronRight}
+            rtl={rtl}
+            showFirstLastNavButtons={props.totalPages > props.maxPagesToShow}
+            paginationMode={mobileView ? 'input' : 'pages'}
+            {...props}
```
jonathanadler: @mbutsykin @sdoomz 
I think we're also missing `showInputModeTotalPages={mobileView}`
And that's it!

src/components/Pagination/Pagination.st.css
```diff
@@ -0,0 +1,198 @@
...
+  content: "";
+  position: absolute;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+}
+
+.root:not(:disabled)::pageButton:hover {
+  opacity: 0.7;
```
jonathanadler: Designer asked to add a `.2s` transition on hover

mbutsykin: done =)

PR #114, opened 2019-06-24
Project cleanup

jonathanadler:


jonathanadler:
> Looks great. I would try to think how to make the fallback functionality more simple if possible

Funny, because this method also bothered me.
I thought of adding a `default` argument, to every method, as the last argument.
For instance: `color(someColor, someDefaultColor)`.
However, this solution wont be documented, and very readable.

webpack.config.storybook.js
```diff
@@ -4,7 +4,7 @@ const wixStorybookConfig = require('yoshi/config/webpack.config.storybook');
 const StylableWebpackPlugin = require('@stylable/webpack-plugin');
 const project = require('yoshi-config');
 const { resolveNamespaceFactory } = require('@stylable/node');
-const autoprefixer = require('autoprefixer')({ grid: true, browsers: ['>1%'] });
+const autoprefixer = require('autoprefixer')({ grid: true, overrideBrowserslist: ['>1%'] });
```
shlomitc: what is the future of this configuration?
it seems to me like we need to understand if it's related only to storybook or also to actual build.
As I see it - this won't work on production builds unless done in `yoshi`

jonathanadler: Right, we should check with `yoshi` guys.
At least it's "only" for ie11 support

src/common/styles.js
```diff
@@ -0,0 +1,29 @@
...
+
+
+function createStyleFunc (func) {
+    return function () {
+        var args = Array.prototype.slice.call(arguments, 0);
+        args.unshift(func);
+        return applyStyleMethod.apply(null, args);
+    };
+}
+
```
argshook: the es5 here makes it look like something complex is happening. maybe it cannot be es6 for some reason?

if it can, then it's this:

```js
const applyStyleMethod = (func, ...args) => {
  const unquotedArgs = args.map(arg => arg.replace(/"/g, '')).join(', ')
  return `"${func}(${unquotedArgs})"`
};

const createStyleFunc = func => (...args) => applyStyleMethod(func, ...args)
```

jonathanadler: That's true.
I'll see if it works 👍

shlomitc: I think stylable doesn’t transpire es6, but please check

jonathanadler: > I think stylable doesn’t transpire es6, but please check

Yeah, that's what I thought. But i'll try.

argshook: it's meh, i just stumbled upon and couldn't resist :) you don't have to do this

jonathanadler: Done!

PR #118, opened 2019-06-25
generate testkits exports files with wuf

argshook:
this PR introduces `wuf export testkits` to automatically generate:

1. `testkit/index.ts`
1. `testkit/enzyme.ts`
1. `testkit/protractor.ts`

this setup is the same as in wix-style-react. It's not ideal, but it
works, no longer involves custom scripts or manual work.

# impact:

* export missing components (divider, stripButton were not exported)
* ~lose typesript definitions~
* lose exports for raw drivers (those not called with `testkitFactoryCreator`)

# next steps

* [x] generate exported file with typescript definitions
* reduce the amount of `.wuf/update-components.sh`

both steps are things planned to do in wix-ui-framework


argshook:
unfortunately dropping typescript from exported files is a no no and becomes sort of a blocker (build doesn't pass, potentially library users will get failing builds too)

thus, back to drawing board...

argshook:
@jonathanadler it's alive!

i updated wuf to support ejs. This means that each testkit exports file
(`src/testkit/index.ts` and similar) have a corresponding template file
written in ejs (`.wuf/testkits/vanilla.template.ejs`).

This way there's total control on how testkit exports file should look
like. Template file has two globals:

1. `components` - list of all components that are found in `.wuf/components.json`
2. `utils` - just some helpful functions, currently only `toCamel`

locally i manage to get all tests passed. PR seems to be ready to merge
but please skim through.

.wuf/components.json
```diff
@@ -0,0 +1,47 @@
+{
```
shlomitc: Is this auto generated or manual?

.wuf/testkits/definitions.js
```diff
@@ -0,0 +1,70 @@
+/*
```
shlomitc: Since wuf is common across multiple libraries - this should be documented in the wuf repo and not per file in project.

Please treat this as a public API like any open source utility that you’d use

.wuf/components.json
argshook: It's generated

.wuf/testkits/definitions.js
argshook: I agree but im not yet sure of the overall flow, so dont spend too much time on documenting.

I dont want exactly this as public api. For now its copy paste from wix-style-react but I don't think this is ideal, there should be no `definitions` thing at all. My approach is first make it work (but dont break anything), then find patterns, then APIify them (abd of course document).

So its in progress, i wont leave it like that forever :)

.wuf/components.json
shlomitc: So put it in gitignore. It would avoid unnecessary PR conflicts which should not change anything in the user flow

.wuf/testkits/definitions.js
shlomitc: So just copy-paste it to the wuf project instead of duplicating it again.

.wuf/testkits/definitions.js
```diff
@@ -0,0 +1,70 @@
...
+  },
+
+  Tabs: {
+    unidriver: true
+  },
+
+  Divider: {
+    noTestkit: true
+  }
+};
```
jonathanadler: I'm not sure I understand.
Which components do we need to add a configuration for here?

.wuf/required-component-files.json
```diff
@@ -0,0 +1,7 @@
+{
```
jonathanadler: Why is this file needed?
I'm asking to understand how this works

argshook:
I wrote a little bit here
https://github.com/wix/wix-ui/blob/master/packages/wix-ui-framework/docs/usage.md

I keep evolving the approach so didnt document everything yet, as it
changes often. Its getting more and more stable but I'm not there yet.

Ideally there should be no configuration at all, but since not all
components adhere to convention (or there is no convention) something
somewhere has to tell wuf how to deal with different shapes.

My goal now is to make things work and reduce manual labor. Next is to
reduce all these configs to minimum.

I can tell you more tomorrow once im back to office :)

.wuf/required-component-files.json
argshook: this file describes the shape of components, aka the convention.

`wuf update` reads this file, then goes through file system and looks
for folders that look like this shape.

folders that look similar are considered components and are saved in `.wuf/components.json` file.

then, `.wuf/components.json` file is used in `wuf export-testkits` to generate testkits exports file.

later on this file will be used for many more things:

1. storybook index
1. components exports file (something like src/index.js)
1. `code` section in autodocs (instead of currently manual `allComponents` part in `code({ components: allComponents })`)
1. auditing (what components do not adhere to conventions)

argshook: later on this file could even be used to describe some of file contents (it's complex but doable)

for example, to ensure that testkit file exports testkit and a type
definition. In different library, maybe testkit file exports something
else, so there's a place to set this shape

.wuf/testkits/definitions.js
argshook: i wrote more examples here: https://github.com/wix/wix-ui/blob/master/packages/wix-ui-framework/docs/testkits-definitions.md

the idea is that this file is for "snowflake" components.
i'll update this so that non unidrivers are considered snowflakes, not the opposite

package.json
```diff
@@ -43,7 +44,9 @@
     "compile-base-ui": "node ./.storybook/base-ui-compiler.js",
     "storybook:build": "build-storybook -s ./src/assets",
     "generate-stylable-index": "stc --srcDir=\"./dist/src/components\" --diagnostics --indexFile=index.st.css",
-    "generate": "wuf generate",
+    "generate": "wuf generate --codemods .wuf/generator/codemods --templates .wuf/generator/templates",
+    "postgenerate": "npm run update-components",
+    "update-components": ".wuf/update-components.sh",
```
jonathanadler: Super picky comment: Maybe this should be called `generate-testkits`? or something like that?

argshook: good comment :)

maybe it should, but looking at future this will do more than generate testkits so i chose this name. also, it's same as in wsr, so maybe consistency.

just so you know, i don't like the `update-components.sh`, it should be just `wuf update`. But for now, multiple commands are needed and i didn't want to add them all to `package.json`

i can name it `generate-testkits` if you want, tell me :)

jonathanadler: I agree, consistency at this point might be more important.
and +1 for `wuf update`

argshook:
Checked as many wix-ui-tpa consumers as i could find, this change looks safe.

I managed to find only enzyme testkit users. This means that raw driver exports that are now gone (which is good) will have no impact.

Merging, if there're issues, please point me to them, thanks :)

PR #121, opened 2019-06-27
Pagination component

mbutsykin:
- disabled nav buttons opacity


jonathanadler:
@mbutsykin 
The designers say that there should be a default color of `color-3` for the disabled color.
TPA's should use this color, unless they want to let the users override the color.
Is that enough for you?

mbutsykin:
So this PR is redundant, closing =]

PR #127, opened 2019-07-08
fix wix-style-processor hmr problems in storybook

jonathanadler:
hack wix-style-processor to fix hmr problems in storybook when changing st.css files

Problem is happening because of this line:
https://github.com/wix/wix-style-processor/blob/cf38dd80ba3e0b142b25b4313a42d6a7f9baede2/src/styleUpdater.ts#L25

Which causes `wix-style-processor` to ignore changes to style tags

plus

I had to add a timeout to let `wix-style-processor` run after hmr runs

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -1,30 +1,17 @@
 import {configure, addParameters} from '@storybook/react';
 import styleProcessor from 'wix-style-processor';
-import domService from 'wix-style-processor/dist/es/src/domService';
-
-const fixedDomService = {
-  ...domService,
-  overrideStyle: (tag, css) => {
-    const result = domService.overrideStyle(tag, css);
-    delete tag.originalTemplate;
-    return result;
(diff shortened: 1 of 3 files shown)


jonathanadler:
Another possible fix is to change this:
https://github.com/wix/wix-ui-tpa/blob/f79cf83a5ee610eabb624ce8abae7848fd722812/webpack.config.storybook.js#L17

from ```includeCSSInJS: !stylableSeparateCss,```
to ```includeCSSInJS: true,```

then hmr will also work in this case

jonathanadler:
@shlomitc 
I checked, and the OOI project doesn't seem to have HMR at all.
Maybe I can sit with Ran, and find a solution for this, a solution that we might use here as well.

In the meantime we can use the solution in this PR, wdyt?

shlomitc:
@jonathanadler ok

PR #131, opened 2019-07-10
Fix visual tests + run them (fix tests not running)

jonathanadler:


jonathanadler:
@shlomitc I'd like to hear what you think.
Until this PR the visual test did not run (because of a stupid mistake I made - searching for the wrong env variable).
So I fixed that, and fixed the visual tests a bit.
But then I noticed the tests are flaky , because the snapshot are sometimes taken before `wix-style-processor` has finished running (or the browser updating).
So I used the `waitBeforeScreenshots` configuration, and added a wrapper component which, for each visual test, runs ```wixStyleProcessor.init()```, and then sets the `data-*` attribute as written in the config.
It seems to be working fine now, but i'd like to hear what you think.

jonathanadler:
> Looks good, however I don’t think we want to initialize the style processor for each component but rather only once.
> 
> What if you keep the style processor initialization to one place and write on some global variable the returned promise. Then the VisualContainer can wait for it to be resolved and set state.

I also thought about that, but for some reason it doesn't work. 
Maybe because each story is inside a different iframe, the script needs to be run for each one.
Also, if applitools run a test for each story, it might not make a difference, since the style processor will be called for each test in any case.

jonathanadler:
> > Looks good, however I don’t think we want to initialize the style processor for each component but rather only once.
> > What if you keep the style processor initialization to one place and write on some global variable the returned promise. Then the VisualContainer can wait for it to be resolved and set state.
> 
> I also thought about that, but for some reason it doesn't work.
> Maybe because each story is inside a different iframe, the script needs to be run for each one.
> Also, if applitools run a test for each story, it might not make a difference, since the style processor will be called for each test in any case.

Wait, I think I got it working.
I'll push changes.

.storybook-test/config.js
```diff
@@ -1,23 +1,14 @@
...
-configure(loadStories, module);
 
-addParameters({
-  options: {
-    name: 'wix-ui-tpa',
-    url: 'https://github.com/wix-private/wix-ui-tpa',
-    showAddonPanel: false,
-  }
-});
+  setTimeout(init);
```
jonathanadler: @shlomitc 
I put the `styleProcessor.init()` call here, in the `config` file.
However I have to add a timeout for this to work.
Not the `VisualContainerElement` only needs to wait for the promise to resolve, and the `init` is called once.
Wdyt?

test/visual/StyleProcessorUtil.ts
```diff
@@ -0,0 +1,15 @@
+import styleProcessor from 'wix-style-processor';
```
jonathanadler: @shlomitc 
This is where the `styleProcessor` is initiated.

PR #132, opened 2019-07-15
<Avatar/> component

kranux:


src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,103 @@
...
+        {...styles('root', { size }, rest)}
+        imgProps={{ src }}
+        placeholder={placeholder()}
+        text={name && name.split('')[0]}
+      />
+    );
+  }
+}
+
+function getDefaultAvatar(avatarSize: AvatarSize = AvatarSize.medium) {
```
jonathanadler: I think we should create a `Anonymous.svg` file in the Avatar's folder, and use it instead of having it inline in this file.
So you'll have something like this:
```tsx
import { ReactElement as AnonymousSVG } from './Anonymous.svg';
...
render() {
 return (
  <AnonymousSVG width={widthBySize} height={heightBySize} />
 )
}
```

jonathanadler: So you should probably take the biggest SVG, and use it for the file

src/components/Avatar/Avatar.visual.tsx
```diff
@@ -0,0 +1,34 @@
...
+  }
+}
+
+const tests = [
+  {
+    describe: 'basic',
+    its: [
+      {
+        it: 'default',
+        props: {},
```
jonathanadler: We should render the three types of Avatars, with the different sizings for the visual tests.
You can take a look at other `*.visual` files as a reference.

src/components/Avatar/README.md
```diff
@@ -0,0 +1,4 @@
+## Avatar
```
jonathanadler: I recommend taking a look at other README's, and using the same format for this one.

src/components/Avatar/Avatar.meta.tsx
```diff
@@ -0,0 +1,12 @@
+import { Avatar } from '.';
+import Registry from '@ui-autotools/registry';
+
+const AvatarMetadata = Registry.getComponentMetadata(Avatar);
+AvatarMetadata.nonReactStrictModeCompliant = true;
+
+AvatarMetadata.addSim({
+  title: 'render',
+  props: {
+
```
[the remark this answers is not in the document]
kranux: Added few more simulations. Question: should I cover all the cases?

src/components/Avatar/Avatar.visual.tsx
kranux: Added visual tests for all the scenarios;

src/components/Avatar/README.md
kranux: Yes, but this one is different from others, as we do not have any re-usable theme properties here, right?

src/components/Avatar/Avatar.tsx
kranux: Are you sure this trick works in *ts?

kranux: Sorry, it's `ReactComponent` and it works in ts :)

src/components/Avatar/Avatar.meta.tsx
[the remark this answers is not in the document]
jonathanadler: I think the sizes might be less important to check, but the different states: `default`, `initials` & `anonymous` are more imprtant

src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,55 @@
...
+        {...styles('root', { size }, rest)}
+        imgProps={{ src }}
+        placeholder={<AnonymousSVG width={dimmentionBySize(size)} height={dimmentionBySize(size)} />}
+        text={name && name.split('')[0]}
+      />
+    );
+  }
+}
+
+function dimmentionBySize(avatarSize: AvatarSize = AvatarSize.medium) {
```
jonathanadler: it's a bit picky, but i'd rather have this function be a private method within the `Avatar` class

src/testkit/enzyme.ts
```diff
@@ -1,4 +1,5 @@
 /* eslint-disable */
+/* tslint:disable */
```
jonathanadler: This is irrelevant because the file is autogenerated

src/testkit/index.ts
```diff
@@ -1,4 +1,5 @@
 /* eslint-disable */
+/* tslint:disable */
```
jonathanadler: This is irrelevant because the file is autogenerated

src/testkit/protractor.ts
```diff
@@ -1,4 +1,5 @@
 /* eslint-disable */
+/* tslint:disable */
```
jonathanadler: This is irrelevant because the file is autogenerated

src/components/Avatar/README.md
jonathanadler: Nevertheless this should be edited, and not left with the default generated text.
You can write a short description, and how to use.
Also you can add a link that its implementation is used with the `wix-ui-core` component.

src/components/Avatar/Avatar.meta.tsx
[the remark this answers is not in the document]
kranux: Ok, great, that's what I did: added all the different states.

src/components/Avatar/Avatar.tsx
kranux: Moved to static map;

src/components/Avatar/README.md
kranux: Added usage example and links to wix-ui-core;

src/components/Avatar/docs/examples.ts
```diff
@@ -0,0 +1,19 @@
...
+export const example = Object.keys(AvatarSize).reduce((acc, size) => {
+  return {
+    ...acc,
+    [size]: getAvatars(size),
+  };
+}, {});
+
+function getAvatars(size) {
+  return `
+  <>
```
jonathanadler: Lets use `<div>` instead of `<>`.
Because this shows in the documentation.

kranux: Done;

src/components/Avatar/Avatar.st.css
```diff
@@ -0,0 +1,52 @@
...
+    -st-extends: Avatar;
+    -st-states: size(enum(large, medium, small, xLarge));
+}
+
+.root::content {
+    align-items: center;
+    background-color: #d8d8d8;
+    color: #979797;
+    display: flex;
+    font-family: "HelveticaNeueW01-65Medium";
```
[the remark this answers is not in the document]
kranux: Let me know when I can change it :)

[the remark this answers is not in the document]
jonathanadler: So this should be `HelveticaNeueW01-65Medi`, and maybe lets have a callback for `Arial, sans-serif` or something like that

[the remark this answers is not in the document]
kranux: Thanks, added! :)

jonathanadler:
@kranux can you try to freeze the `@applitools/eyes-storybook` version to see if that fixes the failing tests.
You can use: 2.6.25

jonathanadler:
> @kranux can you try to freeze the `@applitools/eyes-storybook` version to see if that fixes the failing tests.
> You can use: 2.6.25

Nevermind, I already added a commit for this

src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,58 @@
...
+      <CoreAvatar
+        {...styles('root', { size }, rest)}
+        imgProps={{ src }}
+        placeholder={
+          <Anonymous
+            height={Avatar.dimmentionBySize[size]}
+            width={Avatar.dimmentionBySize[size]}
+          />
+        }
+        text={name && name.split('')[0]}
```
jonathanadler: As we discussed, it's better to change this to `name`

kranux: Yes, did that.

PR #133, opened 2019-07-16
Icon button

zhukovka:
https://zeroheight.com/7sjjzhgo2/p/9607e0/t/97a111


src/components/IconButton/IconButton.driver.ts
```diff
@@ -0,0 +1,19 @@
+import {
+  BaseUniDriver,
+  baseUniDriverFactory,
+} from 'wix-ui-test-utils/base-driver';
+import { StylableUnidriverUtil, UniDriver } from 'wix-ui-test-utils/unidriver';
+import style from './IconButton.st.css';
+
+export interface IconButtonDriver extends BaseUniDriver {
+  isMobile(): Promise<boolean>;
```
jonathanadler: From what I see in the spec, there is no `mobile` design for this component.
So you can remove all the references to `mobile` that were generated automatically

src/components/IconButton/IconButton.e2e.ts
```diff
@@ -0,0 +1,27 @@
+import { browser } from 'protractor';
```
jonathanadler: Since we don't have any special interactions we need to test, we can remove this file

src/components/IconButton/IconButton.spec.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { ReactComponent as StarIcon } from '../../assets/icons/Star.svg';
+
+describe('IconButton', () => {
+  const createDriver = createUniDriverFactory(iconButtonDriverFactory);
+
+  it('should render', async () => {
+    const driver = createDriver(<IconButton icon={<StarIcon />} />);
+    expect(await driver.exists()).toBe(true);
+  });
+
```
jonathanadler: Maybe we can add a test to see that the `onClick` is called correctly?

src/components/IconButton/IconButton.st.css
```diff
@@ -0,0 +1,55 @@
...
+    text-decoration: none;
+    border: none;
+    display: inline-block;
+    padding: 0;
+    margin: 0;
+    line-height: 0;
+}
+
+.root:hover .icon {
+    opacity: applyOpacity(value(IconColor), 0.6);;
```
jonathanadler: The `applyOpactiy` changes the opacity of a given color.
In this case your changing the `opacity` so it can just be `0.6`.
Meaning, you can write either:
```
opacity: 0.6;
```
or:
```
stroke: applyOpacity(value(IconColor), 0.6);
```

src/components/IconButton/IconButton.st.css
```diff
@@ -0,0 +1,55 @@
...
+    stroke: color(value(DisabledColor));
+}
+
+.icon svg {
+    display: block;
+}
+
+.icon path {
+    stroke: color(fallback(value(IconColor), value(DefaultColor)));
+    fill: color(fallback(value(IconColor), value(DefaultColor)));
```
jonathanadler: I think the `fill` should always be `transparent`, no?

src/components/IconButton/IconButton.tsx
```diff
@@ -0,0 +1,31 @@
...
+/** An implementation of an IconButton for TPAs, The TPA implementation provides few extendable theme properties with the ability to override the base properties as well,By default the IconButton will pull the theme from the site (site colors), */
+export class IconButton extends React.Component<IconButtonProps, State> {
+  static displayName = 'IconButton';
+  static defaultProps: DefaultProps = {};
+
+  render() {
+    const { icon, disabled, ...rest } = this.props;
+    return (
+      <ButtonNext
+        aria-label={'icon button'}
```
jonathanadler: This should be passed by the developer using the component, since the content of the `aria-label` should be translated into the language of the site.
So it's better to leave it to them

src/components/IconButton/IconButton.visual.tsx
```diff
@@ -0,0 +1,40 @@
...
+  }
+}
+
+const tests = [
+  {
+    describe: 'basic',
+    its: [
+      {
+        it: 'default',
+        props: {},
```
jonathanadler: Let's add a test for the disabled state.
You can see how we did it in the `LikeButton` or `IconToggle`

src/components/IconButton/README.md
```diff
@@ -0,0 +1,4 @@
+## IconButton
```
jonathanadler: We should probably have a more informative README.
You can look at the other buttons for instance.

src/components/IconButton/docs/examples.tsx
```diff
@@ -0,0 +1,35 @@
+import * as React from 'react';
+
+import CodeExample from 'wix-storybook-utils/CodeExample';
+import * as ExtendedRawSource from '!raw-loader!./IconButtonExtendedExample.tsx';
+import * as ExtendedCSSRawSource from '!raw-loader!./IconButtonExtendedExample.st.css';
+import { MockSettings } from '../../../../stories/helperComponents/MockSettings';
+import { IconButtonExtendedExample } from './IconButtonExtendedExample';
+
+export const Example = () => (
```
jonathanadler: Instead of using the example with the `MockSetting` here, please look at the `Card`'s storybook `index` file.
we added a new function called `settingsPanel` that gives a new tab for this.
And instead have a simple example in this file here, maybe just two icons with the `disabled` `true` and `false`.

src/components/IconButton/IconButton.tsx
zhukovka: but it is a default value (if no passed from a user) it'll be overwritten

src/components/IconButton/IconButton.st.css
zhukovka: no this is not the case, the share icon (from my example has to change fill value)

jonathanadler: Hmm, but it's not like that in the zeroheight.
Can you send me the spec you're working with?
I'll ask the designers in the meantime.

src/components/IconButton/IconButton.tsx
jonathanadler: Yeah, I hear what your'e saying.
But still, I think it might be redundant, and maybe not the right product for everyone.
I prefer that it won't say anything if someone leaves this by mistake.
You could say it's kind of removing the responsibility from our side 😄

src/components/IconButton/IconButton.st.css
zhukovka: actually I've downloaded the `share` icon from zeroheight

zhukovka: and it is designed in the way it has to change `fill` prop

src/components/IconButton/IconButton.tsx
zhukovka: Can't make a11y tests pass without it ```Running a11y test...
1. Testing component IconButton render...
  IconButton render - button: (Impact: critical)
  Fix any of the following:
  Element does not have inner text that is visible to screen readers
  aria-label attribute does not exist or is empty
  aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty
  Element's default semantics were not overridden with role="presentation"
  Element's default semantics were not overridden with role="none"
  Element has no title attribute or the title attribute is empty
```
may be you could advice me on how to deal with it?

Comment (jonathanadler) [reply src/components/IconButton/IconButton.tsx]: I understand...
So for this, inside the `IconButton.meta.tsx` file you can pass an `aria-label` prop with `'share'` for instance

File: src/components/IconButton/IconButton.meta.tsx [hunk 10]
```diff
@@ -0,0 +1,14 @@
...
+import { ChevronLeft } from '../../assets/icons/ChevronLeft';
+
+const IconButtonMetadata = Registry.getComponentMetadata(IconButton);
+IconButtonMetadata.nonReactStrictModeCompliant = true;
+
+IconButtonMetadata.addSim({
+  title: 'render',
+  props: {
+    icon: ChevronLeft,
+  },
```
jonathanadler: You can add the `aria-label` here:
```
IconButtonMetadata.addSim({
  title: 'render',
  props: {
    icon: ChevronLeft,
    ['aria-label']: 'share',
  },
});
```

src/components/IconButton/IconButton.driver.ts
```diff
@@ -0,0 +1,16 @@
...
+  BaseUniDriver,
+  baseUniDriverFactory,
+} from 'wix-ui-test-utils/base-driver';
+import { StylableUnidriverUtil, UniDriver } from 'wix-ui-test-utils/unidriver';
+import style from './IconButton.st.css';
+
+export interface IconButtonDriver extends BaseUniDriver {}
+
+export const iconButtonDriverFactory = (base: UniDriver): IconButtonDriver => {
+  const stylableUtil = new StylableUnidriverUtil(style);
```
jonathanadler: You can remove this, since it's not used

src/components/IconButton/IconButton.st.css
```diff
@@ -0,0 +1,55 @@
...
+    padding: 0;
+    margin: 0;
+    line-height: 0;
+}
+
+.root:hover .icon {
+    opacity: 0.6;
+}
+
+.root:disabled:hover .icon {
```
jonathanadler: I think we also need to add a `cursor: default` for the `disabled` mode

src/components/IconButton/IconButton.st.css
jonathanadler: I still haven't got an answer from the designers

src/components/IconButton/IconButton.tsx
jonathanadler: So you can remove this now 😄

zhukovka:
> If you switch the `disabled` toggle on, it looks a bit blurry.
> I couldn't understand why that happens.
yes, it is blurry because I've broke it with adding the `fill` value 
`fill` has to be set for my `share` button but it breaks icons that should be outlined with `stroke` (`fill:none` shouldn't be overwritten), I think I need to ask a designer if he(she) can change this `share` icon so it could be changed with just `stroke` prop.

jonathanadler:
So I got an update from the designers, that there should be an option for `full` icon buttons.
You can see it here:
https://zeroheight.com/7sjjzhgo2/p/9607e0/t/97a111

This means we need to add a prop `isFull` or something like that.

src/assets/icons/ShareFilled.svg
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jonathanadler: I think we should have only one icon, and it should be possible to "fill" using css.
If it doesn't work with the `Share.svg` icon then we need to ask the designers for a better asset.
But this file should be removed.

src/assets/icons/Share.svg
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jonathanadler: can you please add an export for this file in `src/assets/icons/index.tsx`?

src/components/IconButton/IconButton.st.css
```diff
@@ -0,0 +1,67 @@
...
+    text-decoration: none;
+    border: none;
+    display: inline-block;
+    padding: 0;
+    margin: 0;
+    line-height: 0;
+}
+
+.root:hover .icon {
+    opacity: 0.6;
```
jonathanadler: For some reason the `hover` and `disabled` states don't work.
I tried here:
https://wix-wix-ui-tpa-pr-133.surge.sh/?activeTab=Playground&path=%2Fstory%2Fcomponents--iconbutton

I think in order to fix this we need to have the `.icon` have `display: inline-block` (or `block`).

src/components/IconButton/IconButton.tsx
```diff
@@ -0,0 +1,28 @@
+import * as React from 'react';
+import { ButtonProps } from '../Button';
+import style from './IconButton.st.css';
+import { ButtonNext } from 'wix-ui-core/button-next';
+
+export interface IconButtonProps extends ButtonProps {
+  icon: React.ReactNode;
+  isFull?: boolean;
```
jonathanadler: I'm sorry for driving you crazy with this 😬 
But i'm still not sure how we should add the `full` prop.
Now i'm thinking maybe we should have a `skin` prop with: 
```
enum Skins {
  Line = 'line';
  Full = 'full';
}
```
and `Line` will be default.
This way, if in the future we want to add another skin, we can add it here.
What do you think?

src/components/IconButton/IconButton.visual.tsx
```diff
@@ -0,0 +1,51 @@
...
+        <IconButton
+          icon={<ShareIcon />}
+          isFull
+          data-hook={'storybook-e2e-IconButton'}
+          {...this.props}
+        />
+        <IconButton
+          icon={<ShareIcon />}
+          data-hook={'storybook-e2e-IconButton'}
+          isFull
```
jonathanadler: Nice!
We should also add tests for not-full icons

src/components/IconButton/docs/index.story.tsx
```diff
@@ -0,0 +1,108 @@
...
+HeartIcon.displayName = 'Heart';
+
+const iconExamples = [
+  {
+    label: 'Share',
+    value: <ShareIcon />,
+  },
+  {
+    label: 'ShareFilled',
+    value: <ShareFilledIcon />,
```
jonathanadler: So we should remove this, and have an example with the `full` skin/prop

src/assets/icons/Share.svg
zhukovka: there is no `src/assets/icons/index.tsx` file but there is `src/assets/icons/index.js` so should I add the import to a `.js` file or create `src/assets/icons/index.tsx`?

jonathanadler: You're correct!
It the `.js` file.
I should have created a `.ts` file 🤔 
I'll fix that in a different PR.
Thanks!

src/components/IconButton/IconButton.visual.tsx
```diff
@@ -0,0 +1,50 @@
+import * as React from 'react';
+import { storiesOf } from '@storybook/react';
+import { IconButton, Skins } from './';
+import { ReactComponent as ShareIcon } from '../../assets/icons/Share.svg';
+import { VisualContainerElement } from '../../../test/visual/VisualContainerElement';
+
+class IconButtonVisual extends React.Component<any> {
+  static defaultProps = {
+    mobile: false,
```
jonathanadler: This mobile reference can be removed

src/components/IconButton/IconButton.visual.tsx
```diff
@@ -0,0 +1,50 @@
...
+          icon={<ShareIcon />}
+          data-hook={'storybook-e2e-IconButton'}
+          {...this.props}
+        />
+        <IconButton
+          icon={<ShareIcon />}
+          data-hook={'storybook-e2e-IconButton'}
+          skin={Skins.Full}
+          {...this.props}
+          disabled
```
jonathanadler: instead of having the `disabled` property here, lets add it to the array of tests below:
```js
const tests = [
  {
    describe: 'basic',
    its: [
      {
        it: 'default',
        props: {},
      },
      {
        it: 'default',
        props: {
          disabled: true,
        },
      },
    ],
  },
];
```

jonathanadler: and the render function will be:
```
<VisualContainerElement>
        <IconButton
          icon={<ShareIcon />}
          {...this.props}
        />
        <IconButton
          icon={<ShareIcon />}
          skin={Skins.Full}
          {...this.props}
        />
      </VisualContainerElement>
    );
```
(we can also remove the data hooks, we'll remove it from generator)

src/components/IconButton/docs/index.story.tsx
```diff
@@ -0,0 +1,102 @@
...
+  component: IconButton,
+  componentPath: '../IconButton.tsx',
+  componentProps: () => ({
+    'data-hook': 'storybook-IconButton',
+    icon: iconExamples[0].value,
+    disabled: false,
+    skin: Skins.Line,
+  }),
+  exampleProps: {
+    icon: iconExamples,
```
jonathanadler: Let's add the `skin` prop to the `exampleProps` as well.
You can see how we did it in the button story for instance.

src/components/IconButton/IconButton.st.css
```diff
@@ -0,0 +1,71 @@
...
+    text-decoration: none;
+    border: none;
+    display: inline-block;
+    padding: 0;
+    margin: 0;
+    line-height: 0;
+}
+
+.root:hover {
+    opacity: 0.6;
```
jonathanadler: For some reason this isn't working for `line` skin.
You can see here:
https://wix-wix-ui-tpa-pr-133.surge.sh/?activeTab=Playground&path=%2F%2Fstory%2Fcomponents--iconbutton

src/components/IconButton/IconButton.st.css
```diff
@@ -0,0 +1,71 @@
...
+.root:disabled {
+    cursor: default;
+}
+
+.root:disabled:hover {
+    opacity: 1;
+}
+
+.root:disabled .icon path {
+    stroke: color(value(DisabledColor));
```
jonathanadler: This works for the `'full` disabled, but not for `line` for some reason.
Maybe the icons aren't good.
I think it's because you have the `.root:skin(line) .icon path` rule below.
Maybe if you put it before it will work?

jonathanadler: or add a rule for `.root:skin(line):disabled .icon path`

src/components/IconButton/IconButton.st.css
zhukovka: I can see that hover works for both skins...

zhukovka:
🥳🎉

Issue #94, opened 2019-06-05
upgrade to latest storybook 5.1.x


liorMar:
https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#from-version-50x-to-51x

jonathanadler:
resolved

PR #136, opened 2019-07-16
Improve linting flow

giladsegal:
- Add prettier
- Exclude autogenerated testkit files
- Add automatic linting before commit

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,8 +84,10 @@
     "enzyme-adapter-react-16": "^1.11.0",
     "eyes.it": "^2.0.3",
     "gh-pages-auto-release": "latest",
+    "husky": "^3.0.0",
     "identity-obj-proxy": "^3.0.0",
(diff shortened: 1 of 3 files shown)


jonathanadler:
fixes #94

package.json
```diff
@@ -102,6 +104,14 @@
     "yoshi": "^4.9.0",
     "yoshi-style-dependencies": "^4.6.0"
   },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
```
argshook: @giladsegal would you consider `pre-push` instead?

jonathanadler: I was also thinking about that, but i'm not sure.
On the one hand, having it `pre-commit` lets the developer add the lint fixes directly to the commit.
On the other hand, having it in `pre-push` reduces the amount of times linting will run, and might be less of a "pain".

PR #138, opened 2019-07-17
<AvatarGroup/> component

kranux:


src/components/AvatarGroup/docs/index.story.tsx
```diff
@@ -0,0 +1,83 @@
...
+        ],
+      }),
+
+      ...[
+        { title: 'API', sections: [api()] },
+        { title: 'TestKit', sections: [testkit()] },
+        { title: 'Playground', sections: [playground()] },
+      ].map(tab),
+    ]),
+  ],
```
jonathanadler: Please add a `SettingsPanel` tab.
You can see an example for it in the `Card`'s storybook `index` file.
It will let us play with the border color, as if it was in the editor.

kranux:
@jonathanadler after I added this one: https://github.com/wix/wix-ui-tpa/pull/138/commits/8183bb70617c1f3559cb68f1cd0433ac73ab83ee Started thinking that this part is not right: https://github.com/wix/wix-ui-tpa/pull/138/files#diff-c47927270c9e95105753300ec7b2e24dR50 https://github.com/wix/wix-ui-tpa/pull/138/files#diff-f103c7dccb4729a4b0d0c41f267df3e8R76 

It's because we have multiple top level elements with states implemented in single Stylable file. In another hand I am unable to use -st-extends with complex selector.

What do you think?

src/components/AvatarGroup/docs/index.story.tsx
kranux: Added settings panel here: https://github.com/wix/wix-ui-tpa/pull/138/commits/d9f4fedba9cbc3539c226ff988d0c8a9c6eba314

Trying to understand why component has some css class duplicated and looses stacking effect https://wix-wix-ui-tpa-pr-138.surge.sh/?path=///story/components--avatargroup

Probably, something wrong with Stylable setup in AvatarGroup component.

src/components/AvatarGroup/AvatarGroup.spec.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { avatarGroupTestkitFactory as enzymeAvatarGroupTestkitFactory } from '../../testkit/enzyme';
+
+describe('AvatarGroup', () => {
+  const createDriver = createUniDriverFactory(avatarGroupDriverFactory);
+
+  it('should render', async () => {
+    const driver = createDriver(<AvatarGroup items={[]} />);
+    expect(await driver.exists()).toBe(true);
+  });
+
```
jonathanadler: I just thought that it would be nice to have a test to see that the Avatar never shows more than the `max` number of avatars.
It's good as a guard, so that if anyone makes changes in the future, we can be assured that this behaviour will not break.

src/components/AvatarGroup/AvatarGroup.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  render() {
+    const { items, size, maxAmount, ...rest } = this.props;
+
+    return (
+      <div {...styles('root', { size }, rest)}>
+        {items.slice(0, maxAmount).map(({ name, src }) => (
+          <div className={styles.avatarContainer}>
+            <Avatar
+              {...styles('avatar', { size }, rest)}
```
jonathanadler: The problem lies here.
Instead of passing the the "spreaded" `styles` object, you should only pass the class name, like this: ```className={styles.avatar}```

src/components/AvatarGroup/AvatarGroup.spec.tsx
kranux: I'm kind of checking this in visual tests. But maybe it makes sense to add it here as well :)

src/components/AvatarGroup/AvatarGroup.tsx
kranux: Thought about this. There are few things: 
1) Avatar does not accept `className` prop;
2) Avatar must have `size`

I believe this is related with https://github.com/wix/wix-ui-tpa/pull/138#issuecomment-512679463

src/components/AvatarGroup/AvatarGroup.spec.tsx
kranux: Agree, adding tests :)

src/components/AvatarGroup/AvatarGroup.tsx
jonathanadler: So, the `Stylable` automatically accepts a `className` prop (that's why we send the `rest` or `props` to the spreaded `styles` object i.e. ```...styles('root', {}, rest)```.
So this works out of the box.
Regarding the size, it should be passed as a prop then, and not in stylable.

Usually I prefer having a stylable only on the `root`.
It makes things a bit more readable and neat IMO.

src/components/AvatarGroup/AvatarGroup.spec.tsx
kranux: Added;

src/components/AvatarGroup/AvatarGroup.tsx
kranux: Issue is here:
```
Type '{ className: string; name: string; src: string; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<Avatar> & Pick<Readonly<AvatarProps> & Readonly<{ children?: ReactNode; }>, "children" | "name" | "src"> & Partial<Pick<Readonly<...> & Readonly<...>, "size">> & Partial<...>'.
```

kranux: I can bypass TS thing by spreading props, or improving props. Thats ok.

And talking about `size` we agreed that we will not have all the sizes in avatar, as designers added different sets to Avatar vs AvatarGroup.

kranux: Thanks to @mykas I found a way to solve it properly. 

I missed some concepts on how Stylable works :)

src/components/AvatarGroup/AvatarGroup.st.css
```diff
@@ -0,0 +1,129 @@
...
+}
+
+.root:size(small) .avatar::content,
+.root:size(medium) .avatar::content,
+.root:size(large) .avatar::content {
+    border-width: 2px;
+}
+
+.root:size(large) .avatar::content {
+    border-radius: 26px;
```
jonathanadler: I see in the storybook, that the border radius isn't working so good.
What I think we can do, is instead of defining all the different `border-radius`'s let's add:
```border-radius: 50%;``` to `.root::content`
It should work for all the avatar sizes

File: src/components/AvatarGroup/AvatarGroup.tsx [hunk 5]
```diff
@@ -0,0 +1,55 @@
...
+  };
+
+  render() {
+    const { items, size, maxAmount, ...rest } = this.props;
+
+    return (
+      <div {...styles('root', { size }, rest)}>
+        {items.slice(0, maxAmount).map(({ name, src }, key) => (
+          <div className={styles.avatarContainer} key={key}>
+            <Avatar {...{ className: styles.avatar }} name={name} src={src} />
```
jonathanadler: super picky, and really not important, but this can be changed to:
```<Avatar className={styles.avatar} ...```

Comment (kranux) [reply src/components/AvatarGroup/AvatarGroup.tsx]: No , cannot change, as Avatar Props doesn't have `className`

File: src/components/AvatarGroup/AvatarGroup.st.css (hunk 4 above)
Comment (kranux) [reply src/components/AvatarGroup/AvatarGroup.st.css]: Working on this :)

File: src/components/AvatarGroup/AvatarGroup.tsx (hunk 5 above)
Comment (jonathanadler) [reply src/components/AvatarGroup/AvatarGroup.tsx]: 🤦‍♂ typescript
ok ignore
thanks!

Comment (kranux) [reply src/components/AvatarGroup/AvatarGroup.tsx]: Fix pushed;

File: src/components/AvatarGroup/AvatarGroup.st.css [hunk 6]
```diff
@@ -0,0 +1,125 @@
...
+}
+
+.avatar {
+    -st-extends: Avatar;
+}
+
+.root .avatar::content {
+    border: solid 1px;
+    border-color: color(value(avatarBorderColorValue));
+    border-radius: 50%;
```
jonathanadler: Sorry for the picky comment 😬 
But lets move the `border-radius` to `Avatar.st.css`.

src/components/AvatarGroup/AvatarGroup.driver.ts
```diff
@@ -0,0 +1,20 @@
...
+import style from './AvatarGroup.st.css';
+
+export interface AvatarGroupDriver extends BaseUniDriver {
+  getAvatarCount(): Promise<number>;
+}
+
+export const avatarGroupDriverFactory = (
+  base: UniDriver,
+): AvatarGroupDriver => {
+  const getAvatarCount = () => base.$$(`.${style.root} > *`).count();
```
jonathanadler: another thing, I should have asked this before,
Let's use a `data-*` instead of `style.root`.
We sometimes have errors when using style imports inside drivers. 
It can make problems in e2e tests sometimes.
So maybe you can pass a `data-hook="avatar"` to the `Avatar`'s or something like that,
And then in this file you can use `base.$$('[data-hook="avatar"]').count()`?

src/components/AvatarGroup/AvatarGroup.st.css
kranux: Done

src/components/AvatarGroup/AvatarGroup.driver.ts
kranux: Did that;

PR #140, opened 2019-07-17
<Badge/>

nataliia-khitsan:
Badge base implementation according to this spec https://zeroheight.com/7sjjzhgo2/p/723c40/t/39dc09

Still waiting for an answer from designers which styles can be overridable.


src/components/Badge/Badge.driver.ts
```diff
@@ -0,0 +1,23 @@
+import {
+  BaseUniDriver,
+  baseUniDriverFactory,
+} from 'wix-ui-test-utils/base-driver';
+import { StylableUnidriverUtil, UniDriver } from 'wix-ui-test-utils/unidriver';
+import style from './Badge.st.css';
```
jonathanadler: I know it's not documented anywhere, but we're having some problems importing `*.st.css` files inside tests.
It would be better if you could, instead of using the `StylableUnidriverUtil` here to check if it has the styles,
use a `data-*` hook for the `priority`.
So for instance to check if `isPrimary` we'll just need to check if `[data-priority="primary"]` exists.
It will work better, cause then tests will not need to import stylable files, which sometimes causes problems.
It's something that we'll need to fix in general in the library.

src/components/Badge/Badge.e2e.ts
```diff
@@ -0,0 +1,27 @@
+import { browser } from 'protractor';
```
jonathanadler: This file can be removed, if there are no special cases to check here (other than `should render`)

src/components/Badge/Badge.spec.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { badgeTestkitFactory as enzymeBadgeTestkitFactory } from '../../testkit/enzyme';
+
+describe('Badge', () => {
+  const createDriver = createUniDriverFactory(badgeDriverFactory);
+
+  it('should render', async () => {
+    const driver = createDriver(<Badge>I'm default badge</Badge>);
+    expect(await driver.exists()).toBe(true);
+  });
+
```
jonathanadler: Maybe we can check to see that we have different priorities set successfully

src/components/Badge/Badge.visual.tsx
```diff
@@ -0,0 +1,43 @@
...
+  }
+}
+
+const tests = [
+  {
+    describe: 'basic',
+    its: [
+      {
+        it: 'default',
+        props: {},
```
jonathanadler: Let's add the different priorities for the visual tests, so we know that the different styles won't get broken

src/components/Badge/README.md
```diff
@@ -0,0 +1,2 @@
+## Badge
```
jonathanadler: You can look at the `Button` README as a reference for the Badge's README

src/components/Badge/docs/index.story.tsx
```diff
@@ -0,0 +1,63 @@
...
+        ],
+      }),
+
+      ...[
+        { title: 'API', sections: [api()] },
+        { title: 'TestKit', sections: [testkit()] },
+        { title: 'Playground', sections: [playground()] },
+      ].map(tab),
+    ]),
+  ],
```
jonathanadler: Please add a `SettingsPanel` tab.
You can see the implementation in the `Card` story `index` file.
It will give us an opportunity to play with the component's styling

src/components/Badge/Badge.st.css
```diff
@@ -0,0 +1,43 @@
+:import {
+    -st-from: "wix-ui-core/index.st.css";
+    -st-named: Badge;
```
jonathanadler: I see you're importing the Badge style from `wix-ui-core` but you haven't used the component in the `*.tsx` file.
Is there a reason for this?

src/components/Badge/Badge.e2e.ts
nataliia-khitsan: removed

src/components/Badge/Badge.st.css
nataliia-khitsan: Committed by mistake, do not need it

src/components/Badge/Badge.meta.tsx
```diff
@@ -0,0 +1,12 @@
+import {Badge, BADGE_PRIORITY} from '.';
+import Registry from '@ui-autotools/registry';
+
+const BadgeMetadata = Registry.getComponentMetadata(Badge);
+BadgeMetadata.nonReactStrictModeCompliant = true;
+
+BadgeMetadata.addSim({
+  title: 'render',
+  props: {
+    priority: BADGE_PRIORITY.default,
```
jonathanadler: We should probably also add a `children` prop with some text

src/components/Badge/Badge.visual.tsx
```diff
@@ -0,0 +1,60 @@
...
+class BadgeVisual extends React.Component<any> {
+  static defaultProps = {
+    mobile: false,
+  };
+
+  render() {
+    const { mobile } = this.props;
+
+    return (
+      <TPAComponentsProvider value={{ mobile }}>
```
jonathanadler: We can also remove the `TPAComponentsProvider` here, since there's no `mobile` state.
I know it's a bit of a pain.
We're considering removing the `mobile` state from the generator, since it's causing so much refactoring of the code.

src/components/Badge/README.md
```diff
@@ -0,0 +1,18 @@
...
+An implementation of an Badge for TPAs.
+
+The Badge TPA implementation provides a few default overridable styles. By default site theme is used.
+
+## Theme properties (all are optional)
+| propName          | propType | description                                                                                                        |
+|-------------------|----------|--------------------------------------------------------------------------------------------------------------------|
+| BadgeBgColor      | string   | The color of the badge background instead of the default color for certain type                                    |
+| BadgeTextColor    | string   | The color of the bade text instead of the default color for certain type                                           |
+| BadgeBorderColor  | string   | The color of the bade border instead of the default color for certain type. Can be set only for the **Light** badge|
```
jonathanadler: The border is not only for the `light` priority.
It can also be changed by the user in other priorities.

src/components/Badge/docs/index.story.tsx
```diff
@@ -0,0 +1,65 @@
...
+  examples: <ExtendedExample/>
```
jonathanadler: The extended example looks good, but please use the `SettingsPanel` function.
You can see how we use it here:
https://github.com/wix/wix-ui-tpa/blob/master/src/components/Card/docs/index.story.tsx#L106

src/components/Badge/Badge.meta.tsx
nataliia-khitsan: But `children` do not exist in BadgeProps

jonathanadler: 🤔 but we use `children` in `Badge.st.css`

src/components/Badge/Badge.st.css
```diff
@@ -0,0 +1,60 @@
...
+.root {
+    -st-states: priority(enum(default, light, primary));
+    display: inline-block;
+    border-radius: 12px;
+    padding: 0 12px;
+    line-height: 20px;
+    font-size: 12px;
+    border-width: 1px;
+    border-style: solid;
+    background-color: applyOpacity(value(badgeBgColor), 0.4);
```
jonathanadler: The opacity should be applied only on the preset colors.
When users override the colors, they should be able to choose whatever color and opacity they want.
I suggest moving the `applyOpacity` function to `vars` definition in lines 25-27.

stories/utils/SettingsPanel.tsx
```diff
@@ -11,7 +17,7 @@ export function settingsPanel({ example, rawSource, rawCSSSource, params }) {
     <div>
       <div className="tpa-container">
         <CodeExample
-          title="Card Extended"
+          title={title}
```
jonathanadler: Nice catch! Thanks 🙏

stories/index.tsx
```diff
@@ -16,4 +16,4 @@ require('../src/components/ToggleSwitch/docs'); // TODO: convert to .story
 require('../src/components/Tooltip/docs/index.story');
 require('../src/components/Tabs/docs/TabsTestStory');
 require('../src/components/Pagination/docs/index.story');
-require('../src/components/Avatar/docs/index.story');
```
jonathanadler: Just add the `Avatar` story back.
Probably happened when rebasing.

PR #146, opened 2019-07-18
New card

jonathanadler:
work in progress


src/components/NewCard/docs/NewCardExample.tsx
```diff
@@ -0,0 +1,140 @@
...
+              <Text typography={TYPOGRAPHY.smallTitle}>Overlapping</Text>
+              <br />
+              <Text typography={TYPOGRAPHY.runningText}>
+                I have 200px min-width, and the media has 300px min-width, so I
+                break when the total width is &lt; 500px
+              </Text>
+            </div>
+          </NewCard.Container>
+        </NewCard>
+      </div>
```
jonathanadler: So the designers asked me to remove this example, since they don't want people to use this type of layout.

src/components/NewCard/docs/NewCardExample.st.css
```diff
@@ -0,0 +1,101 @@
...
+    margin-left: calc(-1 * value(overlappingMargin) / 2);
+}
+
+.overlapping.card::container.info .overlapContent {
+    border: 1px solid black;
+    background-color: white;
+    height: 100%;
+    width: 100%;
+    padding: value(overlappingMargin);
+}
```
jonathanadler: Same here for the overlapping example

src/components/NewCard/README.md
```diff
@@ -0,0 +1,4 @@
+## NewCard
+An implementation of an NewCard for TPAs
+This Readme was auto-generated and needs to be filled manually
```
jonathanadler: Will need to add a pretty short description and example on how to use the component.
We should also state that this is the formal and correct `Card` to use and that this will be moved to the current `Card` component, which will be deprecated in the near future.

PR #147, opened 2019-07-19
Feat/rating

bordi:
Base implementation of `<Ratings />` component. 
Spec: https://zeroheight.com/7sjjzhgo2/p/05a744/t/708026 
All labels are **not** implemented.


src/components/Rating/README.md
```diff
@@ -0,0 +1,4 @@
+## Rating
```
jonathanadler: Add an extended README.
You can see how the rest of the components have their README.
Button for instance

src/components/Rating/Rating.driver.ts
```diff
@@ -0,0 +1,19 @@
+import {
+  BaseUniDriver,
+  baseUniDriverFactory,
+} from 'wix-ui-test-utils/base-driver';
+import { StylableUnidriverUtil, UniDriver } from 'wix-ui-test-utils/unidriver';
+import style from './Rating.st.css';
+
+export interface RatingDriver extends BaseUniDriver {
+  isMobile(): Promise<boolean>;
```
jonathanadler: If there's no `mobile` state for the component, you can remove all the references to the `mobile` state

src/components/Rating/Rating.e2e.ts
```diff
@@ -0,0 +1,27 @@
+import { browser } from 'protractor';
```
jonathanadler: I would add a test that hovers over the component, and tests that the correct number of stars are "hovered"

src/components/Rating/Rating.meta.tsx
```diff
@@ -0,0 +1,13 @@
+import { Rating, Mode } from '.';
```
jonathanadler: This currently isn't working. I'm working on it

src/components/Rating/Rating.spec.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { ratingTestkitFactory } from '../../testkit';
+import { ratingTestkitFactory as enzymeRatingTestkitFactory } from '../../testkit/enzyme';
+
+describe('Rating', () => {
+  const createDriver = createUniDriverFactory(ratingDriverFactory);
+
+  it('should render', async () => {
+    const driver = createDriver(<Rating value={3} />);
+    expect(await driver.exists()).toBe(true);
+  });
```
jonathanadler: We would probably like to have tests for the onclick callback.

src/components/Rating/Rating.st.css
```diff
@@ -0,0 +1,97 @@
...
+
+:vars {
+    /* Overrides */
+    IconColor: --overridable;
+    IconEmptyColor: --overridable;
+}
+
+.root {
+    -st-states: mobile;
+
```
jonathanadler: Maybe we should have the `root` be `inline-block`?
Because currently it's taking the whole width of its parent.
In addition `rtl` isn't working

src/components/Rating/Rating.st.css
```diff
@@ -0,0 +1,97 @@
...
+.root:hover .icon:not(:disabled) path {
+    stroke: color(fallback(value(IconColor), value(DefaultColor)));
+}
+
+.iconList {
+    display: flex;
+    align-items: center;
+    flex-direction: row-reverse;
+
+    float: left;
```
jonathanadler: We need to support `rtl` so it's better to not use floats

src/components/Rating/Rating.st.css
```diff
@@ -0,0 +1,97 @@
...
+}
+
+.icon {
+    -st-states: checked, disabled, error, mode( enum(edit, preview) ), iconSize( enum(large) );
```
jonathanadler: It's better to add all the `states` in the `root` selector, and then do something like this:
```
.root:disabled .icon {
...
}
``` 
using states in other selectors can cause unwanted bugs.

src/components/Rating/Rating.st.css
```diff
@@ -0,0 +1,97 @@
...
+    padding: 0 value(IconGap);
+}
+
+.root .icon:first-child {
+    padding-right: 0;
```
jonathanadler: This might be problematic for `rtl`, we need to think how we can solve this without padding

jonathanadler: Maybe grid?

src/components/Rating/Rating.st.css
```diff
@@ -0,0 +1,97 @@
...
+}
+
+.root .icon:last-child {
+    padding-left: 0;
```
jonathanadler: The same as 👆

src/components/Rating/Rating.st.css
jonathanadler: In addition, we probably want to add a `cursor: pointer` on the `.root` when it's not enabled.

src/components/Rating/Rating.tsx
```diff
@@ -0,0 +1,97 @@
+import * as React from 'react';
+import { Checkbox as CoreCheckbox } from 'wix-ui-core/checkbox';
+import styles from './Rating.st.css';
+import { ReactComponent as StarIcon } from '../../assets/icons/Star.svg';
+
+export enum IconSize {
+  LARGE = 'large',
+}
+
+export enum Mode {
```
jonathanadler: I don't think we need modes.
Is this from the spec?

src/components/Rating/Rating.tsx
```diff
@@ -0,0 +1,97 @@
+import * as React from 'react';
+import { Checkbox as CoreCheckbox } from 'wix-ui-core/checkbox';
+import styles from './Rating.st.css';
+import { ReactComponent as StarIcon } from '../../assets/icons/Star.svg';
+
+export enum IconSize {
+  LARGE = 'large',
```
jonathanadler: Lets use capitalized for the enum props:
```
Large = 'large'
```
I know there is some inconsistency in the library, but I want to start enforcing this.

jonathanadler: And we should also add a `small` size, which will be the default size.

src/components/Rating/Rating.tsx
```diff
@@ -0,0 +1,97 @@
...
+                key={idx}
+                {...styles('icon', {
+                  checked,
+                  disabled,
+                  error,
+                  iconSize,
+                  mode,
+                })}
+              >
+                <CoreCheckbox
```
jonathanadler: Very interesting approach for the component  🤔 
I haven't thought about this kind of implementation.

However, i'm not sure checkboxes are the way to go, in an a11y point of view.
The correct way would probably be using something like a radio group.
You can see what they suggest here: https://www.w3.org/WAI/tutorials/forms/custom-controls/#a-star-rating

I'm not sure we really have to use the radio buttons, but maybe we'll have to pass the correct `aria-*` attributes.

src/components/Rating/Rating.visual.tsx
```diff
@@ -0,0 +1,49 @@
...
+    describe: 'basic',
+    its: [
+      {
+        it: 'default',
+        props: {},
+      },
+      {
+        it: 'mobile',
+        props: {
+          mobile: true,
```
jonathanadler: Again we can probably remove reference for `mobile` state.
In addition we should probably add tests showing different selected `values`, and states (`disabled`, `error`).

src/components/Rating/Rating.tsx
```diff
@@ -0,0 +1,97 @@
...
+export enum Mode {
+  EDIT = 'edit',
+  PREVIEW = 'preview',
+}
+
+export interface RatingProps {
+  value: number;
+  mode: Mode;
+  onSelect?(value: number): void;
+  maxRating?: number;
```
jonathanadler: I talked to the designers, and they said they want the rating component to have only 5 stars.
So I guess we should remove the `maxRating` prop.

src/components/Rating/Rating.tsx
bordi: Will do

src/components/Rating/Rating.tsx
bordi: OK, I'll change it to `radio group`. 
I was a bit confused cs in docs was mention that this component should be based on `IconToggle` and this component implemented on `CoreCheckbox` base.

src/components/Rating/Rating.st.css
bordi: But we can in `rtl` use `float: right`. It's the standard implementation with floats (even in your example from w3.org they are using floats)

src/components/Rating/Rating.spec.tsx
```diff
@@ -0,0 +1,111 @@
...
+    expect(await driver.getActiveStars()).toBe(4);
+  });
+
+  it('should show error state', async () => {
+    const driver = createDriver(<Rating error value={0} />);
+
+    expect(await driver.hasError()).toBeTruthy();
+  });
+
+  it('should show errdisabledor state', async () => {
```
jonathanadler: small typo

src/components/Rating/Rating.spec.tsx
```diff
@@ -0,0 +1,111 @@
...
+  it('should call onSelect with actual value with error state', async () => {
+    const onSelectSpy = jest.fn();
+    const driver = createDriver(<Rating error onSelect={onSelectSpy} />);
+
+    await driver.clickOnStar(3);
+
+    expect(onSelectSpy).toHaveBeenCalledWith(2);
+  });
+
+  it('should NOT call onSelect with mode="display"', async () => {
```
jonathanadler: Excellent 👍

src/components/Rating/Rating.st.css
```diff
@@ -0,0 +1,114 @@
...
+    stroke: color(fallback(value(IconColor), value(DefaultColor)));
+}
+
+.root .icon:checked path {
+    stroke: color(fallback(value(IconColor), value(DefaultColor)));
+    fill: color(fallback(value(IconColor), value(DefaultColor)));
+ }
+
+.root:not(:mode(display)) .icon:hover path,
+.root:not(:mode(display)) .icon:hover ~ .icon path {
```
jonathanadler: 😮 for using `~`

src/components/Rating/Rating.st.css
```diff
@@ -0,0 +1,114 @@
...
+    -st-states: checked;
+
+    padding: 0 value(IconGap);
+}
+
+.root .icon path {
+    stroke: color(fallback(value(IconColor), value(DefaultColor)));
+}
+
+.root .icon:checked path {
```
jonathanadler: Note that the behaviour currently is not exactly the way we want it to be.
When the component is in `input` mode and has a value of 3, for instance, then if the user hovers the first icon, we want to show only 1 full icon, and not 3.
So maybe we need this selector to be something like `:checked:not(:hover)`, or something like that

src/components/Rating/Rating.tsx
```diff
@@ -0,0 +1,88 @@
...
+interface DefaultProps {
+  value: number;
+  disabled: boolean;
+  error: boolean;
+  mode: Mode;
+  iconSize: IconSize;
+}
+
+/** Rating component based on IconToggle */
+export class Rating extends React.Component<RatingProps> {
```
jonathanadler: I know it's super super picky, but I talked to the designers, and they say, that in order to be consistent with the editor, this component should be called `Ratings` and not `Rating` 🤦‍♂ .
Sorry for not checking earlier.

src/components/Rating/Rating.visual.tsx
```diff
@@ -0,0 +1,70 @@
...
+class RatingVisual extends React.Component<any> {
+  static defaultProps = {
+    mobile: false,
+  };
+
+  render() {
+    const { mobile } = this.props;
+
+    return (
+      <TPAComponentsProvider value={{ mobile }}>
```
jonathanadler: The `TPAComponentProvider` can be removed, since we're not using a `mobile` state for this component.

src/components/Rating/Rating.visual.tsx
```diff
@@ -0,0 +1,70 @@
...
+        <VisualContainerElement>
+          <Rating {...this.props} />
+        </VisualContainerElement>
+      </TPAComponentsProvider>
+    );
+  }
+}
+
+const tests = [
```
jonathanadler: Very nice 👍

src/components/Rating/Rating.st.css
bordi: No prob. Also, I will write a test for this case

src/components/Rating/Rating.tsx
bordi: No worries

src/components/Ratings/docs/index.story.tsx
```diff
@@ -0,0 +1,87 @@
...
+export default {
+  category: 'Components',
+  storyName: 'Ratings',
+  component: Ratings,
+  componentPath: '../Ratings.tsx',
+  componentProps: () => ({
+    'data-hook': 'storybook-Ratings',
+  }),
+  exampleProps: {
+    //
```
jonathanadler: Can you add the two example props: `mode` & `iconsize`?
You can see how it's done in `Button`.

src/components/Ratings/Ratings.tsx
```diff
@@ -0,0 +1,184 @@
...
+                key={idx}
+                {...styles(
+                  classNames(styles.icon, `${styles.icon}${humanOrder}`),
+                  {
+                    checked,
+                  },
+                )}
+              >
+                <CoreRadio
+                  aria-label={`${5 - idx} star`}
```
jonathanadler: We need to find a solution for the `aria-label`s here.
I think what we can do is:
If we have `inputOptions` then we can use the as `aria-label`'s,
If not, then lets just use numbers `1`, `2`, `3`, `4`, `5` (`aria-label={5 - idx}`)

src/components/Ratings/docs/RatingsExtendedExample.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { Ratings, Mode } from '../index';
+import extendedStyles from './RatingsExtendedExample.st.css';
+
+export const RatingsExtendedExample = props => (
+  <div>
+    <div>
+      <h3>Ratings Input</h3>
+      <Ratings
+        value={0}
+        onSelect={val => console.log(val)}
```
jonathanadler: In order to make this example more clear for consumers, let's add a state that saves this `Ratings`' value, and change it on `onSelect`.
I think you can use React hooks if you don't want to create a class for this.

src/components/Ratings/Ratings.meta.tsx
```diff
@@ -0,0 +1,48 @@
...
+RatingsMetadata.addSim({
+  title: 'render with value mode="display"',
+  props: {
+    value: 3,
+    mode: Mode.Display,
+  },
+});
+
+RatingsMetadata.addSim({
+  title: 'render with value mode="display" and labels',
```
jonathanadler: I think for some reason the `=` sign is breaking the test:
```shell
Error: Simulation titles must be alphanumeric
```

src/components/Ratings/Ratings.tsx
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  _renderContent = () => <StarIcon />;
+
+  _renderInputOptions = () => {
+    const { inputOptions, value } = this.props;
+
+    return (
+      <div className={styles.labelList}>
+        {inputOptions.map((el, idx) => (
```
jonathanadler: The problem is that `inputOptions` doesn't have a default value, so this might explode.
Maybe we should have `inputOptions` to be an empty array as a default.

src/components/Ratings/docs/index.story.tsx
```diff
@@ -0,0 +1,109 @@
...
+  storyName: 'Ratings',
+  component: Ratings,
+  componentPath: '../Ratings.tsx',
+  componentProps: () => ({
+    'data-hook': 'storybook-Ratings',
+  }),
+  exampleProps: {
+    mode: Object.values(Mode),
+    size: Object.values(Size),
+    layout: Object.values(Layout),
```
jonathanadler: After fixing the `inputOptions` default value thingy, you can add the prop to the `exampleProps` like this:
```js
exampleProps: {
  ...
  inputOptions: [
    { value: [], label: 'no hovers' },
    { value: ['Very Baasa', 'Baasa', 'OK', 'Achla', 'Very Achla'], label: 'with hovers' },
  ]
  ...
}
```

src/components/Ratings/Ratings.tsx
```diff
@@ -0,0 +1,187 @@
...
+      <div
+        {...styles('root', { disabled, error, size, mode, layout }, rest)}
+        {...this.getDataAttributes()}
+        data-hook={this.props['data-hook']}
+      >
+        <div className={styles.iconList}>
+          {ratingList.map((_el, idx: number) => {
+            const humanOrder = ratingListLength - idx;
+            const checked = humanOrder <= value;
+            const ariaLabel = inputOptions
```
jonathanadler: I think we need to add a check for:
```
inputOptions.length ? inputOption[5 - idx - 1] : (5 - idx).toString();
```

src/components/Ratings/Ratings.tsx
```diff
@@ -0,0 +1,281 @@
...
+  };
+
+  handleKeyDown = (e: React.KeyboardEvent<HTMLDivElement>) => {
+    const { currentFocus } = this.state;
+
+    if (currentFocus === 0) {
+      return;
+    }
+
+    switch (e.key) {
```
jonathanadler: It's not really important, but I think I would do something like:

```js
let direction = 0;

switch (key) {
  case UP:
  case RIGHT:
    direction = 1;
  break;
  case DOWN:
  case LEFT:
    direction = -1;
  break;
}

if (direction) {
  let nextFocus = ((currentFocus + direction) % 5);
  nextFocus = nextFocus < 0 ? nextFocus + 5 : nextFocus;
  e.preventDefault();
  this.setState({ currentFocus: nextFocus }, () => {
   this.iconRefs[nextFocus].focus();
  });
  return false;
}
```
This way there's no repetitive code, and the `return` value is in one place.

jonathanadler: btw, we also need to listen for `SPACE` and `ENTER` in order to select the focused star.
And also, we need to "clean" the `currentFocus` on the root's `blur` and `focus` events.

src/components/Ratings/Ratings.st.css
```diff
@@ -0,0 +1,184 @@
...
+.root:size(large) .ratingInfo {
+    font: value(LargeFont);
+}
+
+.root:size(large) .icon {
+    margin-top: 7px;
+}
+
+/* DISABLED */
+.root:hover:not(:disabled):not(:mode(display)) .icon path {
```
jonathanadler: I'm not sure what this rule does, but I am sure that this causes the stars to get blank when hovering over the label.
For instance when I hover over "Achla" all the stars get empty.

src/components/Ratings/Ratings.st.css
```diff
@@ -0,0 +1,184 @@
...
+
+.root:not(:mode(display)):hover .icon:hover path,
+.root:not(:mode(display)):hover .icon:hover ~ .icon path,
+.root:not(:mode(display)) .icon:focus-within path,
+.root:not(:mode(display)) .icon:focus-within ~ .icon path {
+    stroke: color(fallback(value(IconColor), value(DefaultColor)));
+    fill: color(fallback(value(IconColor), value(DefaultColor)));
+}
+
+.root:not(:mode(display)):hover .icon:not(:hover) path {
```
jonathanadler: Again, I'm not sure what this does, and it might be causing some problems with the hover and focus, but i'm not sure.

src/components/Ratings/Ratings.st.css
```diff
@@ -0,0 +1,184 @@
...
+
+.root:not(:focus-within) .icon:checked path {
+    stroke: color(fallback(value(IconColor), value(DefaultColor)));
+    fill: color(fallback(value(IconColor), value(DefaultColor)));
+ }
+
+.root:not(:mode(display)):hover .icon:hover path,
+.root:not(:mode(display)):hover .icon:hover ~ .icon path,
+.root:not(:mode(display)) .icon:focus-within path,
+.root:not(:mode(display)) .icon:focus-within ~ .icon path {
```
jonathanadler: regarding `focus-within` i'm not 100% sure we can use it, because I think it's not supported in ie11 and edge.
If we can't use `focus`, then we might need to use the solution written here: https://stackoverflow.com/a/50788621/891850
(also, there's a similar solution in the `Tab.st.css` file, which is part of the `Tabs` component)

src/components/Ratings/Ratings.tsx
```diff
@@ -0,0 +1,268 @@
...
+              <span
+                data-hook={RATINGS_DATA_HOOKS.IconWrapper}
+                key={idx}
+                {...styles(styles.icon, { checked })}
+                onMouseEnter={() => this.handleHoverIcon(humanOrder)}
+                onMouseLeave={this.handleUnhoverIcon}
+                ref={this.setRef}
+                tabIndex={-1}
+              >
+                <CoreRadio
```
jonathanadler: I think our main problem lies here:
When we focus on the component the `span` above ☝️ is taking the focus, and if we `TAB` again, the CoreRadio component catches the focus.
We need to pass a `tabIndex={-1}` to the CoreRadio, which I'm not sure it currently supports.
Which means we will need to create a simple PR for `wix-ui-core` in order to add the `tabIndex` property.

PR #150, opened 2019-07-22
feat(TextArea): create text area component

AndriiTsarenko:


src/components/TextArea/TextArea.driver.ts
```diff
@@ -0,0 +1,77 @@
...
+function errorIcon(base: UniDriver) {
+  return base.$(`[data-hook="${TEXT_AREA_ERROR_ICON}"]`);
+}
+
+function errorText(base: UniDriver) {
+  return base.$(`[data-hook="${TEXT_AREA_ERROR_DESCRIPTION}"]`);
+}
+
+export const textAreaDriverFactory = (base: UniDriver): TextAreaDriver => {
+  const stylableUtil = new StylableUnidriverUtil(style);
```
jonathanadler: Great usage of data hooks!
I would also like if you could use `data-*` attributes instead of using the `StyleableUnidriverUtil`.
Because using `StyleableUnidriverUtil` requires us to import the `*.st.css` file into the driver, and this can cause problems in the tests.
So for instance, instead of using: ```stylableUtil.hasStyleState(base, 'disabled')```, we can do:
```
base.$(`[${TEXT_AREA_DATA_DISABLED}="true"]`)
```

jonathanadler: I know it came as part of the generator, it's something that we need to fix

src/components/TextArea/TextArea.e2e.ts
```diff
@@ -0,0 +1,27 @@
+import { browser } from 'protractor';
```
jonathanadler: We need to think if this test is really needed, or maybe we can add cases to test in e2e.
Do we have any cases where we need real dom for the test?
If not, we can remove this file.

src/components/TextArea/TextArea.st.css
```diff
@@ -0,0 +1,83 @@
...
+    /*Defaults params goes here*/
+}
+
+:vars {
+    /*Overrides*/
+    BorderColor: --overridable;
+    BackgroundColor: --overridable;
+    TextColor: --overidable;
+    PlaceholderColor: --overidable;
+    FontPreference: --overidable;
```
jonathanadler: a picky comment: I'm not sure `FontPreference` is an understandable prop name.
We can maybe call it `TextFont`.

src/components/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,106 @@
+import * as React from 'react';
+import styles from './TextArea.st.css';
+import {
+  TEXT_AREA_SUCCESS_ICON,
+  TEXT_AREA_DATA_HOOK,
+  TEXT_AREA_ERROR_ICON,
+  TEXT_AREA_ERROR_DESCRIPTION,
+} from './dataHooks';
+// tslint:disable-next-line:no-duplicate-imports
+import { ChangeEvent } from 'react';
```
jonathanadler: No need to import this.
You can use `React.ChangeEvent`.
That's what we do in other components.

src/components/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,106 @@
...
+import { Tooltip } from '../Tooltip';
+
+export enum TextAreaTheme {
+  box = 'box',
```
jonathanadler: just for consistency, lets have the prop names capitalized:
```
Box = 'box',
Line = 'line'
```

src/components/TextArea/TextArea.visual.tsx
```diff
@@ -0,0 +1,99 @@
...
+  static defaultProps: TextAreaProps = {
+    placeholder: 'Placeholder',
+    onChange: () => {},
+    value:
+      'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.',
+  };
+
+  render() {
+    return (
+      <TPAComponentsProvider value={{ mobile }}>
```
jonathanadler: Again, we can remove all the `mobile` references, if we don't have a design for mobile

src/components/TextArea/TextArea.visual.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+interface Test {
+  describe: string;
+  its: {
+    it: string;
+    props: Partial<TextAreaProps>;
+  }[];
+}
+
+const tests: Test[] = [
```
jonathanadler: Good tests 👍

src/components/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,106 @@
...
+import styles from './TextArea.st.css';
+import {
+  TEXT_AREA_SUCCESS_ICON,
+  TEXT_AREA_DATA_HOOK,
+  TEXT_AREA_ERROR_ICON,
+  TEXT_AREA_ERROR_DESCRIPTION,
+} from './dataHooks';
+// tslint:disable-next-line:no-duplicate-imports
+import { ChangeEvent } from 'react';
+import { Tooltip } from '../Tooltip';
```
jonathanadler: This tooltip isn't the same one the `Textarea` should use.
However, we currently don't have one that fits for this component.
You can leave it like this for now, and i'll check with the designers.

src/components/TextArea/TextArea.e2e.ts
AndriiTsarenko: Done

src/assets/icons/Error.svg
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jonathanadler: Let's also add an export of this icon in `icons/index.js`

src/components/TextArea/TextArea.st.css
```diff
@@ -0,0 +1,94 @@
...
+    background-color: value(backgroundColor);
+}
+
+.textArea::placeholder,
+.textArea::-webkit-input-placeholder
+{
+    color: value(placeholderColor);
+}
+
+.errorIcon {
```
jonathanadler: Note that there should be an extra padding when the icon appears (see the zeroheight spec, "Filled w. Error" example).
So I guess this should probably be added as a stylable state to the root, and added to the textarea when the icon appears

jonathanadler: In addition, we need to add rtl direction for this icon.
What I suggest you can do, is to have something like this:
```jsx
<div {...styles('root', {...}, this.props} >
  ...
  <div className={styles.errorIconWrapper}>
    <Tooltip><ErrorIcon /></Tooltip>
  </div>
  ...
</div>
```
```css
.errorIconWrapper {
  position: absolute;
  top: 0;
  left: 0;
  width: 100%;
  display: flex;
  justify-content: flex-end;
}
```

src/components/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,124 @@
+import * as React from 'react';
+import styles from './TextArea.st.css';
+import { TEXT_AREA_DATA_HOOK, TEXT_AREA_ERROR_DATA_HOOK } from './dataHooks';
+// tslint:disable-next-line:no-duplicate-imports
```
jonathanadler: I think this can probably be removed, no?

src/components/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,123 @@
...
+            success,
+            disabled,
+          },
+          rest,
+        )}
+      >
+        <textarea
+          value={value}
+          onChange={e => {
+            if (!disabled) {
```
jonathanadler: This is actually not needed.
Instead we should have a `disabled={disabled}` attribute.

src/components/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,129 @@
...
+export enum TextAreaTheme {
+  Box = 'box',
+  Line = 'line',
+}
+
+export interface TextAreaProps {
+  value: string;
+  error?: boolean;
+  success?: boolean;
+  dataHook?: string;
```
jonathanadler: There's no need to add the `dataHook` prop unless you're using it in your tests.
In any case, if you want to leave this, so we now want to use the native `data-hook` instead of `dataHook`.
Just so it will be more like the native html components.

src/components/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,129 @@
...
+  error: boolean;
+  success: boolean;
+  disabled: boolean;
+  dataHook: string;
+  placeholder: string;
+  theme: TextAreaTheme;
+  errorDescription: string;
+}
+
+interface State {}
```
jonathanadler: The `State` can be removed, and not used in the class declaration.

src/components/TextArea/TextArea.driver.ts
```diff
@@ -0,0 +1,79 @@
+import {
+  BaseUniDriver,
+  baseUniDriverFactory,
+} from 'wix-ui-test-utils/base-driver';
+import { UniDriver } from 'wix-ui-test-utils/unidriver';
+import { TEXT_AREA_DATA_HOOK, TEXT_AREA_ERROR_DATA_HOOK } from './dataHooks';
+import { TextAreaTheme } from './TextArea';
```
jonathanadler: We're having some problems lately with importing the component file inside the drivers.
So in order to avoid importing the whole component, let's move the `TextAreaTheme` to a different file, and then both the driver and the component files can import it.
The component file (`TextArea.tsx`) can also export it again.

jonathanadler:
@AndriiTsarenko 
The designers asked that we add an `overflow: hidden` on the textarea, by default, and then `overflow: auto` on `hover` & `focus`

PR #154, opened 2019-07-22
Improve generator

jonathanadler:
Add "is mobile" question


scripts/generate.js
```diff
@@ -0,0 +1,42 @@
+const prompts = require('prompts');
```
argshook: if you're ok with a little bash, and don't want `prompts` as `devDep`, it could be merely:

```sh
#!/usr/bin/env bash

read -p "Does the new component have a specific mobile design (y/n)?" CHOICE

case "$CHOICE" in
  y|Y )
    npm run generate:component:mobile
  n|N )
    npm run generate:component
  * )
    echo "please choose y or n";;
esac
```

package.json
```diff
@@ -48,7 +48,9 @@
     "compile-base-ui": "node ./.storybook/base-ui-compiler.js",
     "storybook:build": "build-storybook -s ./src/assets",
     "generate-stylable-index": "stc --srcDir=\"./dist/src/components\" --diagnostics --indexFile=index.st.css",
-    "generate": "wuf generate --codemods .wuf/generator/codemods --templates .wuf/generator/templates",
+    "generate": "node scripts/generate.js",
+    "generate:component": "wuf generate --codemods .wuf/generator/codemods --templates .wuf/generator/templates/desktop",
```
argshook: both generator templates now have `src/components` folders. you could remove them and have just component folder with `--output` flag:

```
wuf generate --codemods .wuf/generator/codemods --templates .wuf/generator/templates/desktop --output src/components,
```

it's not critical at all, just thinking maybe templates shouldn't include `src/components` as part of them

scripts/generate.js
jonathanadler: Yes!
I wanted to use bash instead of node for this, but didn't know how.

package.json
jonathanadler: Excellent!

scripts/generator.sh
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
+
+v='\xE2\x9c\x94'
```
argshook: @jonathanadler 	well if you're going to the lengths of this, then i wouldn't advise bash script. i thought you wanted just a yes/no prompt

jonathanadler: Well, it's already done 🤷‍♂ 
I prefer bash because with node the only option is to spawn children which I don't like to handle

jonathanadler: @argshook so what do you say?

argshook: this code is a little scary and could bite back in future but im not against it :)

i would probably not add any colors or the loop, just a simple `y/n` and that's it. But you're the maintainer, it's your homeland, you can do everything, the only limit is yourself (https://html5zombo.com/)

jonathanadler: 😂

PR #157, opened 2019-07-25
Render Pagination in compact mode for mobile devices

sdoomz:
This PR adds mobile mode support to `Pagination` component

screencast 2019-07-25 20-25-51

### DIFF
--- a/src/components/Pagination/Pagination.tsx
+++ b/src/components/Pagination/Pagination.tsx
@@ -20,18 +20,18 @@ export interface PaginationProps extends CorePaginationProps {}
 export const Pagination: React.FunctionComponent<PaginationProps> = props => {
   return (
     <TPAComponentsConsumer>
-      {({ mobile: mobileView, rtl }) => {
+      {({ mobile, rtl }) => {
         return (
           <CorePagination
-            {...styles('root', { mobileView, rtl }, props)}
+            {...styles('root', { mobile, rtl }, props)}
             previousLabel={ChevronLeft}
             nextLabel={ChevronRight}
             firstLabel={DoubleChevronLeft}
             lastLabel={DoubleChevronRight}
             rtl={rtl}
             showFirstLastNavButtons={props.totalPages > props.maxPagesToShow}
-            paginationMode={mobileView ? 'input' : 'pages'}
-            showInputModeTotalPages={mobileView}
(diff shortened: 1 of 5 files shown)


sdoomz:
@jonathanadler 

According to Search design we should limit font size in mobile layout to 14px. Should we apply this restriction inside `wix-ui-tpa` or it's up to consumer to define this behaviour? 
IMO, it should be inside `wix-ui-tpa`.

src/components/Pagination/Pagination.tsx
```diff
@@ -20,18 +20,18 @@ export interface PaginationProps extends CorePaginationProps {}
 export const Pagination: React.FunctionComponent<PaginationProps> = props => {
   return (
     <TPAComponentsConsumer>
-      {({ mobile: mobileView, rtl }) => {
+      {({ mobile, rtl }) => {
         return (
           <CorePagination
-            {...styles('root', { mobileView, rtl }, props)}
+            {...styles('root', { mobile, rtl }, props)}
```
jonathanadler: Theoretically, this is a breaking change (since consumers can manipulate the component's state using stylable).
However, I don't think this will be a problem since I don't think anyone is using this.
I'll check first.

jonathanadler:
> @jonathanadler
> 
> According to Search design we should limit font size in mobile layout to 14px. Should we apply this restriction inside `wix-ui-tpa` or it's up to consumer to define this behaviour?
> IMO, it should be inside `wix-ui-tpa`.

You're right!
This is also documented here:
https://zeroheight.com/7sjjzhgo2/p/7404ef/t/b1a9

sdoomz:
@jonathanadler You know, I have checked documentation again and didn't find a note about overriding font-size for mobile layout, in our design it's also `16px`. 

So I assume it's a valid behaviour and consumer can always override it.

src/components/Pagination/Pagination.tsx
sdoomz: Yes, you are right, but I didn't manage to find any other consumers except `Wix Search`

sdoomz: @jonathanadler build is green

sdoomz:
UPDATE: default font-size for mobile set to `14px`

PR #161, opened 2019-08-01
<Toast/>

nataliia-khitsan:
Toast (AKA Notification Banner)

Doc https://zeroheight.com/7sjjzhgo2/p/447d36/b/981089/t/075ef2


src/components/Toast/Toast.st.css
```diff
@@ -0,0 +1,60 @@
...
+    statusColor: #212121;
+}
+
+.root {
+    -st-states: rtl, mobile, skin(enum(success, error, status));
+    display: inline-flex;
+    color: #FFFFFF;
+    border-radius: 2px;
+    font-size: 14px;
+    font-family: "HelveticaNeueW01-55Roma";
```
giladsegal: You also need to provide fallbacks, in `wix-style-react` these are the fallbacks:
```
"HelveticaNeueW01-55Roma", "HelveticaNeueW02-55Roma", "HelveticaNeueW10-55Roma", "Helvetica Neue", "Helvetica", "Arial", "メイリオ, meiryo", "ヒラギノ角ゴ pro w3", "hiragino kaku gothic pro", "sans-serif"
```

src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,67 @@
...
+    const { skin, shouldShowCloseButton, children, ...rest } = this.props;
+
+    return (
+      <TPAComponentsConsumer>
+        {({ mobile, rtl }) => (
+          <div
+            {...styles('root', { mobile, rtl, skin }, rest)}
+            data-skin={skin}
+          >
+            <span className={styles.gapBeforeMessage}></span>
```
giladsegal: Can you remove the two `span` elements and apply padding to `.root` element?

src/components/Toast/helpers.ts
```diff
@@ -0,0 +1,16 @@
+import { TOAST_SKIN } from './Toast';
+
+export function getMessage(skin: TOAST_SKIN): string {
```
giladsegal: If you use the following code, when a new skin is added a compile error would occur (and then you would have to update this object) instead of having an empty string being returned in run time
```
const skinMessages: {[skin in keyof typeof TOAST_SKIN]: string} = {
  success: '24 photos were uploaded successfully.',
  error: 'These files exceed the upload limit.',
  status: 'Uploading...',
};
```

src/components/Toast/Toast.driver.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export const toastDriverFactory = (base: UniDriver): ToastDriver => {
+  const stylableUtil = new StylableUnidriverUtil(style);
+
+  const getCloseButton = async () => base.$('[data-hook="closeButton"]');
+  const getMessage = async () => base.$('[data-hook="message"]');
+
+  return {
+    ...baseUniDriverFactory(base),
+    isMobile: async () => stylableUtil.hasStyleState(base, 'mobile'),
```
giladsegal: You can add `data-mobile` like `data-skin` and then you can remove `StylableUnidriverUtil` completely

src/components/Toast/Toast.st.css
```diff
@@ -0,0 +1,60 @@
+:import {
+    -st-from: "../../common/formatters.st";
+    -st-named: color;
```
giladsegal: Is this unused?

src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,67 @@
...
+    onClose && onClose();
+  };
+
+  render() {
+    const { skin, shouldShowCloseButton, children, ...rest } = this.props;
+
+    return (
+      <TPAComponentsConsumer>
+        {({ mobile, rtl }) => (
+          <div
```
giladsegal: I think the root element should have `role="alert"` so it will create an `aria-live` region
(it indicates to screen readers that the text here important and they will watch for changes in this element)
You can read more about it in the [spec](http://w3c.github.io/aria-practices/#alert)

src/components/Toast/Toast.st.css
```diff
@@ -0,0 +1,60 @@
...
+.gapAfterMessage {
+    display: inline-block;
+    min-width: 12px;
+}
+
+.message {
+    margin: 16px 20px;
+    display: inline-block;
+    max-height: 40px;
+    line-height: 20px;
```
giladsegal: Should we add multiline ellipsis here if the text is longer then 2/3 lines (desktop / mobile)?
It isn't very easy but possible - what do you think?

src/components/Toast/Toast.tsx
jonathanadler: I told Nataliia to do this, so it's ok.
It's in order to easily support rtl and use `flex`'s native support for rtl

src/components/Toast/Toast.st.css
jonathanadler: only if there's a css solution for it.
The editor doesn't allow us to us javascript for things like this.

giladsegal: You can review an example [here](http://hackingui.com/front-end/a-pure-css-solution-for-multiline-text-truncation/) or [here](https://css-tricks.com/multi-line-truncation-with-pure-css/)
but it works for ltr, we need to test if it will work for rtl

src/components/Toast/Toast.tsx
jonathanadler: The `rtl` state can be removed from here

src/components/Toast/Toast.tsx
giladsegal: so we should use `role="presentation"` for these `span` elements (which tells the screen reader to ignore those elements)

nataliia-khitsan: @giladsegal Are you mean gapBefore and gapAfter? We can not remove it, because we have not symmetric left-right padding. So for saving correct rtl, we (I mean @jonathanadler and me) decided to add gap-element.

src/components/Toast/Toast.st.css
nataliia-khitsan: Yes, it might not be the most elegant solution, but possible. I can do this. But when I talk with @jonathanadler, he said just cut.

nataliia-khitsan: ok, i'll do this only with css

nataliia-khitsan:
@jonathanadler @giladsegal I've resolved the comments and ellipsis is implemented too. Can you review?

src/components/Toast/Toast.st.css
jonathanadler: Sorry I didn't reply, I'm not sure we can use these solutions because they change the styling of the text (justify).
I'll need to check with the designers about this.

jonathanadler: So yeah, I checked with the designers, and we shouldn't use this solution.
Sorry for not getting back to you earlier.
TPA's using this component should know not to add more than 2 rows of text in the toast.

nataliia-khitsan: reverted

src/components/Toast/CloseIconButton/CloseIconButton.tsx
```diff
@@ -0,0 +1,14 @@
+import * as React from 'react';
+import { IconButton } from '../../IconButton';
+import { ReactComponent as PlusIcon } from '../../../assets/icons/plus.svg';
```
jonathanadler: Please use this icon https://zeroheight.com/7sjjzhgo2/p/884c1f/i/3206562 instead of the Plus.

src/components/Toast/Toast.driver.ts
```diff
@@ -0,0 +1,55 @@
+import {
+  BaseUniDriver,
+  baseUniDriverFactory,
+} from 'wix-ui-test-utils/base-driver';
+import { UniDriver } from 'wix-ui-test-utils/unidriver';
+import { TOAST_SKIN } from './Toast';
```
jonathanadler: The reason the tests are failing, is because we're importing the component's file here.
This causes the protractor tests to import uncompiled files, and they don't know how to deal with it.
We need to move the `TOAST_SKIN` to a different file (maybe the `helpers`), and then use that file here and in `Toast.tsx`

src/assets/icons/Close.svg
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jonathanadler: Sorry for not mentioning earlier, but can you add this svg file to the `icons/index` file 😬 thanks!

src/components/Toast/README.md
```diff
@@ -0,0 +1,5 @@
+## Toast
+An implementation of an Toast for TPAs.
+Toasts are used to display important notifications or errors for the entire page.
+
+The Toast TPA implementations provides a few default fixed styles. It means none of them can be overridable.
```
jonathanadler: Maybe just add a 
```md
### How To Use
...
```
You can see the `Avatar/README` as an example

src/components/Toast/Toast.meta.tsx
```diff
@@ -0,0 +1,20 @@
...
+ * (foreground color: #ffffff, background color: #57c487, font size: 10.5pt (14px), font weight: normal).
+ * Expected contrast ratio of 4.5:1
+ */
+
+// import { Toast, TOAST_SKIN, ToastProps } from '.';
+// import Registry from '@ui-autotools/registry';
+//
+// const ToastMetadata = Registry.getComponentMetadata(Toast);
+// ToastMetadata.nonReactStrictModeCompliant = true;
+//
```
jonathanadler: I've investigated this a bit.
For now, you can uncomment this and add:
```js
ToastMetadata.nonA11yCompliant = true;
```
This should let the `a11y` tests for this component be ignored.
In the future, we might change the design to be more a11y compliant

src/components/Toast/Toast.st.css
```diff
@@ -0,0 +1,65 @@
...
+.root {
+    -st-states: rtl, mobile, shouldAnimate, isShown, skin(enum(success, error, status));
+    display: inline-flex;
+    color: #FFFFFF;
+    border-radius: 2px;
+    font-size: 14px;
+    font-family: "HelveticaNeueW01-65Medi", "Arial", "sans-serif";
+    box-shadow: 0 0 4px 0 rgba(0,0,0,0.10), 0 4px 8px 0 rgba(0,0,0,0.10);
+    opacity: 0;
+}
```
jonathanadler: I think we're missing here a transform also:
```css
transform: translateY(-50%);
```

src/components/Toast/Toast.st.css
```diff
@@ -0,0 +1,65 @@
...
+
+.root:isShown {
+    opacity: 1;
```
jonathanadler: And then add:
```css
transform: translateY(0);
```

src/components/Toast/Toast.st.css
```diff
@@ -0,0 +1,65 @@
...
+}
+
+.root:shouldAnimate {
+    transition: opacity 0.2s ease;
```
jonathanadler: And:
```css
transition: opacity 0.2s ease, transform 0.4s ease;
```

src/components/Toast/Toast.tsx
```diff
@@ -0,0 +1,62 @@
...
+      ...rest
+    } = this.props;
+
+    return (
+      <TPAComponentsConsumer>
+        {({ mobile, rtl }) => (
+          <div
+            {...styles(
+              'root',
+              { mobile, rtl, skin, shouldAnimate, isShown },
```
jonathanadler: I think we can remove the `rtl` state from here (and also the `.st.css` file)

PR #163, opened 2019-08-02
feat(ToolTip): error skin

AndriiTsarenko:


src/components/Tooltip/Tooltip.st.css
```diff
@@ -93,3 +94,13 @@
     opacity: 0;
     transition: transform 80ms linear, applyOpacity 80ms linear;
 }
+
+.root:skin(error)::popoverContent {
+    background-color: #F64D43;
```
jonathanadler: I think we can use the `errorColor` from `palette.st.css`, right?

src/components/Tooltip/Tooltip.st.css
```diff
@@ -93,3 +94,13 @@
...
+    padding: 5px 12px;
+    border: 1px solid rgba(255,255,255,0.25);
```
jonathanadler: btw, I'm not sure the padding and the border are necessary here (I'll check with the designers), even though it looks pretty good in the storybook

src/components/Tooltip/Tooltip.st.css
AndriiTsarenko: Yes. You're right.

src/components/Tooltip/Tooltip.st.css
AndriiTsarenko: Ok. I'll wait for the answer.

AndriiTsarenko: I've deleted padding.

src/components/Tooltip/docs/testProps.ts
```diff
@@ -0,0 +1,10 @@
+import { TooltipSkin } from '../TooltipEnums';
+
+export const testPropsList: Partial<any | { ['data-hook']: string }>[] = [
```
giladsegal: `Partial<any | { ['data-hook']: string }>[]` resolves to `Partial<any>[]` which can be expressed as `{[key:string]: any}[]`

src/components/Tooltip/Tooltip.tsx
```diff
@@ -17,15 +18,24 @@ function customArrow(placement: Placement, arrowProps) {
   );
 }
 
-export interface TooltipProps extends CoreTooltipProps {}
+export interface TooltipProps extends CoreTooltipProps {
+  skin?: TooltipSkin;
+}
+
+export interface TooltipDefaultProps {
```
giladsegal: I would go for
```
export type TooltipDefaultProps = Required<Pick<TooltipProps, 'skin'>>
```
So if you update `TooltipProps` you will get a compile error for the default props (forcing you to update it as well)

src/components/Tooltip/Tooltip.tsx
```diff
@@ -17,15 +18,24 @@ function customArrow(placement: Placement, arrowProps) {
   );
 }
 
-export interface TooltipProps extends CoreTooltipProps {}
+export interface TooltipProps extends CoreTooltipProps {
+  skin?: TooltipSkin;
```
giladsegal: If you would add a comment here it will update the storybook API tab

src/components/Tooltip/docs/testProps.ts
AndriiTsarenko: Initially I was using it like `Partial<TooltipProps | { ['data-hook']: string }>[]`. But I faced with the problem that I can't separate  `TooltipProps` and `jsx` syntax. `jsx` syntax is not supported in `e2e` so I put it as `any`.

AndriiTsarenko: I'll change it as you are telling me.

src/components/Tooltip/Tooltip.tsx
AndriiTsarenko: Done

src/components/Tooltip/Tooltip.tsx
AndriiTsarenko: Added.

src/components/Tooltip/Tooltip.st.css
jonathanadler: So the designers said we can also remove the border.

PR #166, opened 2019-08-09
feat: ToggleSwitch Component

pycm:


.wuf/testkits/definitions.js
```diff
@@ -14,7 +14,7 @@ module.exports = {
   LikeButton: { legacyDriver: true },
   Pagination: { legacyDriver: true },
   Text: { legacyDriver: true },
-  ToggleSwitch: { legacyDriver: true, noTestkit: true },
+  ToggleSwitch: { legacyDriver: false },
```
jonathanadler: I think we can remove this line altogether

src/components/ToggleSwitch/ToggleSwitch.driver.ts
```diff
@@ -0,0 +1,56 @@
+import {
+  BaseUniDriver,
+  baseUniDriverFactory,
+} from 'wix-ui-test-utils/base-driver';
+import { UniDriver } from 'wix-ui-test-utils/unidriver';
+import style from './ToggleSwitch.st.css';
```
jonathanadler: This should be removed, since it's not used

src/components/ToggleSwitch/ToggleSwitch.e2e.ts
```diff
@@ -0,0 +1,27 @@
...
+  const storyUrl = createStoryUrl({
+    kind: 'Components',
+    story: 'ToggleSwitch',
+    withExamples: true,
+  });
+  const dataHook = 'storybook-ToggleSwitch';
+
+  beforeEach(() => browser.get(storyUrl));
+
+  eyes.it('should render', async () => {
```
jonathanadler: If we don't have an additional test other than this one, then we can remove this file.
For visual tests, we prefer using only `.visual.tsx` tests

src/components/ToggleSwitch/ToggleSwitch.spec.tsx
```diff
@@ -1,16 +1,29 @@
...
+    const driver = createDriver(<ToggleSwitch disabled />);
+
+    expect(await driver.isDisabled()).toBeTruthy();
+  });
+
+  it('should change state', async () => {
+    const driver = createDriver(<ToggleSwitch />);
 
-  afterEach(() => wrapper.detach());
+    await driver.click();
```
jonathanadler: We should probably add 
```
expect(await driver.isChecked()).toBeFalsy();
```
before clicking

src/components/ToggleSwitch/ToggleSwitch.st.css
```diff
@@ -1,154 +1,75 @@
 :import {
     -st-from: "../../common/formatters.st";
-    -st-named: applyStyleMethod, applyOpacity;
+    -st-named: color, fallback, font, applyStyleMethod, applyOpacity;
```
jonathanadler: There's no need to use `applyStyleMethod` in this file.
You can just use: `color(color-3)`, etc.
`applyStyleMethod` was created only for edge cases where the `color` name is taken by a variable (or something like that)

src/components/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -0,0 +1,36 @@
...
+  static contextType = TPAComponentsContext;
+  static displayName = 'ToggleSwitch';
+  static defaultProps = { ...CoreToggleSwitch.defaultProps };
+
+  render() {
+    const { checked, disabled, ...rest } = this.props;
+
+    return (
+      <TPAComponentsConsumer>
+        {({ mobile }) => (
```
jonathanadler: We're not using the `mobile` state, so we can remove `TPAComponsntsConsumer` and the `mobile` render prop

src/components/ToggleSwitch/docs/examples.st.css
```diff
@@ -0,0 +1,3 @@
+.root {
```
jonathanadler: I guess this file isn't needed

src/components/ToggleSwitch/docs/index.story.tsx
```diff
@@ -0,0 +1,63 @@
...
+          description(Readme),
+          importExample(
+            "import { ToggleSwitch } from 'wix-ui-tpa/dist/src/components/ToggleSwitch';",
+          ),
+
+          divider(),
+
+          title('Examples'),
+
+          <Examples key="toggle-switch-examples" />,
```
jonathanadler: Please use the `settingsPanel` util we have, instead of this.

You can look at how the other recently created `index.story.tsx` files are structured as an example (AvatarGroup, Counter, Ratings, etc.)

PR #169, opened 2019-08-12
Create DotNavigation component

charmingelle:


src/components/DotNavigation/DotNavigation.e2e.ts
```diff
@@ -0,0 +1,64 @@
...
+  const storyUrl = createStoryUrl({
+    kind: 'Components',
+    story: 'DotNavigation',
+    withExamples: true,
+  });
+  const dataHook = 'storybook-DotNavigation';
+
+  beforeEach(() => browser.get(storyUrl));
+
+  eyes.it(
```
jonathanadler: No need to use `eyes.it` in these tests, just regular `it`.
We prefer to have visual tests in the `*.visual.tsx` file.

src/components/DotNavigation/DotNavigation.meta.tsx
```diff
@@ -0,0 +1,10 @@
+import { DotNavigation } from '.';
+import Registry from '@ui-autotools/registry';
+
+const DotNavigationMetadata = Registry.getComponentMetadata(DotNavigation);
+DotNavigationMetadata.nonReactStrictModeCompliant = true;
+
+DotNavigationMetadata.addSim({
+  title: 'render',
+  props: {},
```
jonathanadler: We should probably add some props for this test (this tests ssr + hydration and also a11y).
Maybe a length and currentIndex?

src/components/DotNavigation/DotNavigation.spec.tsx
```diff
@@ -0,0 +1,330 @@
...
+
+    it('should render 4 dots with length equal to 4', async () => {
+      const driver = createDriver(<DotNavigation length={4} />);
+      const dots = driver.getDots();
+
+      expect(await dots.count()).toBe(4);
+    });
+
+    it('should render 9 dots with length greater than 5', async () => {
+      const driver = createDriver(<DotNavigation length={6} />);
```
jonathanadler: I feel like i'm missing something here.
The length passed to the component is `6` but the result is `9`?

src/components/DotNavigation/DotNavigation.st.css
```diff
@@ -0,0 +1,130 @@
...
+  DefaultSelectedBorderColor: color-1;
+  DefaultSelectedDotColorLight: color-1;
+  DefaultSelectedBorderColorLight: color-5;
+}
+
+:vars {
+  /*Overrides*/
+  DotNavigationSelectedDotColor: --overridable;
+  DotNavigationSelectedBorderColor: --overridable;
+}
```
jonathanadler: These two variables are sort of the components style "API".
Consumers will use these variables in order to override the colors.
So let's have a bit shorter names.
I think we can use `SelectedDotColor` and `SelectedBorderColor`.

src/components/DotNavigation/DotNavigation.st.css
```diff
@@ -0,0 +1,130 @@
+:import {
+  -st-from: '../../common/formatters.st';
+  -st-named: color, fallback, value;
+}
+
+:vars {
```
jonathanadler: We're using the `wix-ui-core/RadioButton`, so we should also import its style here, and use
```-st-extend: RadioButton` in the correct class.

File: src/components/DotNavigation/DotNavigation.visual.tsx [hunk 6]
```diff
@@ -0,0 +1,123 @@
...
+class DotNavigationVisual extends React.Component<any> {
+  static defaultProps = {
+    mobile: false,
+  };
+
+  render() {
+    const { mobile } = this.props;
+
+    return (
+      <TPAComponentsProvider value={{ mobile }}>
```
jonathanadler: We can remove all references to the `TPAComponentsProvider` and `TPAComponentsConsumer` and the `mobile` state, because I think we don't have any special style for it

src/components/DotNavigation/DotNavigation.tsx
```diff
@@ -0,0 +1,400 @@
...
+
+  render = () => {
+    const { length, showBorder, theme, ...rest } = this.props;
+
+    if (!isNumber(length) || length <= 0) {
+      return null;
+    }
+
+    return (
+      <TPAComponentsConsumer>
```
jonathanadler: We can remove the `TPAComponentsConsumer` from here

src/components/DotNavigation/DotNavigation.tsx
```diff
@@ -0,0 +1,400 @@
...
+            )}
+            {...this._getDataAttributes()}
+          >
+            {length <= MAX_SHORT_LIST_LENGTH
+              ? this.renderShortList()
+              : this.renderLongList()}
+          </div>
+        )}
+      </TPAComponentsConsumer>
+    );
```
jonathanadler: A general note about using `...styles(..)`.
We usually prefer using this **only** on the root element.
Stylable can introduce some bugs when using it like this.

If it's used for states that are for sub-component, then it's better we just toggle classes on/off (I think we import `classnames` in the project)

src/components/DotNavigation/DotNavigation.e2e.ts
charmingelle: Sure. The thing is that we cannot take screenshots of these two cases (dot hover and dot click) by just passing props to the component as we do in  <code>*.visual.tsx</code>. If we don't need such screenshots at all, I'll remove them.

src/components/DotNavigation/DotNavigation.spec.tsx
charmingelle: That's true. If length > 5, 9 dots are rendered:
3 middle dots,
2 small dots for moving back and forward,
2 tiny dots for going to start and to end,
2 additional tiny dots at the beginning and at the end for correct animation.

src/components/DotNavigation/DotNavigation.st.css
charmingelle: The thing is that wix-ui-core/RadioButton component accepts two props: checkedIcon and uncheckedIcon of React.ReactNode types. These are components with their own styles. As far as I can see, nothing mentioned in wix-ui-core/RadioButton.st.css can be used in our case.

charmingelle: The thing is that <code>wix-ui-core/RadioButton</code> component accepts two props: checkedIcon and uncheckedIcon of React.ReactNode types. These are components with their own styles. As far as I can see, nothing mentioned in <code>wix-ui-core/RadioButton.st.css</code> can be used in our case.

src/components/DotNavigation/DotNavigation.st.css
charmingelle: Done)

src/components/DotNavigation/DotNavigation.meta.tsx
charmingelle: Done)

src/components/DotNavigation/DotNavigation.visual.tsx
charmingelle: Done)

src/components/DotNavigation/DotNavigation.tsx
charmingelle: Done)

src/components/DotNavigation/DotNavigation.tsx
charmingelle: I thought that using states is the main stylable feature. Anyway, I replaces them with toggling classes with classnames)

src/components/DotNavigation/DotNavigation.spec.tsx
jonathanadler: I see, so maybe we should have a different name for this.
The driver's methods will also be used by consumers (via the testkit), it might confuse them.
Maybe we should have a distinction between `getVisibleDots` and `getRenderedDots`.
I'm not sure the naming is good, but something like that.

src/components/DotNavigation/DotNavigation.st.css
jonathanadler: I have to say that I really agree with you, but as a convention we prefer adding the `extend` rule.
There might be changes in the future that we would like to use.

src/components/DotNavigation/DotNavigation.tsx
jonathanadler: That's true, however, we still encounter strange behaviours when we use "nested" states.
Thanks for the change 💪

src/components/DotNavigation/DotNavigation.e2e.ts
jonathanadler: That's a very good point.
I want to add the ability to be able to do these things inside the `*.visual.tsx` tests (which I hope will be done next week).
But you're right, let's leave it like this for now, and when I'll add the capability, will also merge all the tests from eyes.it to visual.

src/components/DotNavigation/DotNavigation.spec.tsx
charmingelle: As component logic changed, this test is no longer valid.

src/components/DotNavigation/DotNavigation.st.css
charmingelle: Done)

src/components/DotNavigation/DotNavigation.st.css
```diff
@@ -0,0 +1,133 @@
...
+}
+
+.dot {
+  width: value(DotSize);
+  height: value(DotSize);
+  border-radius: 100%;
+  background-color: value(CheckedDotColor);
+  opacity: 0.4;
+  transition: background-color 0.2s ease, box-shadow 0.2s ease, width 0.2s ease, height 0.2s ease;
+}
```
jonathanadler: Comment from the designer:
The dot's background should also change with an animation (the same timing as the moving animation)

src/components/DotNavigation/docs/index.story.tsx
```diff
@@ -0,0 +1,115 @@
...
+        { title: 'Playground', sections: [playground()] },
+        {
+          title: 'Settings Panel',
+          sections: [
+            settingsPanel({
+              example: (
+                <DotNavigationExtendedExample
+                  length={7}
+                  showBorder
+                  theme={Theme.Light}
```
jonathanadler: Since the background of the example is white, it might be better to start with the dark theme.

src/components/DotNavigation/docs/DotNavigationExtendedExample.tsx
```diff
@@ -0,0 +1,9 @@
+import * as React from 'react';
+import extendedStyles from './DotNavigationExtendedExample.st.css';
+import { DotNavigation, DotNavigationProps } from '../DotNavigation';
+
+export class DotNavigationExtendedExample extends React.Component<
```
jonathanadler: Let's add a `currentIndex` state prop, and change it with the `onSelect`, so that we can play with the example

src/components/DotNavigation/DotNavigation.tsx
```diff
@@ -0,0 +1,280 @@
...
+  renderLongVersion = () => (
+    <div
+      className={classNames(styles.radioGroup, {
+        [styles.back]: this.state.animation === Animation.Back,
+        [styles.forward]: this.state.animation === Animation.Forward,
+      })}
+      key={this.state.longListKey}
+    >
+      {this.renderFakeRadio('fake-start')}
+      {!isNumber(this.props.currentIndex) || this.props.currentIndex < 2
```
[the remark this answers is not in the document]
jonathanadler: And by the way, in general I think it can be something like:
```js
const NUM_OF_NAV_DOTS = 2; // for the small and medium dot (the name should be something better than this)
this.props.currentIndex < MAX_SHORT_LIST_LENGTH - NUM_OF_NAV_DOTS;
...
this.props.currentIndex >= this.props.length - (MAX_SHORT_LIST_LENGTH - NUM_OF_NAV_DOTS);
```

src/components/DotNavigation/DotNavigation.st.css
charmingelle: Done) But I also left transition here for hovering over the dot.

src/components/DotNavigation/DotNavigation.tsx
```diff
@@ -0,0 +1,280 @@
...
+    nextProps.length <= MAX_SHORT_LIST_LENGTH
+      ? prevState
+      : DotNavigation.getLongListAnimation(nextProps, prevState);
+
+  callOnSelect = (index: number) => () => this.props.onSelect(index);
+
+  getRadio = ({ key, checked, size, fakeRadio }: RadioProps) => (
+    <div className={classNames(styles.radio)}>
+      <div
+        key={key}
```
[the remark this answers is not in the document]
charmingelle: Do you mean this callOnSelect method? It is for renderRadioButton call to look better, like this:

```this.renderRadioButton({
          key: index,
          checked: index === this.props.currentIndex,
          size: Size.Normal,
          onChange: this.callOnSelect(index),
          ariaLabel: `${this.props['aria-label']} ${index}`,
        });```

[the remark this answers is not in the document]
Comment (charmingelle) [reply src/components/DotNavigation/DotNavigation.tsx]: Do you mean this callOnSelect method? It is for renderRadioButton call to look better, like this:
```
this.renderRadioButton({
  key: index,
  checked: index === this.props.currentIndex,
  size: Size.Normal,
  onChange: this.callOnSelect(index),
  ariaLabel: `${this.props['aria-label']} ${index}`,
});
```

File: src/components/DotNavigation/docs/DotNavigationExtendedExample.tsx (hunk 11 above)
Comment (charmingelle) [reply src/components/DotNavigation/docs/DotNavigationExtendedExample.tsx]: Done)

File: src/components/DotNavigation/docs/index.story.tsx (hunk 10 above)
Comment (charmingelle) [reply src/components/DotNavigation/docs/index.story.tsx]: The thing is that default SelectedBorderColor for the dark theme is color-1, it is the same as the color of every palette background, so you cannot see the border by default unless you change its color in the palette.

If you are ok with this, I'll remove the light theme from here.

File: src/components/DotNavigation/DotNavigation.tsx [hunk 14]
```diff
@@ -0,0 +1,280 @@
...
+
+  static displayName = 'DotNavigation';
+
+  static defaultProps: DotNavigationProps = {
+    currentIndex: 0,
+    length: MAX_SHORT_LIST_LENGTH,
+    onSelect(index: number) {},
+    showBorder: false,
+    theme: Theme.Dark,
+    ariaLabel: 'Dot Navigation',
```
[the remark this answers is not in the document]
charmingelle: Done)

src/components/DotNavigation/DotNavigation.tsx
```diff
@@ -0,0 +1,280 @@
...
+
+interface RadioButtonProps {
+  key: number | string;
+  checked?: boolean;
+  size: 'normal' | 'small' | 'tiny';
+  onChange?: any;
+  ariaLabel: null | string;
+}
+
+const isNumber = (value: any) => typeof value === 'number' && !isNaN(value);
```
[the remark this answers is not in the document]
charmingelle: Done)

src/components/DotNavigation/DotNavigation.tsx
```diff
@@ -0,0 +1,280 @@
...
+        ? this.renderStartList()
+        : this.props.currentIndex >= this.props.length - 2
+        ? this.renderEndList()
+        : this.renderMiddleList()}
+      {this.renderFakeRadio('fake-end')}
+    </div>
+  );
+
+  renderShortVersion = () =>
+    new Array(this.props.length).fill(null).map((el, index) =>
```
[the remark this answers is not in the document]
charmingelle: Replaced the fill method with other stuff in two places.

src/components/DotNavigation/DotNavigation.tsx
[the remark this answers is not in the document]
charmingelle: Done)

src/components/DotNavigation/DotNavigation.tsx
```diff
@@ -0,0 +1,280 @@
...
+        }
+      : prevState;
+  };
+
+  static getDerivedStateFromProps = (
+    nextProps: DotNavigationProps,
+    prevState: DotNavigationState,
+  ) =>
+    nextProps.length <= MAX_SHORT_LIST_LENGTH
+      ? prevState
```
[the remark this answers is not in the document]
charmingelle: Done)

src/components/DotNavigation/DotNavigation.tsx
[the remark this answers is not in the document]
charmingelle: If I need to render additional dots for correct animation at the beginning and in the end, I use getRadio method call with key prop in it. The keys are "'fake-start'" and "fake-end". As these elements are parts of the list, they should have keys.

If I call getRadio method to get checkedIcon and uncheckedIcon props of wix-ui-code RadioButton, I don't pass the key prop.

src/components/DotNavigation/DotNavigation.tsx
```diff
@@ -0,0 +1,294 @@
...
+
+  renderRadioButton = ({
+    key,
+    checked,
+    size,
+    onChange,
+    ariaLabel,
+  }: RadioButtonProps) => (
+    <RadioButton
+      key={key}
```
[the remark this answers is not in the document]
charmingelle: The focus didn't change not because of this key but because of another one. I removed it and rewrote the animation logic slightly.

src/components/DotNavigation/DotNavigation.tsx
```diff
@@ -0,0 +1,294 @@
...
+          [styles.light]: this.props.theme === Theme.Light,
+          [styles.checked]: checked,
+        })}
+        data-hook={fakeRadio ? undefined : DotNavigationDataHooks.Dot}
+      />
+    </div>
+  );
+
+  renderFakeRadio = (key: string) =>
+    this.getRadio({ key, checked: false, size: Size.Tiny, fakeRadio: true });
```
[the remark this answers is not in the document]
charmingelle: I added tabIndex={-1} and aria-hidden to HTML element, which represents a fake dot. There is no need to change <code>wix-ui-core</code> for this as I don't use their <code>RadioButton</code> for fake dots.

I tried though. The problem was that <code>RadioButton</code> had an optional prop aria-label, which I didn't pass because the dots were fake. But accessibility tests failed because this prop was absent. So I decided not to use this component at all. To be honest, we don't need any of its functionality for fake dots as they are only for visual animation purposes.

jonathanadler:
Great work!
In order to fix the "loop" issue we have when using arrow keys, I think we just need to add a callback for `onkeydown` and `preventDefault` on the most left/right dots.

src/components/DotNavigation/DotNavigation.tsx
```diff
@@ -0,0 +1,327 @@
...
+  endKeyDownHandler = (event: RadioButtonKeyDownEvent) =>
+    (event.key === Keys.ArrowRight || event.key === Keys.ArrowDown) &&
+    event.preventDefault();
+
+  getKeyDownHandler = (value: number) =>
+    value === 0
+      ? this.startKeyDownHandler
+      : value === this.props.length - 1
+      ? this.endKeyDownHandler
+      : noop;
```
jonathanadler: It's better to set an `undefined` instead of `noop`.
`noop` will add an event listener, and `undefined` wont.

src/components/DotNavigation/DotNavigation.tsx
```diff
@@ -0,0 +1,327 @@
...
+      aria-label={ariaLabel}
+      {...{ className: styles.extendedRadioButton }}
+    />
+  );
+
+  callOnSelect = (index: number) => () => this.props.onSelect(index);
+
+  startKeyDownHandler = (event: RadioButtonKeyDownEvent) =>
+    (event.key === Keys.ArrowLeft || event.key === Keys.ArrowUp) &&
+    event.preventDefault();
```
jonathanadler: This isn't working for some reason.
Getting `Uncaught TypeError: event.preventDefault is not a function` in the storybook

src/components/DotNavigation/DotNavigation.tsx
```diff
@@ -0,0 +1,327 @@
...
+
+    for (let index = 0; index < this.props.length; index++) {
+      shortVersion = [
+        ...shortVersion,
+        this.renderRadioButton({
+          key: index,
+          checked: index === this.props.currentIndex,
+          size: Size.Normal,
+          onChange: this.callOnSelect(index),
+          onKeyDown: noop,
```
jonathanadler: Same comment about `noop` here

src/components/DotNavigation/DotNavigation.tsx
charmingelle: Re-installed node modules, fixed the issue)

src/components/DotNavigation/DotNavigation.tsx
charmingelle: Done) I received an error that undefined cannot be passed instead of a function before, re-installed node modules, fixed.

src/components/DotNavigation/DotNavigation.tsx
charmingelle: Done)

Issue #103, opened 2019-06-11
<Card/> component looks broken in storybook


nktssh:
https://wix-wix-ui-tpa.surge.sh/?path=/story/components--card

jonathanadler:
@nktssh Can you describe what seems to be broken?
It might have been a momentary issue.

nktssh:
<img width="1285" alt="Crs looks broken" src="https://user-images.githubusercontent.com/1872256/61062785-82575a80-a407-11e9-92ec-5c56063caa04.png">

jonathanadler:
@nktssh Does this persist if you refresh?

nktssh:
@jonathanadler yes

jonathanadler:
@nktssh 
I wasn't able to reproduce this, however, we are aware this is happening.
It seems to be a race condition that's causing this, and not a bug in the components.
We will try to find and fix the root cause of this.
Thanks!

PR #170, opened 2019-08-13
[WIP] ActionsMenu

nktssh:


src/components/ActionsMenu/ActionsMenu.driver.ts
```diff
@@ -0,0 +1,21 @@
...
+import style from './ActionsMenu.st.css';
+
+export interface ActionsMenuDriver extends BaseUniDriver {
+  isMobile(): Promise<boolean>;
+}
+
+export const actionsMenuDriverFactory = (
+  base: UniDriver,
+): ActionsMenuDriver => {
+  const stylableUtil = new StylableUnidriverUtil(style);
```
giladsegal: We usually avoid using `StylableUnidriverUtil` because it causes problems in e2e tests
You can add a `data-` attribute to your component and verify using this attribute

src/components/ActionsMenu/ActionsMenu.st.css
```diff
@@ -0,0 +1,49 @@
...
+    border-width: value(menuBorderWidth);
+    border-color: value(menuBorderColor);
+}
+
+.root:mobile {
+    width: 100%;
+}
+
+.root:alignment(left) {
+    text-align: left;
```
giladsegal: This isn't working as expected in storybook I think, maybe you should use `direction: rtl / ltr` or `flex`

src/components/ActionsMenu/ActionsMenu.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+export enum Alignment {
+  left = 'left',
+  center = 'center',
+  right = 'right',
+}
+
+export interface ActionsMenuProps {
+  alignment?: Alignment;
+  onBlur?(): void;
```
giladsegal: Is it unused because it is a WIP?

src/components/ActionsMenu/ActionsMenu.tsx
```diff
@@ -0,0 +1,44 @@
+import * as React from 'react';
+import styles from './ActionsMenu.st.css';
+import { TPAComponentsConsumer } from '../TPAComponentsConfig';
+import { ActionsMenuItem } from './Item/ActionsMenuItem';
+import { ActionsMenuDivider } from './Divider/Divider';
```
giladsegal: You should import from `./Divider`
This way we can change the internal folder structure more easily

src/components/ActionsMenu/Divider/Divider.st.css
```diff
@@ -0,0 +1,15 @@
+:import {
```
giladsegal: The library already has a `Divider` component, can you use it instead?

src/components/ActionsMenu/Item/ActionsMenuItem.st.css
```diff
@@ -0,0 +1,73 @@
+:import {
+    -st-from: "../../common/formatters.st";
+    -st-named: color, fallback;
+}
+
+
+:vars {
+    /*Defaults*/
+    DefaultItemFont: 16px Avenir;
```
giladsegal: You should use wix font stack

src/components/ActionsMenu/Item/ActionsMenuItem.st.css
```diff
@@ -0,0 +1,73 @@
...
+.root:hover {
+    background-color: rgba(0,0,0, 0.06);
+}
+
+.root:disabled {
+    opacity: 0.3;
+}
+
+.root:disabled:hover {
+    background-color: inherit;
```
giladsegal: I think you should also set `cursor: not-allowed`

src/components/ActionsMenu/Item/ActionsMenuItem.tsx
```diff
@@ -0,0 +1,57 @@
...
+    } = this.props;
+
+    return (
+      <TPAComponentsConsumer>
+        {({ mobile }) => (
+          <div
+            key={content}
+            {...styles(
+              styles.root,
+              { mobile, disabled, withIcon: !!prefixIcon },
```
giladsegal: Do you really need `withIcon` state?

src/components/ActionsMenu/Item/ActionsMenuItem.tsx
```diff
@@ -0,0 +1,57 @@
...
+      subtitle,
+      disabled,
+      onClick,
+      ...rest
+    } = this.props;
+
+    return (
+      <TPAComponentsConsumer>
+        {({ mobile }) => (
+          <div
```
giladsegal: Please refer to http://w3c.github.io/aria-practices/#menu for accessibility 
I think you shouldn't use `aria-pressed` as it is used for toggle buttons mostly, you should use `role="menu"`, set `aria-disabled` for disabled items, use a `ul` tag and more

src/components/ActionsMenu/Item/ActionsMenuItem.st.css
```diff
@@ -0,0 +1,73 @@
+:import {
```
giladsegal: It looks like the styles doesn't completely match zeroheight, can you please notify me once you are done with the styling?

src/components/ActionsMenu/ActionsMenu.e2e.ts
```diff
@@ -0,0 +1,30 @@
+import { browser } from 'protractor';
+import {
+  createStoryUrl,
+  waitForVisibilityOf,
+} from 'wix-ui-test-utils/protractor';
+import { actionsMenuTestkitFactory } from '../../testkit/protractor';
+
+/**
+ * For tests containing interactions.
+ * Can be removed if not used.
```
jonathanadler: This file can be removed if there are no special cases to test, that require a real dom.
Just to see that it is rendered in a browser is done in the visual tests.

src/components/ActionsMenu/ActionsMenu.tsx
jonathanadler: I actually told @nataliia-khitsan to add the onblur callback.
I thought it might be needed for a11y, but i'm not sure anymore.
We can remove it for now.

src/components/ActionsMenu/Item/ActionsMenuItem.st.css
jonathanadler: See `Text.st.css` as an example

src/components/ActionsMenu/Item/ActionsMenuItem.st.css
jonathanadler: @giladsegal i'm not sure changing the cursor is necessary.
I don't see it in the spec.
But we can check with the designers

src/components/ActionsMenu/ActionsMenu.driver.ts
nktssh: Fixed.
It was basically code from generator, maybe we need to change it also?

src/components/ActionsMenu/Item/ActionsMenuItem.st.css
nktssh: @jonathanadler any updates on this?

jonathanadler: No need to change cursor

src/components/ActionsMenu/ActionsMenu.st.css
jonathanadler: As @giladsegal wrote, this isn't working as excepted.
However, using `text-align` is fine in this case, since this is the component's alignment, and not the text-direction.
Try to solve why this isn't working correctly.

jonathanadler: I think you need to import the `ActionsMenuItem.st.css` file here, and pass the alignment to the component.

nktssh:
@jonathanadler alignment is working now. I've tried different approaches and it seems only this one fulfill the specs (e.g. entire row should be clickable and has custom `:hover` state)

Regarding styling on screenshot attached and in story itself - I've asked previously https://github.com/wix/wix-ui-tpa/issues/103
and got an answer 
> we are aware that this is happening
> it seems to be a race condition inside the story itself that's causing this, and not a bug in the component.
> We will try to find and fix the root cause of this.

Also, I'm not really sure how to test editor overrides (like font and colors) properly - is there an example of it?

Please, point me out if something need to be tuned and in which way! 🙏

src/components/ActionsMenu/Item/ActionsMenuItem.tsx
nktssh: Add mentioned comments addressed, but I'm not sure what's behind "more". 
Please ping me if something else need to be changed

src/components/ActionsMenu/ActionsMenu.visual.tsx
```diff
@@ -0,0 +1,122 @@
...
+            subtitle: 'Subtitle',
+            prefixIcon: <ShareIcon />,
+          }),
+        ],
+      },
+      {
+        it: 'mobile',
+        props: {
+          mobile: true,
+        },
```
jonathanadler: We will have to add children to the mobile test, otherwise it will just be empty and the visual test will be redundant.

src/components/ActionsMenu/ActionsMenu.st.css
```diff
@@ -0,0 +1,59 @@
...
+.item {
+    -st-extends: ActionMenuItem;
+}
+
+.root:mobile {
+    width: 100%;
+}
+
+.root:alignment(left) .item {
+    justify-content: flex-start;
```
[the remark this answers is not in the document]
nktssh: `Text-align` is fine for text itself but not working with images. 
I've done this by passing alignment down to `Item` and used absolute positioning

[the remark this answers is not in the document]
nktssh: I've changed code to avoid passing alignment, so now it looks better 🙂

jonathanadler:
I see that you've fixed the tests.
That's great 👍 
But there are still problems with the styling of the components.
Please go over the design spec (here: https://zeroheight.com/7sjjzhgo2/p/05a744/b/46537f/t/53af3f) and once it's done send me a request for review

src/components/ActionsMenu/ActionsMenu.tsx
```diff
@@ -0,0 +1,60 @@
...
+    alignment: Alignment.left,
+  };
+
+  render() {
+    const { alignment, children, ...rest } = this.props;
+    return (
+      <TPAComponentsConsumer>
+        {({ mobile }) => (
+          <ul
+            key={alignment}
```
jonathanadler: I think the `key` here isn't necessary.

src/components/ActionsMenu/ActionsMenu.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+export interface ActionsMenuProps {
+  alignment?: Alignment;
+}
+
+interface DefaultProps {
+  alignment: Alignment;
+}
+
+function getDataAttributes(mobile: boolean) {
```
jonathanadler: I think it's better to have this function as a class method, no?

src/components/ActionsMenu/ActionsMenu.st.css
```diff
@@ -0,0 +1,65 @@
...
+    padding: 12px 0;
+    max-width: 320px;
+    background-color: value(menuBackgroundColor);
+    border-style: solid;
+    border-width: value(menuBorderWidth);
+    border-color: value(menuBorderColor);
+}
+
+.item {
+    -st-extends: ActionMenuItem;
```
jonathanadler: I don't think this is needed if we're not overriding anything here.

src/components/ActionsMenu/ActionsMenu.tsx
```diff
@@ -0,0 +1,60 @@
...
+            key={alignment}
+            {...styles('root', { mobile, alignment }, rest)}
+            {...getDataAttributes(mobile)}
+            role="menu"
+            tabIndex={-1}
+          >
+            {React.Children.map(children, (child: React.ReactElement) =>
+              child.type === ActionsMenuItem
+                ? React.cloneElement(child, { className: styles.item })
+                : React.cloneElement(child, { className: styles.divider }),
```
jonathanadler: I actually think that passing the classNames aren't really necessary here.
I think the static Divider can be a component that wraps the `wix-ui-tpa/Divider`, and adds the necessary padding.

src/components/ActionsMenu/ActionsMenu.visual.tsx
```diff
@@ -0,0 +1,122 @@
...
+            content: 'item 1',
+          }),
+          generateItem({
+            content: 'item 3',
+            subtitle: 'Subtitle 2',
+          }),
+        ],
+      },
+    ],
+  },
```
jonathanadler: Let's add tests for all the different alignments: `left`, `center` & `right`.

src/components/ActionsMenu/Item/ActionsMenuItem.st.css
```diff
@@ -0,0 +1,77 @@
...
+    -st-states: disabled, mobile;
+    height: 40px;
+    padding: 8px 24px;
+    display: flex;
+    cursor: pointer;
+    position: relative;
+}
+
+.root:hover {
+    background-color: rgba(0,0,0, 0.06);
```
jonathanadler: This should be used with `color-5`:
```
applyOpacity(color(color-5), 0.06);
```

src/components/ActionsMenu/Item/ActionsMenuItem.st.css
```diff
@@ -0,0 +1,77 @@
...
+.root:disabled:hover {
+    background-color: inherit;
+}
+
+.root .text {
+    display: flex;
+    flex-direction: column;
+    color: color(color-5);
+    width: 100%;
+}
```
jonathanadler: Theoretically, a more "correct stylable" way to write this is by extending the `Text` component:
```css
:import {
  -st-from: '../../Text/Text.st.css';
  -st-default: Text;
}

.text {
  -st-extends: Text;
  -st-mixin: Text(
    MainTextColor color(color-5)
  );
  ...
}
```
(by the way, the `.root` isn't needed here, stylable knows to deal with it)

src/components/ActionsMenu/Item/ActionsMenuItem.st.css
```diff
@@ -0,0 +1,77 @@
...
+
+.root:mobile .content {
+    font: fallback(font(value(ItemFont)), font(value(ItemFontMobile)));
+}
+
+.root .subtitle {
+    font: fallback(font(value(ItemSubtitleFont)), font(value(ItemSubtitleFontDesktop)));
+    color: applyOpacity(color(color-5), 0.6);
+    width: 100%;
```
jonathanadler: Same as I've written above 👆 

```css
.subtitle {
  -st-extends: Text;
  -st-mixin: Text(
    MainTextColor applyOpacity(color(color-5), 0.6),
    MainTextFont fallback(font(value(ItemSubtitleFont)), font(value(ItemSubtitleFontDesktop)))
  );
  ...
}
```

src/components/ActionsMenu/Item/ActionsMenuItem.tsx
```diff
@@ -0,0 +1,58 @@
...
+    return (
+      <TPAComponentsConsumer>
+        {({ mobile }) => (
+          <li
+            key={content}
+            {...styles('root', { mobile, disabled }, rest)}
+            role="menuitem"
+            tabIndex={-1}
+            aria-disabled={disabled}
+            onClick={disabled ? noop : onClick}
```
jonathanadler: No need to use `noop`.
It's better to pass `undefined` so that no function will be attached to the event.

src/components/ActionsMenu/README.md
```diff
@@ -0,0 +1,25 @@
...
+Child of `ActionsMenu` must be either `Divider` or `ActionsMenuItem`, see Examples section for details.
+
+
+The `ActionsMenu` implementation provides a few default overridable styles. By default site theme is used.
+
+## Theme properties (all are optional)
+| propName              | propType | description                                                |
+|-----------------------|----------|------------------------------------------------------------|
+| MenuBackgroundColor   | string   | The color of the background instead of the default color   |
+| MenuBorderWidth       | string   | The color of the menu width instead of the default one     |
```
jonathanadler: It should be:
`The width of the border...`

src/components/ActionsMenu/dataHooks.ts
```diff
@@ -0,0 +1,3 @@
+export enum ACTIONS_MENU__DATA_KEYS {
```
jonathanadler: A bit picky, but we don't use `__` in our other `dataHooks` files.
So you can use just one `_`, for consistency.

src/components/ActionsMenu/ActionsMenu.st.css
```diff
@@ -0,0 +1,65 @@
...
+    margin: 10px 13px;
+    width: auto;
+}
+
+.root:mobile {
+    width: 100%;
+}
+
+.root:alignment(left) .item::container {
+    left: 24px;
```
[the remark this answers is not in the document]
nktssh: to imitate the padding, since we absolutely position here

src/components/ActionsMenu/ActionsMenu.tsx
nktssh: now it has no deps from props or class instance, so I've put it aside.
Could do it as static method though

src/components/ActionsMenu/ActionsMenu.st.css
nktssh: without it alignment would not work

src/components/ActionsMenu/Item/ActionsMenuItem.st.css
nktssh: ok, will use `Text`
but in that case `.root` is required because we get an error otherwise

src/components/ActionsMenu/Item/ActionsMenuItem.st.css
nktssh: fixed

src/components/ActionsMenu/README.md
nktssh: fixed

src/components/ActionsMenu/ActionsMenu.tsx
nktssh: moved to class

src/components/ActionsMenu/ActionsMenu.tsx
nktssh: removed

src/components/ActionsMenu/Item/ActionsMenuItem.st.css
jonathanadler: That's weird 🤔

src/components/ActionsMenu/Item/ActionsMenuItem.st.css
jonathanadler: This still needs to change

src/components/ActionsMenu/ActionsMenu.tsx
jonathanadler: What about this?

nktssh: resolved

src/components/ActionsMenu/Item/ActionsMenuItem.st.css
nktssh: fixed

nktssh:
@jonathanadler all the comments were fixed or commented

src/components/ActionsMenu/Item/ActionsMenuItem.st.css
```diff
@@ -0,0 +1,97 @@
...
+}
+
+.root .text {
+    -st-extends: Text;
+    -st-mixin: Text(
+        MainTextColor color(color-5)
+    );
+    display: flex;
+    flex-direction: column;
+    width: 100%;
```
jonathanadler: This should also have a `justify-content: center;` since we want the text to always be in the middle of the item

src/components/ActionsMenu/Item/ActionsMenuItem.st.css
```diff
@@ -0,0 +1,97 @@
...
+}
+
+.root {
+    -st-states: disabled, mobile;
+    height: 40px;
+    padding: 8px 24px;
+    display: flex;
+    cursor: pointer;
+    position: relative;
+}
```
jonathanadler: Take a look at this snapshot I took:
image
Note that you can see the main container's border.
And we need the hover background to take all the width (including the border).
So we need to have the item take `width: calc(100% + value(parentsBorder));`
And for that to work we also need to add `box-sizing: border-box` to the ActionsMenu .root

PR #178, opened 2019-08-22
Fix icon import from `wix-ui-icons-common`

bodia-uz:
Currently half of our bundle size is `wix-ui-icons-common`.
Its because of way if icons import in `Tabs` component.

This PR use direct icons import to prevent bundling all of them.

image


jonathanadler:
> @jonathanadler worth adding some lint rule to prevent this from happening in the future

So actually, we need to use our own icons instead of `wix-ui-icons-common`, and remove it from our package.json.
But i'll also see if I can add a lint rule.

PR #182, opened 2019-08-27
forwardRef

mbutsykin:
##### `withForwardRef` HOC
forward refs for
- Button
- TextArea


src/components/WithForwardRef/withForwardRef.tsx
```diff
@@ -0,0 +1,13 @@
+import * as React from 'React';
```
jonathanadler: There's a typo here. 
It's causing the build to fail

src/components/Button/Button.spec.tsx
```diff
@@ -30,15 +30,6 @@ describe('Button', () => {
     expect(await driver.isMobile()).toEqual(true);
   });
 
-  it('should expose focus method using ref', async () => {
```
jonathanadler: Why remove this test?
I think it's better to leave this, since this test was created to ensure this doesn't break.

src/components/WithForwardRef/withForwardRef.spec.tsx
```diff
@@ -0,0 +1,40 @@
+import * as React from 'react';
+import { createUniDriverFactory } from 'wix-ui-test-utils/uni-driver-factory';
+
+import { TextArea } from '../TextArea/TextArea';
+import { textAreaDriverFactory } from '../TextArea/TextArea.driver';
+
+import { Button } from '../Button/Button';
+import { buttonDriverFactory } from '../Button/Button.driver';
+
+describe('withForwardRef', () => {
```
jonathanadler: I think this test file should have a generic component used for the test.
So that we won't need to add a test for each component.

src/components/WithForwardRef/withForwardRef.tsx
```diff
@@ -0,0 +1,13 @@
...
+
+export interface WithForwardRefProps<T extends HTMLElement = HTMLElement> {
+  innerRef?: React.Ref<T>;
+}
+
+export function withForwardRef<T extends HTMLElement, P = {}>(
+  Component: React.ComponentType<P>,
```
jonathanadler: I don't like to use HOC's but I guess there's not a better way to do this.
The main two problems with HOC's are:

1. debugging - We should probably add a different display name (https://reactjs.org/docs/forwarding-refs.html#displaying-a-custom-name-in-devtools).
2. public/static methods and fields won't be externalized

So we need to see this works correctly (maybe add tests for this)

src/components/WithForwardRef/withForwardRef.tsx
```diff
@@ -0,0 +1,13 @@
+import * as React from 'React';
+
+export interface WithForwardRefProps<T extends HTMLElement = HTMLElement> {
+  innerRef?: React.Ref<T>;
```
jonathanadler: I'm also not so sure about this name.
I know changing this name will cause a greater change for the button component,
but maybe something like `exposedRef` will be more readable?
And maybe instead of `withForwardRef` we can have `exposeRef()`?
I'm really not sure about this.
what do you think?

src/components/WithForwardRef/withForwardRef.tsx
jonathanadler: @mbutsykin 
I'm not sure why, but it seems the build is breaking because the TextArea component's displayName isn't being copied correctly.
Can you try to see what's the problem here?
you can run `npm run autotools:sanity` to see the failing test

jonathanadler:
@mbutsykin just pinging you
Can take a look why the build is failing, and then we can merge this?
Thanks!

jonathanadler:
@mbutsykin 
Is this PR still needed?
Or can I close it?

PR #184, opened 2019-09-01
Visual tests abstraction

jonathanadler:


jonathanadler:
@nitayneeman @shlomitc 
I'm working on a POC for writing visual tests in a "common" language.
I'd love to hear what you guys think.
Maybe we can move this to the test-utils lib

shlomitc:
Hey. I like the suggested solutions. If I summarize what I see, is the following:

1. Semi framework for testing (visualize, story, snap)
2. Waiting for async interactions (VisualTestContainer)
3. Common configs in applitiools.config.js (waitBeforeScreenshots)

They are all serious issues to be solved and can be a "one-solution-fits-all" tool or separate ones. At first, I thought to extract only the VisualTestContainer, but now it feels to me like this actually is a framework and would work nicely.

I like the idea to make it like a `describe/it` approach. Maybe naming can be enhanced but overall looks like a good direction.

I think it should go out to a separate library so after merging this PR, refactor it all to a new folder and let's design the API.

PR #185, opened 2019-09-01
#SCHED-11924 | Bookings Accessibility Fixes (full audit) - tabs fixes

noakorkus:


jonathanadler:
@noakorkus 
So from what I understand, the test is failing because the role `tablist` is "legal" only if used on `ol` or `ul` elements.
So, I think what we should do is:
1. check with Neil if it's ok to leave it as a `nav` (MDN [says](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role#Example) it's ok on a `div` so I'm assuming it's also ok on a `nav`).
2. If it's ok, you can add ```tabsMetadata.nonA11yCompliant = true;```, to the `Tabs.meta.tsx` file, which will ignore the failing test.
3. If it's not ok, then we'll need to check with Neil, how this should be implemented

PR #187, opened 2019-09-03
feat(a11y): change tabs to implement navigation a11y guidelines

carmelc:


src/components/Tabs/ScrollableTabs/ScrollableTabs.st.css
```diff
@@ -75,3 +75,9 @@
 .root:variant(fullWidth) .tab {
     flex: 1;
 }
+
+.itemsList {
+    margin: 0;
+    padding: 0;
+    height: 100%;
+}
```
jonathanadler: I think that in order to fix the `fullWidth` issue, the `.itemList` needs to get the `alignments` and `variant` states that the `.nav` received before the `ul` was introduced.
Meaning we need something like: 
```css
.root:alignment(left) . itemsList {
    text-align: left;
}
.root:alignment(right) . itemsList {
    text-align: right;
}
.root:variant(fullWidth) . itemsList {
    display: flex;
}
```

instead of what we have above 👆

carmelc: Done

jonathanadler:
> god bless eyes! - fix mobile too

Totally 😄

PR #193, opened 2019-09-09
use storybook-snapper

jonathanadler:
add storybook-snapper + use it instead of local Snapper


applitools.config.js
```diff
@@ -1,7 +1,7 @@
 const { execSync } = require('child_process');
 const appName = require('./package.json').name;
 const merge = require('lodash/merge');
-const { DATA_READY_HOOK } = require('./test/visual/dataHooks.js');
+const { DATA_READY_HOOK } = require('storybook-snapper/dist/src/index');
```
shlomitc: I'd ditch this one as discussed privately

tsconfig.json
```diff
@@ -10,6 +10,7 @@
   },
   "include": [
     "./node_modules/wix-ui-core/dist/src/globals.d.ts",
+    "./node_modules/storybook-snapper/src/external-types.d.ts",
```
shlomitc: why is this required?

jonathanadler: Good question.
I was having some problems with importing the library.
It might not be needed. I'll check.

applitools.config.js
jonathanadler: So you're saying `storybook-snapper` shouldn't have a default selector for this?

My only problem is that we need this selector both in the config file, and in the storybook  react component.
I think their design is kind of limiting in this way.

And since I really like having a single source of truth, I'll still need to import the hook from a datahook file.

PR #194, opened 2019-09-10
Optional text link in AvatarGroup

dovilel:


dovilel:
> Looks pretty good!
> 
> Just a few comments:
> I think it would be better to use `TextButton` everywhere instead of `TextLink` just for the sake of consistency (even though the zeroheight spec wrote text link).
> 
> In addition, I'm not 100% sure about the `TextButton` onClick callback.
> Users might want to render it as an anchor and pass an href.
> So i'm thinking if it might be better to have a `textButtonProps` instead (even though I usually don't like to have props like these)?
> wdyt?

Agreed. How about compound component then? I don't really like passing props as object either.
AvatarGroup.TextButton or something, that's prestyled, but otherwise is a normal TextButton and takes all TextButton props.

jonathanadler:
> > Looks pretty good!
> > Just a few comments:
> > I think it would be better to use `TextButton` everywhere instead of `TextLink` just for the sake of consistency (even though the zeroheight spec wrote text link).
> > In addition, I'm not 100% sure about the `TextButton` onClick callback.
> > Users might want to render it as an anchor and pass an href.
> > So i'm thinking if it might be better to have a `textButtonProps` instead (even though I usually don't like to have props like these)?
> > wdyt?
> 
> Agreed. How about compound component then? I don't really like passing props as object either.
> AvatarGroup.TextButton or something, that's prestyled, but otherwise is a normal TextButton and takes all TextButton props.

Yeah, that's sounds like a great idea!
So it can maybe even extend the `TextButton` component?

PR #202, opened 2019-09-18
Fix AvatarGroup text button overrides

dovilel:


dovilel:
@jonathanadler Hey. Turns out I really did not do overrides correctly and it worked for me in examples by accident. Fixing, so I can actually override it from where I need to :)

src/components/AvatarGroup/AvatarGroup.st.css
```diff
@@ -156,19 +156,16 @@
...
-       MainTextColor '"--TextButtonColor"',
-       MainTextFont '"--TextButtonFont"'
+       MainTextColor value(TextButtonColor),
+       MainTextFont value(TextButtonFont)
     );
 }
 
-:size(xSmall) .textButton,
-:size(xxSmall) .textButton {
+.smallTextButton {
```
jonathanadler: Why add a class and not use the `state`?

dovilel: turns out starting with just :size doesn't work on processed css. I added .root now before it, and it's ok. So changed back to original classes, looks clean now.

dovilel:
@jonathanadler please check again! Thanks

PR #203, opened 2019-09-18
Add BaseProps to all components

nktssh:


src/components/TextField/TextField.tsx
```diff
@@ -60,12 +62,16 @@ export const TextField: React.FunctionComponent<TextFieldProps> = (
...
-      })}
+      {...style(
+        'root',
+        {
+          error,
+          theme,
+          success,
+          disabled,
+        },
+        { className },
```
jonathanadler: instead of passing just the className, lets pass `...rest`

src/types.ts
```diff
@@ -0,0 +1,3 @@
+export interface TPAComponentProps {
+  className?: string;
+}
```
jonathanadler: Let's also add
```
'data-hook'?: string;
```

src/components/Input/Input.tsx
```diff
@@ -5,8 +5,9 @@ import {
 } from 'wix-ui-core/input';
 import style from './Input.st.css';
 import { ErrorMessageWrapper, ErrorProps } from '../ErrorMessageWrapper';
+import { TPAComponentProps } from '../../types';
 
-export interface TPAInputProps {
+export interface TPAInputProps extends TPAComponentProps {
```
jonathanadler: This component is theoretically deprecated, so let's not change it.

jonathanadler: I still prefer we don't change this

nktssh:
Rolled `Input` back

PR #210, opened 2019-09-29
Add rtl to Avatar Group

dovilel:


jonathanadler:
Hey @dovilel 
Can you explain why this is needed.
As far as I can see, the AvatarGroup component supports rtl out of the box, with no need for adding a state.
It does, however, rely on having a `direction="rtl"` on one of the ancestors.  
You can see it working when toggling the `imitate RTL` toggle in the playground here:
https://wix-wix-ui-tpa.surge.sh/?activeTab=Playground&path=%2Fstory%2Fcomponents--avatargroup

dovilel:
> Hey @dovilel
> Can you explain why this is needed.
> As far as I can see, the AvatarGroup component supports rtl out of the box, with no need for adding a state.
> It does, however, rely on having a `direction="rtl"` on one of the ancestors.
> You can see it working when toggling the `imitate RTL` toggle in the playground here:
> https://wix-wix-ui-tpa.surge.sh/?activeTab=Playground&path=%2Fstory%2Fcomponents--avatargroup

Hey,
it doesn't work correctly with the extra text link. dir="rtl" does not handle right/left margins and paddings (shouldn't it?).
I see that TextArea had the same thing and also added rtl class specifically to changing padding.

Screen Shot 2019-10-02 at 11 25 07 AM

jonathanadler:
Oh I see, you're right.
My bad for not noticing this back when we merged this.
Maybe instead of introducing the `rtl`, we can add a padding by using 
```css
.avatarContainer::after {
  content: '\00a0';
  width: thePadding;
}
```
Do you think it would work?

dovilel:
> Oh I see, you're right.
> My bad for not noticing this back when we merged this.
> Maybe instead of introducing the `rtl`, we can add a padding by using
> 
> ```css
> .avatarContainer::after {
>   content: '\00a0';
>   width: thePadding;
> }
> ```
> 
> Do you think it would work?

For paddings - yeah. Do you know a trick for negative margins?

src/components/AvatarGroup/AvatarGroup.st.css
```diff
@@ -142,18 +172,21 @@
...
+:size(xSmall) .textButtonContainer::before {
+    content: '\00a0';
+    width: 10px;
 }
 
-:size(xxSmall) .textButtonContainer {
-    margin-left: 8px;
+:size(xxSmall) .textButtonContainer::before {
+    content: '\00a0';
+    width: 8px;
```
jonathanadler: I think only this part is required.
The rest of the changes (above) aren't needed.
Since the  padding is canceled by the margins.

dovilel: No, the other changes are also needed, or the label does not position itself correctly (since margin is on the wrong side of the last avatar).
So the avatars happen to look correct in rtl, but their bounding box is a bit off.

jonathanadler: I think you just need to take into account the negative margins.
So for instance something like this:
```css
:vars {
  largeMargin: 14px;
}

.root:size(large) .avatarContainer {
    margin-left: calc(-1 * value(largeMargin));
}

.root:size(large) .textButtonContainer::before {
  content: '\00a0';
  width: calc(12px + value(largeMargin));
}
```
I tried this locally, I think this works.

dovilel: Not really, since margin is only there on one side, and it needs to be compensated for only in either ltr or rtl. So then the padding ends up different sizes.
Basically, without some sort of specific rtl handling we are not allowed to write "x-left" on anything

I think it's getting a bit hacky at this point - I feel like we are trying to run away from proper rtl handling. Either generated css, where all the left/right stuff is switched for [dir=rtl], or just being able to use [dir=rtl] in the css? Can we in stylable? It seems to work, but Stylable gets mad at me.

Screen Shot 2019-10-02 at 4 05 48 PM

jonathanadler: But 'ltr' is working for the avatargroup (without the text), so i'm not so sure why the change to `.avatarContainer` css rules is necessary?
Sorry for nagging, I just really prefer not adding the `rtl` flag.
Regarding `[dir=rtl]`, it's better to use the `rtl` state than using this selector, if we decide to use the `rtl` state.
I'm just not sure yet that it's needed.

dovilel: Maybe I'll add extra container to contain the Avatars then, because right now the padding, that compensates on margins, is on whole component, and anything besides Avatars hides under them in rtl:

Screen Shot 2019-10-02 at 4 48 23 PM

So it's working in rtl, but it's not symmetrical css wise :) Which makes adding anything beside it wonky.

jonathanadler: 🤔 ok I see now.
Bummer 😞 
👍 
If it won't work we might need to add the `rtl` state.

dovilel:
@jonathanadler how about this solution?

src/components/AvatarGroup/AvatarGroup.visual.tsx
```diff
@@ -3,6 +3,7 @@ import { storiesOf } from '@storybook/react';
 import { AvatarGroup } from './';
 import { AvatarGroupSize } from './AvatarGroup';
 import { VisualTestContainer } from '../../../test/visual/VisualTestContainer';
+import { TPAComponentsProvider } from '../TPAComponentsConfig';
```
jonathanadler: Oh, I guess this isn't necessary though, right?

dovilel: whops! removed

PR #211, opened 2019-10-03
Storybook style api

jonathanadler:
Added a automatically generate Style API table


jonathanadler:
In this PR I used the style api tab on the `Tabs` component, as a POC.
Will add to the generator if it seems to be a good solution.

jonathanadler:
See: https://wix-wix-ui-tpa-pr-211.surge.sh/?activeTab=Style%20API&path=%2Fstory%2Fcomponents--tabs

src/components/Tabs/docs/index.story.tsx
```diff
@@ -35,5 +58,65 @@ export default {
...
+        { title: 'Style API', sections: [styleApi({ css })] },
+        { title: 'TestKit', sections: [testkit()] },
+        { title: 'Playground', sections: [playground()] },
+        {
+          title: 'Settings Panel',
+          sections: [
+            settingsPanel({
+              title: 'Tabs Panel',
+              example: <TabsExtendedExample items={items} />,
+              rawSource: ExtendedRawSource,
```
shlomitc: I would aim to have this created automatically

src/components/Tabs/docs/index.story.tsx
```diff
@@ -35,5 +58,65 @@ export default {
...
+              rawCSSSource: ExtendedCSSRawSource,
```
shlomitc: I would aim to have this created automatically

stories/utils/StyleApi/StyleApi.tsx
```diff
@@ -0,0 +1,61 @@
+import * as React from 'react';
```
shlomitc: This file is handling two concerns that needs to be split.

1. scrape the data
1. render the settings panel

the first one, is something that should be done during "built time", or more accurate during the webpack bundling of the storybook.

The second one is already separated so it's good. this file should just aggregate the parsed data and render the table and as a next step - an automated panel

src/components/Tabs/docs/index.story.tsx
```diff
@@ -35,5 +58,65 @@ export default {
...
+        { title: 'API', sections: [api()] },
+        { title: 'Style API', sections: [styleApi({ css })] },
+        { title: 'TestKit', sections: [testkit()] },
+        { title: 'Playground', sections: [playground()] },
+        {
+          title: 'Settings Panel',
+          sections: [
+            settingsPanel({
+              title: 'Tabs Panel',
+              example: <TabsExtendedExample items={items} />,
```
shlomitc: I would aim to have this created automatically, look very much like `playground`.
so that the same meta data that you scrape from the `StyleApi` will be available for this panel as well.
it will be an auto-generated opinionated panel

PR #214, opened 2019-10-04
Feat/new card

bordi:


src/components/NewCard/NewCard.tsx
```diff
@@ -0,0 +1,68 @@
...
+interface DefaultProps {
+  stacked: false;
+  'data-hook': string;
+}
+
+const Container: React.FunctionComponent<ContainerProps> = props => {
+  const { className, children, minWidth } = props;
+  return (
+    <div
+      style={{ flexBasis: `${minWidth}px` }}
```
jonathanadler: 🤔 I think we might need to add the minWidth conditionally, only if it exists.
wdyt?

src/components/NewCard/NewCard.visual.tsx
```diff
@@ -0,0 +1,54 @@
...
+  render() {
+    return (
+      <div>
+        <NewCardWiringExample />
+      </div>
+    );
+  }
+}
+
+visualize('New Card', () => {
```
jonathanadler: 💪

src/components/NewCard/README.md
```diff
@@ -0,0 +1,4 @@
+## NewCard
+An implementation of an NewCard for TPAs
+This Readme was auto-generated and needs to be filled manually
+
```
jonathanadler: We should add a short description on how to use the card.
no need to add the stylable variables table and all that.

src/components/NewCard/NewCard.visual.tsx
jonathanadler: btw, since this test doesn't need to be async, we can remove the `onDone` prop from the `NewCardVisual` component.
You can see the README here: https://www.github.com/wix-incubator/storybook-snapper

jonathanadler:
For some reason the stacked example isn't working correctly
image

src/components/NewCard/NewCard.tsx
bordi: It won't be added if `minWidth = undefined`. 
I can add a condition so that logic will be explicit.

src/components/NewCard/NewCard.tsx
```diff
@@ -0,0 +1,63 @@
...
+    return {
+      [NEWCARD_DATA_KEYS.Stacked]: stacked,
+    };
+  }
+
+  render() {
+    const { stacked, children, className, ...rest } = this.props;
+
+    return (
+      <section
```
jonathanadler: 👍 for using `section`

src/components/NewCard/docs/NewCardWiringExample.tsx
```diff
@@ -0,0 +1,124 @@
...
+        <Text typography={TYPOGRAPHY.smallTitle}>I'm overlapped</Text>
+        <br />
+        <Text typography={TYPOGRAPHY.runningText}>
+          I have minWidth={300} and media block have minWidth={300}
+          and I'm Overlapped!
+        </Text>
+      </>
+    </NewCard.Container>
+  </NewCard>
+);
```
jonathanadler: So the designers asked me to remove the overlapping example, since they don't want to encourage people to use this type of layout.

src/components/NewCard/docs/NewCardWiringExample.st.css
```diff
@@ -0,0 +1,54 @@
...
+.card.overlapped::container.info {
+    border: 1px solid #000;
+}
+
+.card.overlapped::container.info {
+    border: 1px solid #000;
+    background: #fff;
+
+    margin: value(paddingValue) 0  value(paddingValue) calc(-1 * value(paddingValue));
+}
```
jonathanadler: Same here for the overlapping

src/components/NewCard/README.md
```diff
@@ -0,0 +1,41 @@
+## NewCard
+An implementation of an NewCard for TPAs
+
```
jonathanadler: We should probably add that this is the formal `Card` that should be used, and that the former `Card` component will be deprecated.

PR #215, opened 2019-10-06
add eyes server url env variable

jonathanadler:


shlomitc:
I guess this is temporary, right? It should be working out of the box and not configurable per project

jonathanadler:
> I guess this is temporary, right? It should be working out of the box and not configurable per project

Yes, I wasn't expecting you to go over this 😄
It's just to see that all flows work as expected.
Currently it seems that the `eyes.it` aren't using the new server, or at least the checks in the pr (test/applitools).

PR #231, opened 2019-11-05
<Modal/>

nataliia-khitsan:


src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+interface State {}
+
+/** The modal component provides a solid foundation for creating dialogs, popovers, lightboxes, or whatever else. */
+export class Modal extends React.Component<ModalProps, State> {
+  static displayName = 'Modal';
+  static defaultProps: DefaultProps = {
+    inFullScreen: false,
+    withCloseButton: true,
+    withBackground: true,
```
jonathanadler: Even though it makes sense to have these two props true by default, I think it's weird to have it a `with...` prop be true without actually passing the prop.
So i'd either change the name to: `hideCloseButton`, or remove this default value.
(For both `withCloseButton` and `withBackground`)

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,113 @@
...
+  };
+}
+
+interface State {}
+
+/** The modal component provides a solid foundation for creating dialogs, popovers, lightboxes, or whatever else. */
+export class Modal extends React.Component<ModalProps, State> {
+  static displayName = 'Modal';
+  static defaultProps: DefaultProps = {
+    inFullScreen: false,
```
jonathanadler: I think this can be just `fullScreen`

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,113 @@
...
+      return null;
+    }
+
+    return (
+      <TPAComponentsConsumer>
+        {({ mobile }) => (
+          <div
+            {...styles('root', { mobile, withBackground, inFullScreen }, rest)}
+            data-mobile={mobile}
+            onClick={onRequestClose}
```
jonathanadler: We need to add a prop if `clickOutside` should close the modal.
In addition, it should be clickable only if the background exists.
By the way, the designers are telling me to have it clickable on mobile by default.

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,113 @@
...
+          >
+            <div
+              className={styles.contentWrapper}
+              onClick={this.handleContentClick}
+              role="dialog"
+              aria-modal="true"
```
jonathanadler: I see you haven't used the `aria` prop.

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,113 @@
...
+/** The modal component provides a solid foundation for creating dialogs, popovers, lightboxes, or whatever else. */
+export class Modal extends React.Component<ModalProps, State> {
+  static displayName = 'Modal';
+  static defaultProps: DefaultProps = {
+    inFullScreen: false,
+    withCloseButton: true,
+    withBackground: true,
+    aria: {
+      labeledBy: 'dialog-label',
+    },
```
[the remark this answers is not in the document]
jonathanadler: In addition, we shouldn't have a default value for this, since this is something we really want our consumers to use.
And since this also needs to be translated to fit the users language, we should let consumers configure this.

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,113 @@
...
+            <div
+              className={styles.contentWrapper}
+              onClick={this.handleContentClick}
+              role="dialog"
+              aria-modal="true"
+            >
+              {withCloseButton ? (
+                <div className={styles.closeButtonWrapper}>
+                  <IconButton
+                    aria-label="Close the dialog window"
```
jonathanadler: This is problematic.
We can't decide what the `aria-label` should be, as I wrote before (in the `defaultProps` section).
This is making me think we might need to change the implementation of the component.

src/components/Modal/Modal.st.css
```diff
@@ -0,0 +1,117 @@
...
+    left: 0;
+}
+
+.root:mobile .closeButton {
+    padding: 20px;
+}
+
+.root:mobile .closeButton svg {
+    width: 12px;
+    height: 12px;
```
jonathanadler: I'm not sure this is correct.
It seems the icon is very small on mobile.
image

src/components/Modal/Modal.visual.tsx
```diff
@@ -0,0 +1,83 @@
...
+    {
+      it: 'Without background',
+      props: {
+        isOpen: true,
+        onRequestClose: () => {},
+        withBackground: false,
+        mobile: isMobile,
+      },
+    },
+  ];
```
jonathanadler: we should also have an `RTL` visual test

src/components/Modal/docs/index.story.tsx
```diff
@@ -0,0 +1,100 @@
...
+    'data-hook': 'storybook-Modal',
+  }),
+  exampleProps: {
+    isOpen: false,
+    inFullScreen: false,
+    withCloseButton: false,
+    withBackground: false,
+    onRequestClose: () => {
+      console.log('on request close');
+    },
```
jonathanadler: The problem with the `exampleProps` is that once the Modal is opened in the Playground, there's no way to close it.
Maybe we can use `componentWrapper` in order to add a state to the example.
You can see the documentation here:
https://github.com/wix/wix-ui/blob/aca805fe29ba11be17c85f492af310681dda7b86/packages/wix-storybook-utils/docs/usage.md (under `componentWrapper`)

src/components/Modal/index.tsx
```diff
@@ -0,0 +1,2 @@
+export { Modal, ModalProps } from './Modal';
+export { ExampleModalApp } from './docs/ExampleModalApp';
```
jonathanadler: Is there a reason to export the example?

jonathanadler:
Oh, and I also forgot that we need to have an option for the modal to pop-up from the bottom on mobile.
This is what we intentionally wanted to use, for the dropdown.
I think however, it might be better to add a `Dialog` component for instance, which will show as a modal, but also might be opened like a bottom drawer.
The modal component will use this internally, but expose it without the bottom drawer, and the dropdown will use it only with the drawer

jonathanadler:
@zhukovka @nataliia-khitsan 
After going over the zeroheight again, and talking to the designers, it seems we should have a slightly different API than what I thought.
I think we should have an internal `Modal` component, like you developed here.
Just a few changes to the API:
* a `clickOutside` prop (maybe a bad name, we should look at how it's called in other libraries)
* `aria-` props
* Instead of heaving the `CloseButton` component, we can leave it as a prop, but we'll also need to add a `closeButtonAriaLabel` prop (unfortunately). Other ideas are welcome.
*  We need to add a11y props to align with a11y modal specifications: https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/alertdialog.html

Then we can extend this component with a `Dialog` component, and a `Drawer` component (which we can then use for the mobile dropdown).

Please talk to me if there's anything that's not understood.
If you have any limitations, regarding time, etc. please talk to me as well.

zhukovka:
Thank you, Jonathan
we will work on this next week because we now are overloaded with our
current tasks

PR #234, opened 2019-11-13
add SuccessIcon to textField component

ariel7342:
@jonathanadler add SuccessIcon to textField component

### DIFF
--- a/src/assets/icons/CheckSuccess.svg
+++ b/src/assets/icons/CheckSuccess.svg
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="24px" height="24px" viewBox="0 0 24 24" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 56 (81588) - https://sketch.com -->
+    <title>09 Icons / Check / Regular Copy</title>
+    <desc>Created with Sketch.</desc>
+    <g id="09-Icons-/-Check-/-Regular-Copy" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+        <g id="Check" transform="translate(5.000000, 7.000000)" fill="#57C487" fill-rule="nonzero">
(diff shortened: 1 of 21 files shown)


src/assets/icons/CheckSuccess.svg
```diff
@@ -0,0 +1,11 @@
...
+<svg width="24px" height="24px" viewBox="0 0 24 24" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 56 (81588) - https://sketch.com -->
+    <title>09 Icons / Check / Regular Copy</title>
+    <desc>Created with Sketch.</desc>
+    <g id="09-Icons-/-Check-/-Regular-Copy" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+        <g id="Check" transform="translate(5.000000, 7.000000)" fill="#57C487" fill-rule="nonzero">
+            <polygon id="Path-2" points="13.3 0 14 0.7 5.3 10 0 5.8 0.6 5 5.2 8.7"></polygon>
+        </g>
+    </g>
+</svg>
```
jonathanadler: The svg just needs to be cleaned a bit.
Remove id's and unnecessary components (title).
It's just that it causes problems sometimes.

src/components/TextField/TextField.driver.ts
```diff
@@ -62,3 +69,18 @@ export const textFieldDriverFactory = ({ element, eventTrigger }) => {
...
   };
 };
+
+export interface TextFieldDriver extends BaseUniDriver {
+  isSuccessIconExist(): Promise<boolean>;
+  isSuccess(): Promise<boolean>;
+}
+
+export const textFieldUniDriverFactory = (base: UniDriver): TextFieldDriver => {
+  const stylableUtil = new StylableUnidriverUtil(style);
```
jonathanadler: We actually prefer to not use StylableUnidriverUtil, it causes some problems in e2e tests.
Also, of course we prefer using Unidrivers, but in this case, since we're using a driver from wix-ui-core, and also we have automatic utilities exporting these drivers.

So it's best to move these methods to the driver above, instead of creating a new driver.

ariel7342:
@jonathanadler updated after cr

PR #240, opened 2019-11-19
Auto settings api 2 no manifest

vytenisu:


jonathanadler:
Looks very good!
Let's wait for the changes in `wix-storybook-utils` and continue then.

scripts/stylable-metadata-plugin.js
```diff
@@ -0,0 +1,175 @@
+const path = require('path')
+const fs = require('fs')
+const pathFinder = require('react-autodocs-utils/src/path-finder');
+const parse = require('react-autodocs-utils/src/parser/parse');
```
argshook: i never really intended these files to be publicly available, not sure it's a good idea to use them directly (talking about `parse` and `visit`).

since you're using `@babel/types` you can also do `@babel/traverse` (for `visit`) & `@babel/parse` (for, well, `parse`)

scripts/stylable-metadata-plugin.js
```diff
@@ -0,0 +1,175 @@
...
+        }
+      }
+    }
+  })
+
+  resolve(null)
+})
+
+const extractStylableVarStructure = ast =>
+  [].concat.apply(
```
argshook: perhaps you want a `reduce` here? `.apply` is a rare thing these days, but maybe it's here for a reason?

scripts/stylable-metadata-plugin.js
```diff
@@ -0,0 +1,175 @@
...
+const getOverridableVars = ast => {
+  const allVars = getAllStylableVars(ast)
+  const vars = []
+
+  Object.entries(allVars).forEach(([name, info]) => {
+    if (info.value.startsWith('-')) {
+      const defaultValue = getDefaultValue(name, allVars)
+      const commentInfo = parseComment(info.comment)
+
+      vars.push({
```
argshook: `reduce`! :)

stories/utils/SettingsPanel.tsx
```diff
@@ -19,6 +22,81 @@ interface ISettingsPanelConfig {
...
+          const defaultColor = colorMatch ? colorMatch[0] : 'color-1';
+
+          return {
+            label: variableInfo.name,
+            wixParam: `${componentName}-${variableInfo.name}`,
+            defaultColor,
+          };
+        });
+
+      const styleExample = require(`!raw-loader!../../src/connected-components/${componentName}/${componentName}.example.st.css`);
```
argshook: oof this can be heavy, if i recall correctly having dynamic path in `require` moves the whole file system into the bundle (might be even with node_modules). 

this would increase storybook load time quite a lot. Please double check

on a side note, is this really the way to go? is it guaranteed that all `connected-components` have `*.example.st.css`

scripts/stylable-metadata-plugin.js
vytenisu: What I would really like to use is not reduce but flat(). However it appeared in node 11 and since it is relatively new, just in case decided not to use it yet.

stories/utils/SettingsPanel.tsx
vytenisu: That is not exactly correct as far as I know. Dynamic require makes WebPack perform static analysis on code and include whole folder (not whole file system) into the bundle. Currently this folder contains 4 files which is not a lot. Even it it had more - all of those files would be needed in bundle.

Having said that, idea is to get rid of these files (if I am lucky enough - today, if not - maybe will need to leave it for someone else). Initial idea was to generate wrappers which connect Stylable to settings. I have a working generation script in a separate branch (not meant to be merged anymore).

Afterwards we spoke with Jonathan and decided to leave these files sort-of hard-coded for now and aim to generate wrappers in memory instead of generating files. So if that succeeds, these files will be just removed.

src/connected-components/TextArea/TextArea.example.st.css
```diff
@@ -0,0 +1,19 @@
...
+      BackgroundColor --TextArea-BackgroundColor,
+      TextColor --TextArea-TextColor,
+      DisabledTextColor --TextArea-DisabledTextColor,
+      PlaceholderColor --TextArea-PlaceholderColor,
+      TextFont --TextArea-TextFont,
+      ErrorBorderColor --TextArea-ErrorBorderColor,
+      SuccessBorderColor --TextArea-SuccessBorderColor,
+      DisabledBorderColor --TextArea-DisabledBorderColor
+  );
+}
```
jonathanadler: I'd rather have this file inside the `TextArea` component folder, instead of adding the `connected-components` directory.
In addition, I think we already generate a `${componentName}ExtendedExample` file in the component's `docs` folder.
Maybe we can use this file instead of creating a new one?

src/connected-components/TextArea/TextArea.st.css
```diff
@@ -0,0 +1,19 @@
...
+      BackgroundColor --TextArea-BackgroundColor,
+      TextColor --TextArea-TextColor,
+      DisabledTextColor --TextArea-DisabledTextColor,
+      PlaceholderColor --TextArea-PlaceholderColor,
+      TextFont --TextArea-TextFont,
+      ErrorBorderColor --TextArea-ErrorBorderColor,
+      SuccessBorderColor --TextArea-SuccessBorderColor,
+      DisabledBorderColor --TextArea-DisabledBorderColor
+  );
+}
```
jonathanadler: Same as previous comment 👆

src/connected-components/TextArea/index.tsx
```diff
@@ -0,0 +1,7 @@
+import * as React from 'react';
+import styles from './TextArea.st.css';
+import { TextArea as TextAreaOrig } from '../../components/TextArea';
+
+export const TextArea = props => (
+  <TextAreaOrig {...props} {...styles('root', {}, props)} />
+);
```
jonathanadler: Why do we need both `connected-components/TextArea/index.tsx` and `connected-components/TextArea/index.example.tsx` files?
They seem to be the same.
And also, would rather have these files as part of the generated files in the `TextArea` folder, as previously commented

scripts/stylable-metadata-plugin.js
```diff
@@ -0,0 +1,190 @@
...
+
+  return {
+    metadata: {
+      ...metadata,
+      stylable: {
+        overridableVars: getOverridableVars(stylableAst)
+      }
+    }
+  }
+}
```
jonathanadler: Let's move this file to the `.storybook` folder, for now.
We usually use the `scripts` folder for scripts that we run on CI, or build, etc.
And since this file is only used in the `.storybook` webpack configuration, I think we can place it there.

scripts/stylable-metadata-plugin.js
argshook: wix is on node 12, also in CI. It should be just fine to use `Array.prototype.flat`

stories/utils/SettingsPanel.tsx
argshook: hmm i'm not sure about it. i remember having lots of tweaks with things like `require.resolve` and weird regex'es.

Also, `storyName` is not necessarily same as component name.
I didn't delve into whole code in this PR so can't say if this approach is good or not. It just feels very fragile

vytenisu: This is sort-of temporary solution with intention to replace it with dynamic wrapper generation. I am looking into this right now but not sure if I will be able to complete this part due to Stylable generation challenges.

Worst case, will try to tidy this place up a bit and add a fat todo :)

argshook: > Worst case, will try to tidy this place up a bit and add a fat todo :)

for sure, don't sweat it, we'll look into it later on.

But we'll look into it only if there's a fat todo :D

vytenisu:
WIll start working after https://github.com/wix/wix-ui/pull/1629 is merged

PR #244, opened 2019-12-02
add focus functionality to TextField component

ariel7342:
@jonathanadler  here is the focus part, let me know if there is something to fix


ariel7342:
@jonathanadler pushed a fix for focus ring

jonathanadler:
@ariel7342 
I see that the tests are failing.
I'm not sure the padding works as expected (maybe it's only the line theme?)

src/components/TextField/TextField.st.css
```diff
@@ -165,13 +170,19 @@
...
+
 .suffixWrapper {
     height: 24px;
     display: flex;
     margin: 0 -4px;
     justify-content: flex-end;
     white-space: nowrap;
     vertical-align: middle;
+    padding-left: 12px;
+    padding-right: 12px;
```
jonathanadler: It seems that this is working for the SuccessIcon suffix, but not the ErrorIcon

ariel7342: are you sure? because its the same component with same class just different Icon inside

PR #248, opened 2019-12-10
Text field blur

ariel7342:
add blur function for text filed
@jonathanadler


jonathanadler:
@ariel7342 
I see that you have conflicts.
You need to first pull-rebase.

Also, it would help if you could remove the unneeded changes in the `package.json` file

Thanks

ariel7342:
@jonathanadler suppose to be fine now

PR #250, opened 2019-12-17
remove typescript files from assets folder

jonathanadler:
move tsx files from assets folder + change index.ts to js


jonathanadler:
@shlomitc 
I wanted a second pair of eyes to see that this change won't affect consumers.
Can you take a look or reference someone to take a look?

src/assets/icons/index.js
```diff
@@ -0,0 +1,23 @@
+const CaretDown = require('./CaretDown.svg').ReactComponent;
```
shlomitc: Why is this file needed?
I assume is that it is for exposing icons to consumers.
Why can't we expose it just as regular components?

jonathanadler: We used the index file to expose the components,
But I wrote it in typescript.
It appears that typescript files aren't compiled if they're in the `assets` folder.
So now, in order to support consumers using this file, I had to change it to a regular commonJS file.

shlomitc: As you say, `assets` folder isn't transpiled or modified in the build step. Why do the icons need to be in this folder? why not exposing them like other components from `src`?

jonathanadler: You're right, let start over cause I didn't explain what's the purpose of this PR 🤦‍♂ 

We added this index file, and the svg's to the assets a while back.

Now, since the typescript files aren't compiled in the assets folder, it's causing consumers' projects to fail (in certain situations).
I can't remove the files from the assets folder, since this will be a breaking change.
So I'm just trying to fix this issue specifically without making a breaking change.

In the future we'll be able to introduce a breaking change that fixes this icons issue.

I was thinking that the best option would be adding our icons to the wix-ui-icons-common lib, so that we won't need to deal with the whole build process, and ssr, etc.

shlomitc: What if you keep the svgs but not create the index file?
It's not a good idea to expose such index because then assets are not tree-shaked.

jonathanadler: Then it's a breaking change.
I'm looking for a non-breaking change for now, so that it won't hold back developers.

In the future I'd like to move the icons to wix-ui-icons-common

shlomitc: If it's not working for someone already - let's not make it work but rather ask him to import correctly the SVGs.

I prefer not merging the `on-stage` icons with `wix-ui-icons-common`, they are carefully picked by the design system UX team. if we wish to merge them, it should be decided with Ben

jonathanadler: So the problem isn't with the index file at all,.
Their problem is that they use a component that uses an icon that's written in a tsx file, which was in the assets folder (which was a mistake).
So as part of this fix (I moved the tsx icon file to the components folder - which is also kind of a breaking change), I thought I should also change the `index.tsx` file to an `index.js` file.

And regarding the `wix-ui-icons-common` of course I don't mean to mix between the different icons.
But having a separate directory for these icons could be very helpful for us.
I can also create our own `wix-ui-platform-icons-common` package, but it's a bit of a duplication.

shlomitc: I probably still don't understand what's broken now - can you share the case when it breaks?

Regarding the icons, let's discuss it with Alex and Ben

PR #259, opened 2020-01-02
[feat] SectionNotification component

kristina1lyba:
Addresses https://jira.wixpress.com/browse/PAY-2112


src/components/SectionNotification/SectionNotification.driver.ts
```diff
@@ -0,0 +1,43 @@
...
+  hasControls(): Promise<boolean>;
+  getText(): Promise<string>;
+  getIconContainer(): UniDriver;
+  getControlsContainer(): UniDriver;
+}
+
+export const sectionNotificationDriverFactory = (
+  base: UniDriver,
+): SectionNotificationDriver => {
+  const stylableUtil = new StylableUnidriverUtil(style);
```
jonathanadler: Very nice driver
However, we prefer to to not use `StylableUnidriverUtil` since it may cause problems with certain tests.
I believe we removed this from our generator, did you use `npm run generate` to create this new component?

In any case, instead of using `StylableUnidriverUtil`, you can use `data-` attributes in order to test the styling state of the component.
For instance you can see what we have in https://github.com/wix/wix-ui-tpa/blob/master/src/components/Tabs/Tabs.driver.ts

src/components/SectionNotification/SectionNotification.e2e.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  it('should render', async () => {
+    const driver = sectionNotificationTestkitFactory({ dataHook });
+    await waitForVisibilityOf(
+      await driver.element(),
+      'Cannot find SectionNotification',
+    );
+    expect((await driver.element()).isDisplayed()).toBe(true);
+  });
+});
```
jonathanadler: This test isn't needed, so we can just remove this file.

src/components/SectionNotification/SectionNotification.st.css
```diff
@@ -0,0 +1,115 @@
...
+/* Mobile styles */
+.root:mobile::controls {
+    margin-top: value(ContentPadding);
+    margin-left: 0;
+    margin-right: 0;
+}
+
+.root:mobile::main {
+    flex-direction: column;
+}
```
jonathanadler: Can this be implemented without the `rtl` state maybe?
Maybe just using flex for instance?
It would be better if so.

In any case, instead of writing:
```css
.root:<state-prop>::className {}
```

You should write:
```css
.root:<state-prop> .className {}
```

src/components/SectionNotification/SectionNotification.tsx
```diff
@@ -0,0 +1,67 @@
...
+                >
+                  {text}
+                </div>
+              </div>
+              {!isError && controls ? (
+                <div
+                  className={styles.controls}
+                  data-hook={SECTION_NOTIFICATION_DATA_HOOKS.controlsWrapper}
+                >
+                  {controls}
```
jonathanadler: We might need to have more control over the `icon`, `text` and `controls` in this component.
I'll have to check with the designers and I'll get back to you about this.

src/components/SectionNotification/SectionNotification.visual.tsx
```diff
@@ -0,0 +1,75 @@
...
+      ...getTests({ text: textSamples.long }),
+    ],
+  },
+  {
+    describe: 'mobile',
+    its: [
+      ...getTests({ text: textSamples.short, isMobile: true }),
+      ...getTests({ text: textSamples.long, isMobile: true }),
+    ],
+  },
```
jonathanadler: Let's add visual tests for all the permutations:
- only text
- text with icon
- text with one button
- text with two buttons

src/components/SectionNotification/SectionNotification.tsx
```diff
@@ -0,0 +1,67 @@
...
+                </div>
+              ) : null}
+            </div>
+          </div>
+        )}
+      </TPAComponentsConsumer>
```
jonathanadler: So there's a bit of a change I think we should make here.
It might seem big, but it shouldn't be too problematic. I hope 🙏 

First, from looking at the `st.css` file, I see that you used `flex-wrap` which is great.
This means we don't really need a special case for `mobile`, so we can remove the `TPAComponentsConsumer` and the `mobile` state.

Second, we need more control over the content that's inserted to this component.
What I suggest is to create a compound component, so that consumers will be able to use this component like this:
```jsx
<SectionNotification>
  <SectionNotification.Icon icon={<ErrorIcon/>} />
  <SectionNotification.Text>
    Lorem ipsum...
  </SectionNotification.Text>
  <SectionNotification.Button priority={PRIORITY.primary} onClick={() => {}}>
    Approve
  </SectionNotification.Button>
  <SectionNotification.Button priority={PRIORITY.secondary} onClick={() => {}}>
    Decline
  </SectionNotification.Button>
</SectionNotification>
```

So what this means is just moving the HTML you wrote, and move it to a "child" component:
```jsx
function NotificationText (props) {
  return (
    <div className={styles.text} 
         data-hook={SECTION_NOTIFICATION_DATA_HOOKS.textWrapper}>
      {props.children}
    </div>
  );
}
```
etc.

Regarding the button, it should render our Button component, and just pass the prop that it receives, so that consumers will have enough flexibility.
In addition, the button should also have an additional prop: `type`, with the values: `default` and `text`.
`default` should render a regular Button component, and the `text` should render a `TextButton` component.

All this let's us have control on what goes in the component, and it's styling.
For instance, the `SectionNotification.Button`'s will automatically have the margins between them as in the spec.

src/components/SectionNotification/SectionNotification.st.css
```diff
@@ -0,0 +1,98 @@
...
+/* Mobile styles */
+.root:mobile .controls {
+    margin-top: value(ContentPadding);
+    margin-left: 0;
+    margin-right: 0;
+}
+
+.root:mobile .main {
+    flex-direction: column;
+}
```
jonathanadler: image
Note that the margins don't seem to be correct
(between the icon and text)
and the alignment of the text

jonathanadler:
Hey @kristina1lyba 
I will start the review now (sorry for the delay).
But notice that there's a typescript error in a couple of your files:
```bash
[11:26:55]	[npm run test] /home/builduser/agent00/work/a62f8627ad515be/src/components/SectionNotification/SectionNotification.meta.tsx:17:9
[11:26:55]	[npm run test] ERROR: 17:9  jsx-wrap-multiline  Multiline JSX elements must be wrapped in parentheses
[11:26:55]	[npm run test] ERROR: 20:9  jsx-wrap-multiline  Multiline JSX elements must be wrapped in parentheses
[11:26:55]	[npm run test] 
[11:26:55]	[npm run test] /home/builduser/agent00/work/a62f8627ad515be/src/components/SectionNotification/docs/index.story.tsx:33:7
[11:26:55]	[npm run test] ERROR: 33:7  jsx-wrap-multiline  Multiline JSX elements must be wrapped in parentheses
[11:26:55]	[npm run test] ERROR: 36:7  jsx-wrap-multiline  Multiline JSX elements must be wrapped in parentheses
[11:26:55]	[npm run test] ERROR: 39:7  jsx-wrap-multiline  Multiline JSX elements must be wrapped in parentheses
```
If you could fix this, that would be great.
Thanks!

src/components/SectionNotification/SectionNotification.tsx
```diff
@@ -0,0 +1,58 @@
...
+      }
+    });
+
+    return (
+      <div
+        {...styles('root', {
+          error: isError,
+        })}
+        data-error={isError}
+        data-hook={this.props['data-hook']}
```
jonathanadler: the `data-hook` attribute isn't needed here, since stylable handles this for us.
(that's why we need to pass `this.props` to the stylable class)

src/components/SectionNotification/Text/SectionNotificationText.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { SECTION_NOTIFICATION_DATA_HOOKS } from '../dataHooks';
+import { SectionNotificationTextProps } from '../types';
+import styles from '../SectionNotification.st.css';
+
+export class SectionNotificationText extends React.Component<
+  SectionNotificationTextProps
+> {
+  static displayName = 'SectionNotification.Text';
+  static defaultProps = {
+    'data-hook': SECTION_NOTIFICATION_DATA_HOOKS.text,
```
jonathanadler: Here as well

src/components/SectionNotification/docs/examples.ts
```diff
@@ -0,0 +1,65 @@
+export const importExample = `
+  import { SectionNotification } from 'wix-ui-tpa/dist/src/components/SectionNotification';
+`;
+
+import { NOTIFICATION_TYPE } from '../types';
+
+const errorIcon = `<svg width="24" height="24" xmlns="http://www.w3.org/2000/svg" fillRule="evenodd" clipRule="evenodd"><path d="M12 0c6.623 0 12 5.377 12 12s-5.377 12-12 12-12-5.377-12-12 5.377-12 12-12zm0 1c6.071 0 11 4.929 11 11s-4.929 11-11 11-11-4.929-11-11 4.929-11 11-11zm.5 17h-1v-9h1v9zm-.5-12c.466 0 .845.378.845.845 0 .466-.379.844-.845.844-.466 0-.845-.378-.845-.844 0-.467.379-.845.845-.845z"/></svg>`;
```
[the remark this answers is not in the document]
jonathanadler: strange.
Let me see if I can fix that

src/components/SectionNotification/Button/SectionNotificationButton.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+  render() {
+    const { children, type, ...buttonProps } = this.props;
+    const ButtonComponent = (type === BUTTON_TYPE.text
+      ? TextButton
+      : Button) as React.ComponentClass;
+
+    return (
+      <div className={styles.button}>
+        <ButtonComponent data-hook={this.props['data-hook']} {...buttonProps}>
```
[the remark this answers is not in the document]
kristina1lyba: the wrapper div is needed to add spacing for the button as I can see Button component doesn't accept `className` prop

src/components/SectionNotification/docs/examples.ts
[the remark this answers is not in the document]
jonathanadler: This should work:

```jsx
const errorIcon = `<Error />`;
```
(Sorry I made a mistake earlier, it's `Error` and not `ErrorIcon`)
And you don't need to import it. 
It works just like this

src/components/SectionNotification/SectionNotification.st.css
```diff
@@ -0,0 +1,95 @@
...
+    min-width: value(TextMinWidth);
+    margin: 0 value(ContentPadding);
+}
+
+.text:first-child {
+    margin: 0;
+}
+
+.button  {
+    margin: 0 calc(value(ContentPadding) / 2);
```
jonathanadler: I think the margin here is incorrect.
In the zeroheight it looks like it's supposed to be `8px`, but here it's `4px`.

src/components/SectionNotification/docs/examples.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    ${
+      type !== NOTIFICATION_TYPE.error
+        ? `
+        <br/>
+        <SectionNotification type="${type}">
+          <SectionNotification.Icon icon={${errorIcon}} />
+          <SectionNotification.Text>${exampleText.confrirmation}</SectionNotification.Text>
+          <SectionNotification.Button type="text" key="primary" size={SIZE.tiny} priority={PRIORITY.basic}>Decline</SectionNotification.Button>
+          <SectionNotification.Button type="default" key="secondary" size={SIZE.tiny} priority={PRIORITY.basicSecondary}>Approve</SectionNotification.Button>
```
jonathanadler: Maybe, so the examples would be a bit more clear,
we can have the button's text to be `Text Button` for the text button
and `Regular Button` for the default button.
Just so the example would be a bit more clear.

jonathanadler: Can you also add an example with the `Primary` and `Secondary` Buttons, like in the zeroheight?
image
🙏 thanks

src/components/SectionNotification/SectionNotification.st.css
kristina1lyba: I know it might be hard to read but I did this intentionally.
To have support for RTL I had to make layout symmetrical, so I've introduced a bit complex margins and paddings so that looks fine on rtl, mobile, with only text, or with only buttons (any number) or all the elements together.
Here is what I mean:
Screen Shot 2020-01-15 at 12 22 31 PM
Screen Shot 2020-01-15 at 12 22 39 PM
Screen Shot 2020-01-15 at 12 22 54 PM

src/components/SectionNotification/docs/examples.ts
kristina1lyba: Added 👍

src/components/SectionNotification/SectionNotification.st.css
jonathanadler: You're right, I missed this!

jonathanadler:
@johannaasseraf @alexkronenberg 
Please add your notes here.
Thanks 🙏

johannaasseraf:
Hi @kristina1lyba here are our QA notes: https://www.notion.so/pdg/Section-Notification-QA-Notes-60efefc27f6d44a99fe4a18cf79e4804

You can leave us comments there if you need to ask something

jonathanadler:
Hey @kristina1lyba 
Note that the build is failing.
Looks like there's a typescript error in the `spec` file.
```bash
[15:51:50]        [npm run test] SectionNotification/SectionNotification.spec.tsx
[15:51:50]	[SectionNotification/SectionNotification.spec.tsx] Generic test suite failure
[15:51:50]	[Generic test suite failure] FAILED
[15:51:50]	[Generic test suite failure]   ● Test suite failed to run

    TypeError: Cannot read property 'default' of undefined

      19 |   static displayName = 'SectionNotification.Button';
      20 |   static defaultProps = {
    > 21 |     type: BUTTON_TYPE.default,
         |                       ^
      22 |   };
      23 | 
      24 |   textButtonDefaultProps = {

      at src/components/SectionNotification/Button/SectionNotificationButton.tsx:21:23
      at Object.<anonymous> (src/components/SectionNotification/Button/SectionNotificationButton.tsx:16:1)
      at Object.<anonymous> (src/components/SectionNotification/types.ts:1:1)
      at Object.<anonymous> (src/components/SectionNotification/SectionNotification.tsx:2:1)
      at Object.<anonymous> (src/components/SectionNotification/index.tsx:1:1)
      at Object.<anonymous> (src/components/SectionNotification/SectionNotification.spec.tsx:9:1)
[15:51:50]	[Step 5/6] Failed tests detected
```

kristina1lyba:
> Hey @kristina1lyba
> Note that the build is failing.
> Looks like there's a typescript error in the `spec` file.
> 
> ```shell
> [15:51:50]        [npm run test] SectionNotification/SectionNotification.spec.tsx
> [15:51:50]	[SectionNotification/SectionNotification.spec.tsx] Generic test suite failure
> [15:51:50]	[Generic test suite failure] FAILED
> [15:51:50]	[Generic test suite failure]   ● Test suite failed to run
> 
>     TypeError: Cannot read property 'default' of undefined
> 
>       19 |   static displayName = 'SectionNotification.Button';
>       20 |   static defaultProps = {
>     > 21 |     type: BUTTON_TYPE.default,
>          |                       ^
>       22 |   };
>       23 | 
>       24 |   textButtonDefaultProps = {
> 
>       at src/components/SectionNotification/Button/SectionNotificationButton.tsx:21:23
>       at Object.<anonymous> (src/components/SectionNotification/Button/SectionNotificationButton.tsx:16:1)
>       at Object.<anonymous> (src/components/SectionNotification/types.ts:1:1)
>       at Object.<anonymous> (src/components/SectionNotification/SectionNotification.tsx:2:1)
>       at Object.<anonymous> (src/components/SectionNotification/index.tsx:1:1)
>       at Object.<anonymous> (src/components/SectionNotification/SectionNotification.spec.tsx:9:1)
> [15:51:50]	[Step 5/6] Failed tests detected
> ```

That's a weird one... 
✔️ Fixed by re-arranging types exports

jonathanadler:
Thanks @kristina1lyba !

Issue #257, opened 2019-12-25
<Counter/> - a11y issues


ronnyrin:
We think the plus/minus buttons shouldn't be accessible via sr/tabs because the input should be enough. 
Also, it doesn’t make sense that the minus button would be the first focus.
In our current product page, we have a similar counter and we decided to add `aria-hidden={true}` to the plus/minus buttons. In our case the buttons are visible only on hover so we didn't have to add `tabIndex=-1` but i think we should in this case.
https://github.com/wix-private/ecom/blob/master/client/wixstores-client/wixstores-client-product-page/src/components/ProductOptions/NumberInputSpinner/NumberInputSpinner.tsx#L104-L125

jonathanadler:
@NeilWix 
should we make the buttons non-navigational?

NeilWix:
I think **sighted** keyboard users might find it awkward if the increment & decrement were not focusable, which brings me to the aria-hidden remark. Setting aria-hidden as true on interactive elements such as buttons is a big nono! Doing so would break UX for screen reader users as they would focus a link/button/etc and would only hear: link or button which is meaningless without a label (should be "add to car, button" or "sign in, link". 
So, whenever you wish to hide something from screen reader users you should also hide it from the tab-order with tabindex="-1".  This obviously come at the expanse of sighted keyboard users... so there is a tradeoff here.

As to the order of the elements, indeed it doesn't make sense, from a blind user point of view, to start with the decrementation option. This consideration should not be resolved at the expanse of sighted users who may find the current design instructive/intuitive. So to serve both audiences we can keep the current presentation order but implement it in a different DOM order via flex or grid.

jonathanadler:
@alexkronenberg

jonathanadler:
@ronnyrin @NeilWix @alexkronenberg
We decided on going with the following solution:
Changing the DOM order, so that the incrementation will be first, followed by the input, and then decrementation.
In addition, we'll add a `role="region"` to the wrapping div, and allow the TPA to pass a region aria-label.
What do you think Neil?

NeilWix:
@jonathanadler 
That's great. Would it also be possible to allow TPA's to also set (alternatively to aria-label) an aria-labelledby property? The rational is the case where there is already a labeling element entered for sighted users...

jonathanadler:
> @jonathanadler
[...]

PR #262, opened 2020-01-12
Counter component - Improve aria UX - issue #257

boriskaWix:
related discussion #257

### DIFF
--- a/src/components/Counter/Counter.driver.ts
+++ b/src/components/Counter/Counter.driver.ts
@@ -14,15 +14,17 @@ export interface CounterDriver extends BaseUniDriver {
   isInputValueEqualTo(val: number): Promise<boolean>;
   pressMinus(): Promise<void>;
   pressPlus(): Promise<void>;
+  getCounterAriaLabel(): Promise<string>;
 }
 
 export const counterDriverFactory = (base: UniDriver): CounterDriver => {
   const stylableUtil = new StylableUnidriverUtil(style);
 
-  const getMinusButton = () => base.$$(`.${style.btn}`).get(0);
-  const getPlusButton = () => base.$$(`.${style.btn}`).get(1);
+  const getPlusButton = () => base.$$(`.${style.btn}`).get(0);
+  const getMinusButton = () => base.$$(`.${style.btn}`).get(1);
   const getInput = () => base.$(`.${style.root} input`);
-
+  const getAriaLabel = (label: string) =>
+    base.$(`.${style.root}`).attr(`aria-${label}`);
   return {
     ...baseUniDriverFactory(base),
     isCounterComponentDisabled: async () =>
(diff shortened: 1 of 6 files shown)


src/components/Counter/Counter.tsx
```diff
@@ -65,19 +66,25 @@ export class Counter extends React.Component<CounterProps> {
...
       ...rest
     } = this.props;
 
     return (
-      <div {...style('root', { disabled, error }, rest)} dir="ltr">
+      <div
+        {...style('root', { disabled, error }, rest)}
+        dir="ltr"
+        role="region"
+        aria-labelledby={counterAriaLabel}
```
jonathanadler: Just a note:
I think that if we use this the prop for `aria-labelledby` then we should probably change its name to `counterAriaLabelledBy`.
And while we're at it, we should probably also accept a `counterAriaLabel` prop to pass as a `aria-label` to the root div.
This way, developers using this component can use either one of these props

src/components/Counter/Counter.driver.ts
```diff
@@ -14,15 +14,17 @@ export interface CounterDriver extends BaseUniDriver {
   isInputValueEqualTo(val: number): Promise<boolean>;
   pressMinus(): Promise<void>;
   pressPlus(): Promise<void>;
+  getCounterAriaLabel(): Promise<string>;
+  getCounterAriaLabellledBy(): Promise<string>;
```
jonathanadler: Note the typo (3 `l`'s)

src/components/Counter/Counter.tsx
```diff
@@ -11,6 +11,8 @@ export interface CounterProps extends TPAComponentProps {
   incrementAriaLabel: string;
   decrementAriaLabel: string;
   inputAriaLabel: string;
+  counterAriaLabelledBy?: string;
+  counterAriaLabel?: string;
```
jonathanadler: So, I hope i'm not driving you crazy, but after thinking about it,
I think it would be best to rename the props to be:
```jsx
'aria-labelledby'?: string;
'aria-label'?: string;
```
This goes with the convention and best practices we're trying to implement in all our libraries now.

boriskaWix: shouldn't props be in camelCase? just asking.

jonathanadler: That's true and a legit question!
When it comes to real html attributes, we have a convention of using the real attribute names.
that means using `data-*` and `aria-*` attributes as we would have used them in regular HTML, since React supports this.

jonathanadler: That being said, I am pondering if having a `regionAriaLabel` and `regionAriaLabelledBy` might be a more correct prop than the `aria-` props.
What do you think?

boriskaWix: after looking at other components in this project I see that this convention is being used throughout, so it'll be better to leave it like this and not brake the pattern. aria-labelledby it is?

jonathanadler: @boriskaWix 👍

jonathanadler:
@boriskaWix notice that there's a typescript error in the `meta` file.
It's causing the build to fail

PR #267, opened 2020-01-20
Fix: Text breaking

frontned:


src/components/Text/Text.st.css
```diff
@@ -23,6 +23,7 @@
 
 .root {
     -st-states: typography(enum(smallTitle, runningText, listText, largeTitle)), mobile;
+    word-break: break-word;
```
jonathanadler: I'm not sure this is the behavior we want.
In addition, this CSS rule is deprecated: https://developer.mozilla.org/en-US/docs/Web/CSS/word-break

I'll check with our designers

jonathanadler:
@alexkronenberg
This pull requests makes the text components' words to break in the middle of there isn't enough space.
It's this something we want?
You can see the specification here https://developer.mozilla.org/en-US/docs/Web/CSS/word-break

frontned:
What we are originally trying to fix with this pull request is this: 
image

alexkronenberg:
> What we are originally trying to fix with this pull request is this:
> image

@jonathanadler @frontned shouldn't this case have an ellipsis?

frontned:
@alexkronenberg if it would be the case then other titles for example 'Title Title Title Title Title Title Title' would also have and ellipsis, instead of now they break to another line.

frontned:
@alexkronenberg To be more clear, here is what I mean:
image

jonathanadler:
@alexkronenberg
The `Text` component doesn't have ellipsis.
I'm also not sure that we want that by default. For instance we wouldn't want it for running text, and such.
We can add an `ellipsis` prop perhaps, which can come in handy.
But this is also something that the TPAs can add themselves.

And regarding the case @frontned is talking about, what do you think should be their case?
Should they use ellipsis instead?

BTW, @frontned, thanks for the image, super helpful 🙏

frontned:
With out UX - Paulius Pamiskis we decided that the word-break solution will fit our case, because if we add ellipsis then normal titles with spaces will suffer. It's a question now, should we impelement it on our side and close this pull request?

PR #276, opened 2020-02-03
add isDisabled to dropdown driver

nayish:


nayish:
the Dropdown component has a disabled prop, this driver method allows you to check that it was used correctly

PR #279, opened 2020-02-06
States button remove div

NitzanSirota:
This PR was created because of a bug in the StatesButton - If you set the button's font through settings, and select an underline decoration, it won't show because of the inner StatesButton hierarchy, which puts the button's text inside a div.

### DIFF
--- a/src/components/StatesButton/StatesButton.st.css
+++ b/src/components/StatesButton/StatesButton.st.css
@@ -19,7 +19,7 @@
 .root::content {
     line-height: 1.5;
     display: block;
-    margin: 3px;
+    padding: 3px;
 }
 
 .successIcon {
--- a/src/components/StatesButton/StatesButton.st.css
+++ b/src/components/StatesButton/StatesButton.st.css
@@ -5,7 +5,6 @@
 
 .root {
     -st-extends: TPAButton;
-    padding:16px 8px 16px 8px;
     position: relative;
     cursor: pointer;
     display: block;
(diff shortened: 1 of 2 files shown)


NitzanSirota:
Found a bug - don't merge

jonathanadler:
> Found a bug - don't merge

What's the bug?

NitzanSirota:
There are 2 problems:
1. The .root::content override from the TPA button overrides the StatesButton override. So the margin is 0 auto instead of 3px.
2. When the checkmark is rendered (inside a div), it adds some kind of gutter. I'm researching why.

src/components/StatesButton/StatesButton.st.css
```diff
@@ -5,7 +5,6 @@
 
 .root {
     -st-extends: TPAButton;
```
jonathanadler: We also talked about adding a 
```css
-st-mixin: TPAButton(
  /* All the overridable variables */
)
```
right?

NitzanSirota: I think it's out of this specific PR's scope.. I'll quick check if this can be done easily or it supposed to be a separate feature

jonathanadler: I thought we wanted to add the underline text-decoration, no?

NitzanSirota: Taking the text out of the div solves it. No need for extra CSS.

jonathanadler: I see.
But we still can't merge though, right?

NitzanSirota: Yes, I'll push and update in the next few days

jonathanadler: 💪

src/components/StatesButton/StatesButton.st.css
```diff
@@ -19,7 +19,7 @@
 .root::content {
     line-height: 1.5;
     display: block;
-    margin: 3px;
+    padding: 3px;
```
NitzanSirota: @jonathanadler I've replaced margin with padding because the CSS here: https://github.com/wix/wix-ui-tpa/blob/master/src/components/Button/Button.st.css#L56 is overriding this CSS. In general, I don't like replacing them, but here it feels ok because it is on the content element and there's nothing between the content and the root element.

PR #280, opened 2020-02-10
#EE-22229| bug-fix: TAb on dropdown throws exception and removes elem…

varzager:
bug occurs on trying to reach id of selectedOption that is null

https://jira.wixpress.com/browse/EE-22229

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -69,6 +69,9 @@ export class Dropdown extends React.Component<DropdownProps, State> {
   };
 
   onSelect = (selectedOption: DropdownOptionProps) => {
+    if (!selectedOption) {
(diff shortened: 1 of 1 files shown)


jonathanadler:
@varzager 
Looks good to me.
Currently the build is failing because of a problem in a different library (that I caused 😬)
So i'll just wait for this to pass.

Just to understand the scenario in which the issue is caused,
This happens when tabbing out of the component with no option selected?

varzager:
https://www.screencast.com/t/dtjltZLQE

jonathanadler:
I'll merge this, since it fixes your issue (does it?).
However I do think this might be a problem with the ui-core component.
I'll add an issue for this.

Btw, just note that I'll also need to publish a new version after merging this.
So i'll update you when I do that.

PR #284, opened 2020-02-19
(Tabs) changed the svg fill role from black to currentColor

liorMar:


jonathanadler:
> Since this is somewhat of a breaking change, I prefer we override the color in the `Tabs` stylable file instead.
> That way it will be more specific, and no side effects
> Thanks!

liorMar:
@jonathanadler  it won't work if we will not change the fill role from specific Color (#000000) to currentColor.
the fill is stronger then css definition

jonathanadler:
> @jonathanadler it won't work if we will not change the fill role from specific Color (#000000) to currentColor.
> the fill is stronger then css definition

It should work.
See https://css-tricks.com/almanac/properties/f/fill/

liorMar:
I've noticed that in Pagination Component the issue is not accruing because it uses the tsx (it don't have fill rule) and not the svg.
if we changed to use the tsx too is it still considered as breaking changes?

jonathanadler:
> I've noticed that in Pagination Component the issue is not accruing because it uses the tsx (it don't have fill rule) and not the svg.
> if we changed to use the tsx too is it still considered as breaking changes?

The pagination icon color is defined inside `Pagination.st.css`.

The issue is that the `Chevron.svg` might be used by someone.
That's why I don't want to change it at the moment.
Since this is a hotfix, I prefer changing specifically the `Tabs` component.

liorMar:
So i'm suggesting to revert the change for the svg and change only the Tabs Component

jonathanadler:
> So i'm suggesting to revert the change for the svg and change only the Tabs Component

Yeah.
No need to change the SVG file.
Just add the fill color inside `Tabs.st.css` 👍

PR #286, opened 2020-02-20
feat(wix-ui-tpa a11y): add a11y aria attributes to dropdown

alex5il:


src/components/Dropdown/Dropdown.tsx
```diff
@@ -26,6 +26,8 @@ export interface DropdownProps {
   label?: string;
   alignment?: DROPDOWN_ALIGNMENT;
   placement?: Placement;
+  ariaLabel?: string;
+  ariaLabelledBy?: string;
```
jonathanadler: We're actually trying to start using `aria-` attributes the same way we would regularly use them in regular html.
Meaning naming the props: `aria-label` and `aria-labelledby`.
I should have told you this before, sorry 😬 
Other than that looks great, thanks!

jonathanadler:
Regarding the `aria-expanded` attribute, you're right, we should probably need to add a callback to the `wix-ui-core` component.
Something like `onOpen` and `onClose` or something like that.
It's a bit out of your scope, so you don't need to do it yourself.
Unless you want to 😁

alex5il:
Changed the naming :)
I think i might create a different PR later on for onClose / onOpen currently just that need to finish up some other tasks

jonathanadler:
@alex5il 
It seems the `labelledby` test is failing.
http://pullrequest-tc.dev.wixpress.com/viewLog.html?buildId=6485541&buildTypeId=FedInfra_WixUiTpa
I think it might be because the `DropdownBase` component isn't propagating to the button.
I'm not sure why the `aria-label` is propagating and the `aria-labelledby` isn't
But what I do know, is that stylable "catches" the  `aria-` attributes (when we pass the props in `...styles('root', {}, this.props)`) and attaches them to the root component.
I guess it specifically **doesn't** work for `aria-labelledby`.
Can you maybe try passing the `aria-labelledby` prop from the `DropdownBase` to the `button`?
Thanks!

alex5il:
@jonathanadler Hmm weird it pass in wallaby maybe it was cached 😓 
Ill check and push a fix thx! :)

PR #290, opened 2020-02-24
<Toast/> - change colors to be more accessible

bamai:
changed <Toast/> colors to be more accessible
@jonathanadler


jonathanadler:
Looks good.
I was asked by Harel to add the missing border which appears in the zeroheight but not in the component itself.
Could you add it please?

PR #291, opened 2020-02-27
dont include wix-ui-core definitions + copy paste definitions needed

jonathanadler:
@nitayneeman 
This works for me.

But what can the longterm solution be?


nitayneeman:
The issue was resolved in https://github.com/wix/yoshi/pull/2024, closing.

PR #292, opened 2020-03-01
States button refactor

NitzanSirota:
Adds states to StatesButton that are controlled by the consumer.


src/components/StatesButton/StatesButton.spec.tsx
```diff
@@ -1,58 +1,52 @@
...
-import { StatesButton } from './StatesButton';
-import { statesButtonDriverFactory } from './StatesButton.driver';
-
+import { StatesButton, StatesButtonProps } from './StatesButton';
+import {
+  StatesButtonDriver,
+  statesButtonDriverFactory,
+} from './StatesButton.driver';
+import { BUTTON_STATES } from './constants';
+//file.only
```
jonathanadler: Is this comment needed?

src/components/StatesButton/StatesButton.spec.tsx
```diff
@@ -62,6 +56,118 @@ describe('StatesButton', () => {
...
+            onClick={onClickSpy}
+          />,
+        );
+      });
+
+      it('should render failure content', async () => {
+        expect(await driver.getButtonTextContent()).toEqual('failure');
+      });
+
+      it('should invoke on click', async () => {
```
jonathanadler: Oh, so I think that in your case (Stores' case), you should use the `idle` state for your error, and have the components' error state behave like the success state.
Meaning that the failure state shouldn't be clickable.

src/components/StatesButton/StatesButton.tsx
```diff
@@ -44,18 +62,56 @@ export class StatesButton extends React.Component<
...
+      ...rest
+    } = this.props;
+    const inProgress = state === BUTTON_STATES.IN_PROGRESS;
+
     return (
       <Button
         data-hook={dataHook}
-        disabled={disabled}
-        {...rest}
+        disabled={this.isDisabled()}
```
jonathanadler: So from what I talked with @NeilWix, we don't want the button to be disabled.
The `aria-busy` is enough.
We do, however, need to block the `onClick` from being called.

src/components/StatesButton/StatesButton.tsx
```diff
@@ -44,18 +62,56 @@ export class StatesButton extends React.Component<
...
+      case BUTTON_STATES.IN_PROGRESS:
+        return inProgressContent;
+      case BUTTON_STATES.FAILURE:
+        return failureContent;
+      case BUTTON_STATES.SUCCESS:
+        return successContent ? successContent : this.renderCheck();
+      default:
+        return idleContent;
+    }
+  }
```
jonathanadler: According to the guidelines we got from neil:
https://docs.google.com/document/d/19BWUuPxIUQ4CLKL6KBG6KRsPmLutq_6RFtYimxWZ4kE/edit?usp=sharing

The different content types should be rendered with different tags.
`idle` and `loading` are buttons, but the `success` and `error` should just be divs (absolutely positioned).
While in the `success` and `error` states, the `div`'s and `button` should have a`tabindex=-1`.
We might probably want to add an `aria-label` to the `div`'s in case the content is of type `string`.

src/components/StatesButton/StatesButton.tsx
```diff
@@ -3,39 +3,57 @@ import * as classNames from 'classnames';
...
 export interface StatesButtonProps extends ButtonProps, TPAComponentProps {
+  state: BUTTON_STATES;
   disabled: boolean;
-  text: string;
+  idleContent: string | React.ReactElement;
+  failureContent?: string | React.ReactElement;
+  inProgressContent?: string | React.ReactElement;
+  successContent?: string | React.ReactElement;
+  onSuccessEnd?: Function;
   dataHook?: string;
```
jonathanadler: While we're at it, you can remove the `dataHook` prop.
The `TPAComponentProps` type already introduces a `data-hook` prop (which is the preferred syntax).

src/components/StatesButton/StatesButton.spec.tsx
NitzanSirota: Nope, accidentally got into the code... good catch :)

src/components/StatesButton/StatesButton.tsx
NitzanSirota: According to Johanna's spec, it is supposed to be disabled and also look disabled when in progress.
Basically we can leave it to the consumer to decide, but I'm not sure if that's what Johanna wants..
WDYT?

jonathanadler: Johanna's spec is design only.
This is the a11y behaviour that we agreed with Neil.
BTW, she told she's not sure she wants the button to look disabled.
And I told her that technically the button won't be disabled.

PR #293, opened 2020-03-04
<IconToggle/> - Add animations and hover behaviour

bamai:
Added animations and hover behaviour according to https://www.notion.so/IconToggle-LikeButton-review-issues-bf271e515c964095835ac6a4a963ab11.
@jonathanadler


src/components/LikeButton/LikeButton.tsx
```diff
@@ -10,42 +10,25 @@ export interface LikeButtonProps extends TPAComponentProps {
   checked: boolean;
   onChange?(event: OnChangeEvent): void;
   disabled: boolean;
+  animation: boolean;
```
jonathanadler: I think the `LikeButton` should be animated by default.
So there's no need to add a prop to it, since we don't want users to disable the animation.

src/components/IconToggle/IconToggle.tsx
```diff
@@ -19,22 +19,45 @@ export interface IconToggleProps extends TPAComponentProps {
   onChange?(event: OnChangeEvent): void;
   checked: boolean;
   disabled: boolean;
+  animation: boolean;
```
jonathanadler: I don't remember if we talked about adding an `animation` property.
Even though it's a great idea to avoid introducing a breaking change,
I think in this case we don't need to add the prop.
The animation can be by default.

src/components/IconToggle/IconToggle.tsx
```diff
@@ -50,12 +73,16 @@ export class IconToggle extends React.Component<IconToggleProps> {
   };
 
   render() {
-    const { labelPlacement, onChange, checked, disabled, ...rest } = this.props;
+    const { labelPlacement, onChange, checked, disabled, animation, ...rest } = this.props;
+
+    const { animated } = this.state;
```
jonathanadler: I think we can have the same animation without `this.state.animated`.
Maybe we can try to apply the animation using the `animation-direction` css rule.

src/components/IconToggle/IconToggle.tsx
```diff
@@ -50,12 +61,12 @@ export class IconToggle extends React.Component<IconToggleProps> {
...
   render() {
-    const { labelPlacement, onChange, checked, disabled, ...rest } = this.props;
+    const { labelPlacement, onChange, checked, disabled, animation, ...rest } = this.props;
 
     const content = this._getContent();
-
+    
     return (
-      <span {...styles('root', { checked, disabled, labelPlacement }, rest)}>
+      <span {...styles('root', { checked, disabled, labelPlacement, animation }, rest)} onMouseLeave={this._handleHoverOff}>
```
jonathanadler: @bamai 
Remind me...
We said it might be possible to remove the component's state, right?
And just use the prop for the animation.
(I mentioned `animation-direction`).
What did we decide about this?

bamai: I think we decided to remove the state (the last commit is supposed to work without using internal state)

jonathanadler: Oh, ok, so I guess we can remove the `onMouseLeave` and the `State` interface

PR #298, opened 2020-03-05
<Button/> - upgrade styling

bamai:
@jonathanadler


src/components/Button/Button.st.css
```diff
@@ -189,3 +189,58 @@
     padding: 10px 16px;
     font: fallback(font(value(MainTextFont)), font(value(LargeFont)));
 }
+
+:vars{
+    secondaryBorderWidth: fallback(value(MainBorderWidth), 1px);
```
jonathanadler: I think you're missing a definition for `primaryBorderWidth`, no?
Or maybe you meant to use `PrimaryBorderWidth` (with capital 'P')?

src/components/Button/Button.st.css
```diff
@@ -189,3 +189,58 @@
...
+/*upgrage*/
+.root:upgrade{
+    box-sizing: border-box;
+    padding-left: 16px;
+    padding-right: 16px;
+}
+
+/*upgrade tiny*/
+.root:priority(secondary):upgrade:size(tiny), .root:priority(basicSecondary):upgrade:size(tiny){
+    padding-top: "calculate(-, 6px, value(secondaryBorderWidth))";
```
jonathanadler: Ok, so I found the problem.
First, you should remove the string from all the `calculate` experssions.
i.e. write: 
```padding-top: calculate(...)```
In addition, it seems there's a problem with 
```PrimaryBorderWidth: fallback(value(MainBorderWidth), 0);```
It should be:
```PrimaryBorderWidth: fallback(value(MainBorderWidth), 0px);```
(Notice the `px` at the end)

**Comment by jonathanadler** [comment]
@bamai 
So in order to make this PR pass, we will need to make a change in the [`Button.driver.ts`](https://github.com/wix/wix-ui-tpa/blob/master/src/components/Button/Button.driver.ts) file.

Notice that the driver is using the `StylableUnidriverUtil`.
This util reads the `st.css` file and can help us detect which state the component is in.
Sounds helpful, but unfortunately doesn't work with `e2e` tests.

So to fix our issue, we need to remove this util, and instead add `data-*` attributes in order to detect states.

You can see an example for how we use the `data-*` attributes, in the `Tabs.driver.ts` file here: https://github.com/wix/wix-ui-tpa/blob/2c2983fca6bb45e596686822c9de31f3413e01fd/src/components/Tabs/Tabs.driver.ts#L36
 
After fixing this, everything should work, but I would make another simple change.
Currently we are using the `Button.e2e.ts` test to take a snapshot, using a tool developed here at Wix, called `eyes.it`.
This tool takes a snapshot of a site with the button.

We prefer using the `*.visual.tsx` files for this.

So it would be best to create a `Button.visual.tsx` file for the visual test, instead of the `Button.e2e.ts` test.

Ping me if you have any questions

File: src/components/Button/Button.tsx [hunk 3]
```diff
@@ -27,6 +27,7 @@ export interface ButtonProps extends ButtonNextProps, TPAComponentProps {
   size?: SIZE;
   fullWidth?: boolean;
   innerRef?: React.RefObject<HTMLButtonElement>;
+  upgrade?: boolean;
 }
 
 class ButtonComponent extends React.Component<ButtonProps> {
```
jonathanadler: We also need to add a `console.warn` with a warning if consumers aren't using the `upgrade` prop.
You can see the implementation of the `Card` component.

src/components/Button/Button.tsx
```diff
@@ -36,17 +37,39 @@ class ButtonComponent extends React.Component<ButtonProps> {
...
     fullWidth: false,
+    upgrade: false,
   };
 
+  _getDataAttributes(mobile) {
+    const { fullWidth } = this.props;
+
+    return {
+      ['data-fullwidth']: fullWidth,
+      ['data-mobile']: mobile,
```
jonathanadler: So actually the `[`/`]` aren't needed here :)

src/components/Button/Button.visual.tsx
```diff
@@ -0,0 +1,27 @@
...
+  story(`Size`, () => {
+    Object.values(SIZE).forEach(size => {
+      snap(`${size}`, () => <Button size={size}>{size}</Button>);
+    });
+  });
+}
+
+visualize('Button', () => {
+  priorityTestSuite();
+  sizeTestSuite();
```
jonathanadler: Oh, now that I think about it, we should probably add a test suite with the `upgrade` on.
Didn't think about that earlier

johannaasseraf:
@bamai Please check for each default size: font size, line height, and padding are according to this table:
https://zeroheight.com/7sjjzhgo2/p/9607e0-02-buttons/t/348932

Also live example here: https://codepen.io/johanna87/pen/BaNwroP

Thanks!

src/components/StatesButton/StatesButton.st.css
```diff
@@ -21,6 +22,10 @@
     display: block;
 }
 
+.root:upgrade::content{
+    line-height: 1;
```
jonathanadler, line 26: This needs to be added:
```suggestion
    line-height: 1;
    display: inline-block;
```

PR #299, opened 2020-03-08
<Badge/> - change default colors

bamai:
Changed Badge default colors according to https://www.notion.so/Update-Default-Badge-Wiring-in-Storybook-540ac22d87e84892a337eb842a4207c1.
@jonathanadler

### DIFF
--- a/src/components/Badge/Badge.st.css
+++ b/src/components/Badge/Badge.st.css
@@ -18,13 +18,11 @@
     /*Overrides*/
     BadgeBgColor: --overridable;
     BadgeTextColor: --overridable;
-    BadgeBorderColor: --overridable;
 }
 
 :vars {
     badgeBgColor: color(fallback(value(BadgeBgColor), applyOpacity(color(value(DefaultBgColor)), 0.1)));
     badgeTextColor: color(fallback(value(BadgeTextColor), value(DefaultTextColor)));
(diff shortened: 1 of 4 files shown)


src/components/Badge/Badge.tsx
```diff
@@ -13,8 +13,13 @@ export interface BadgeProps extends TPAComponentProps {
   priority?: BADGE_PRIORITY;
 }
 
+export interface DefaultProps{
```
jonathanadler: I don't think we need this to be exported.
And It causes the build to fail.
So you can just remove the `export`.

jonathanadler:
Looks good!
@bamai Can you ask @alexkronenberg to give a design review?

src/components/Badge/Badge.st.css
```diff
@@ -36,23 +33,21 @@
...
     border: none;
 }
 
 .root:priority(light) {
     background-color: color(value(LightBgColor));
     color: color(value(LightTextColor));
     border-color: value(lightBorderColor);
+    box-sizing: border-box;
+    border-width: 1px;
+    border-style: solid;
```
jonathanadler: I just noticed, that since this priority doesn't have a border, and the others do, we should probably change the `line-height` here.
So if the `.root` `line-height` is `20px`, the `light` priority should probably have `18px` to compensate for the border.
This will make the three priorities have the same height

PR #305, opened 2020-03-15
Tags component

liorMar:


src/components/Tags/Tags.tsx
```diff
@@ -0,0 +1,103 @@
...
+  static displayName = 'Tags';
+  static defaultProps: DefaultProps = {
+    alignment: ALIGNMENT.center,
+    size: SIZE.medium,
+    skin: SKIN.solid,
+  };
+
+  private getTagDataAttributes({ isActive, index }) {
+    return {
+      [TAGS_DATA_KEYS.tabIsActive]: isActive,
```
jonathanadler: `active` => `selected`/`checked`

src/components/Tags/Tags.tsx
```diff
@@ -0,0 +1,103 @@
...
+        checked={item.checked}
+        onChange={() => onClick(item)}
+        value={item.value}
+        className={styles.tag}
+        data-hook={`${TAGS_DATA_HOOKS.tag}-${index}`}
+        {...this.getTagDataAttributes({
+          isActive: item.checked,
+          index,
+        })}
+        {...styles('tag', { selected: item.checked })}
```
jonathanadler: unfortunately, using a state in a sub component in stylable is a bit buggy.
It can cause some issues.
Maybe not at the moment, but once consumers will pass `classNames`, and if we'd like to pass `classNames` to the specific tag, that might cause problems.
So I think for now maybe just use a `selected` class instead of a stylable state.

src/components/Tags/Tags.tsx
```diff
@@ -0,0 +1,103 @@
...
+      [TAGS_DATA_KEYS.skin]: skin,
+      [TAGS_DATA_KEYS.size]: size,
+      [TAGS_DATA_KEYS.rtl]: rtl,
+    };
+  }
+
+  generateTagItems() {
+    const { items, onClick } = this.props;
+    return items.map((item, index) => (
+      <CoreTag
```
jonathanadler: We should probably also pass the rest of the core tag's props.
For instance the `rel` prop should be passed as well.

src/components/Tags/Tags.st.css
```diff
@@ -0,0 +1,139 @@
...
+    SelectedTextColor: --overridable;
+    SelectedBackgroundColor: --overridable;
+    SelectedBorderColor: --overridable;
+}
+
+.root {
+    -st-states: alignment(enum(center, left, right)),
+                size(enum(small, medium, large)),
+                skin(enum(light, solid)),
+                mobile,
```
jonathanadler: I think we can remove this state, right?
It's not being used

src/components/Tags/Tags.visual.tsx
```diff
@@ -0,0 +1,77 @@
...
+class TagsVisual extends React.Component<TagsVisualProps> {
+  static defaultProps = {
+    tagsProps: {},
+    rtl: false,
+  };
+  private readonly driver: TagsDriver;
+
+  render() {
+    const { tagsProps, rtl, compact } = this.props;
+    const style = compact ? { margin: '10px', maxWidth: 200 } : undefined;
```
jonathanadler: I don't see where `compact` is passed as `true`.

src/components/Tags/Tags.visual.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+    return (
+      <TPAComponentsProvider value={{ rtl }}>
+        <VisualTestContainer>
+          <div style={style}>
```
jonathanadler: Maybe we can have each test render the different types of sizes and skins

src/components/Tags/docs/TagsTestStory.tsx
```diff
@@ -0,0 +1,24 @@
...
+    onClick: () => {},
+  };
+  return (
+    <div style={{ margin: '10px', maxWidth: 200 }}>
+      <Tags data-hook={'storybook-e2e-Tags'} {...defaultProps} {...props} />
+    </div>
+  );
+}
+
+storiesOf(kind, module).add('Tags', renderTest);
```
jonathanadler: Is this story used somewhere 🤔 ?

stories/index.tsx
```diff
@@ -41,3 +41,5 @@ require('../src/components/SocialBar/docs/index.story');
 require('../src/components/ShareButton/docs/index.story');
 require('../src/components/CopyUrlButton/docs/index.story');
 require('../src/components/SectionNotification/docs/index.story');
+require('../src/components/Tags/docs/index.story');
+require('../src/components/Tags/docs/TagsTestStory');
```
jonathanadler: I'm not sure this is needed 🤔

src/components/Tags/Tags.driver.ts
```diff
@@ -0,0 +1,32 @@
...
+  getSize(): Promise<string>;
+}
+
+const getTag: any = (base: UniDriver, index: number) =>
+  base.$(`[data-hook="${TAGS_DATA_HOOKS.tag}-${index}"]`);
+
+export const tagsDriverFactory = (base: UniDriver): TagsDriver => {
+  return {
+    ...makeTagsListUniDriver(base),
+    clickTagAt: index =>
```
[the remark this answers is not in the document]
liorMar: i wanted but the interface missing the index argument and type script failing because of it
i'll create a PR and change when it is merged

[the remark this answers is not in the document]
liorMar: https://github.com/wix/wix-ui/pull/1807

src/components/Tags/dataHooks.ts
```diff
@@ -0,0 +1,14 @@
...
+}
+
+export enum TAGS_DATA_KEYS {
+  alignment = 'data-alignment',
+  skin = 'data-skin',
+  size = 'data-size',
+  navButtons = 'data-navbuttons',
+  rtl = 'data-rtl',
+  index = 'data-tab-index',
+  tabIsActive = 'data-tab-active',
```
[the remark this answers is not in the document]
liorMar: i prefer checked, it is a checkbox

src/components/Tags/Tags.visual.tsx
```diff
@@ -0,0 +1,91 @@
...
+        items={generateItems(FEW_ITEMS_COUNT)}
+        onClick={() => {}}
+        size={size}
+      />
+    ));
+  });
+}
+
+function skinTestSuite(skin?) {
+  story(`Size ${skin ? skin : 'default'}`, () => {
```
jonathanadler: Typo: `Size` should be changed to `Skin`

liorMar: fixed

src/components/Tags/Tags.visual.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+const MANY_ITEMS_COUNT = 20;
+const FEW_ITEMS_COUNT = 3;
+
+function generateItems(amount) {
+  return new Array(amount).map(i => ({
+    title: `Title ${i}`,
+    value: `value ${i}`,
+    checked: i % 3 === 0,
+  }));
```
jonathanadler:
```suggestion
  return new Array(amount).fill(0).map(i => ({
    title: `Title ${i}`,
    value: `value ${i}`,
    checked: i % 3 === 0,
  }));
```

jonathanadler: It seems the visual tests are breaking because of this.

jonathanadler:
@liorMar 
I've got the review from the designers:

1. remove animation
2. selected color overrides doesn't seem to work
3. I think selected border color should be the same as the bg color by default
4. add a state to the example in the settings panel, so that visitors can click on the tags
5. restrict border size in the example to a max of 10px (that's the guidelines and we don't want developers think they should give a bigger border than that).

That's all thanks!

src/components/Tags/Tags.visual.tsx
```diff
@@ -0,0 +1,91 @@
...
+    snap('RTL', () => (
+      <TPAComponentsProvider value={{ rtl: true }}>
+        <Tags
+          items={generateItems(FEW_ITEMS_COUNT)}
+          onClick={() => {}}
+          alignment={alignment}
+        />
+      </TPAComponentsProvider>
+    ));
+  });
```
jonathanadler: Oh and another thing.
You're right, there is a problem with the `rtl` visual tests.
It's not working as expected (you can see it here: https://wixeyes.applitools.com/app/test-results/00000251817847633468/?accountId=mE99eP8CUn9UTSXSv98L3OSaJ0NZmAcY6Wndx6WSHODdY110)

So first, I think there should be one `rtl` test, i'm not sure that applitools is case-sensitive.
And second, the test should both have a `<div dir="rtl">...</div>` and also the `<TPAComponentsProvider>` together.
i.e.:
```jsx
<div dir={'rtl'}>
  <TPAComponentsProvider value={{ rtl: true }}>
    <Tags
      items={generateItems(FEW_ITEMS_COUNT)}
      onClick={() => {}}
      alignment={alignment}
    />
  </TPAComponentsProvider>
</div>
```

PR #308, opened 2020-03-19
Tags override spacing

liorMar:
i also exported the TagItem interface and moved tags to be after tabs in the storybook


src/components/Tags/Tags.st.css
```diff
@@ -80,27 +80,43 @@
 .tag {
     padding: 0 value(defaultTagInnerPadding);
     max-width: value(defaultMaxWidth);
-    border-style: solid;
     margin: fallback(value(MainVerticalSpacing), value(defaultTagVerticalSpacing)) fallback(value(MainHorizontalSpacing), value(defaultTagHorizontalSpacing));
-    box-sizing: content-box;
+    display: flex;
+    align-items: center;
+    box-sizing: border-box;
```
jonathanadler: So actually, the guidelines we received is that it should be `content-box`.
That's how it was implemented in `wix-ui-santa`.
I'm checking to see if we can leave this like this or not, and I'll get back to you.

liorMar: I'm changing back to content box
the main reason for border box was to make the border grow inside

jonathanadler: Cool 👍 

Btw, I think we also we forgot to add the option to override the font: https://zeroheight.com/7sjjzhgo2/p/05a744-05-selections/t/9906bf

Sizes have different default font-sizes but they should be able to be overridden

liorMar: but i ended in a different solution

liorMar: font is already overridable one problem that i found was in that i set the height of the box to be from the lineHeight in the font definition but once user override it the lineHeight is gone so i removed the lineHeight and set min height for larger font size

liorMar: there some things that only after you try do use them you find the initial implementation was flawed

liorMar: in wix-base-ui font theme and font size is a total package you have to define them both when you override

jonathanadler: I think what we had was ok.
If the user changes the font, then it will change.
It will no longer be a specific size or theme.
So I think you can revert the min-height additions.

In addition, we need to add the font control in the storybook's settings panel

src/components/Tags/Tags.st.css
```diff
@@ -12,10 +12,11 @@
 :vars {
     /*Defaults params goes here*/
     defaultTagInnerPadding: 12px;
+    defaultMaxBorderWidth: 10px;
```
jonathanadler: btw, I don't think this is needed, or used, right?

src/components/Tags/Tags.st.css
```diff
@@ -12,10 +12,11 @@
...
     defaultMaxWidth: 220px;
-    defaultSmallFont: "{theme: 'Body-M', size: '12px', lineHeight: '24px'}";
-    defaultMediumFont: "{theme: 'Body-M', size: '14px', lineHeight: '32px'}";
-    defaultLargeFont: "{theme: 'Body-M', size: '16px', lineHeight: '36px'}";
+    defaultSmallFont: "{theme: 'Body-M', size: '12px'}";
+    defaultMediumFont: "{theme: 'Body-M', size: '14px'}";
+    defaultLargeFont: "{theme: 'Body-M', size: '16px'}";
```
jonathanadler: I think what we had was good.
So you can add the line-heights back.

liorMar: restored

src/components/Tags/Tags.st.css
liorMar: Screen Shot 2020-03-22 at 15 34 47

this is what happening when the user overrides the font and no min-height is set
if we won't deal with it in the component then each user will need to set it itself

src/components/Tags/docs/index.story.tsx
```diff
@@ -171,6 +171,20 @@ export default {
                 ],
                 fonts: [],
```
jonathanadler: Please add a control to change fonts

src/components/Tags/Tags.st.css
jonathanadler: Then maybe we can have a default padding, no?
I think it will work well with the outer border.

src/components/Tags/docs/index.story.tsx
liorMar: i also added a font size definition
it drive me crazy seeing the size increase to 30px

src/components/Tags/docs/index.story.tsx
```diff
@@ -169,8 +169,29 @@ export default {
...
+                    unit: 'px',
+                    max: 30,
+                  },
+                  {
+                    label: 'Horizontal Spacing',
+                    wixParam: 'horizontalSpacing',
+                    defaultNumber: 4,
+                    unit: 'px',
+                    max: 30,
+                  },
```
jonathanadler:
```suggestion
                  {
                    label: 'Vertical Spacing',
                    wixParam: 'verticalSpacing',
                    defaultNumber: 4,
                    unit: 'px',
                    max: 20,
                  },
                  {
                    label: 'Horizontal Spacing',
                    wixParam: 'horizontalSpacing',
                    defaultNumber: 4,
                    unit: 'px',
                    max: 20,
                  },
```
Harel asked to limit the max to `20px`

PR #309, opened 2020-03-24
#EE-22608 | disallow decimal points in a counter input

sagish:
Allow `inputType` as a prop

A `number` input does not trigger an `onChange` event, therefore, we cannot sanitize it

https://jira.wixpress.com/browse/EE-22608

### DIFF
--- a/src/components/Counter/Counter.spec.tsx
+++ b/src/components/Counter/Counter.spec.tsx
@@ -144,23 +144,6 @@ describe('Counter', () => {
     expect(onChange).toHaveBeenCalledWith('10');
   });
 
-  it('should sanitize a typed input', async () => {
-    const onChange = jest.fn();
-    const driver = createDriver(
-      <Counter
-        inputAriaLabel={'amount'}
-        incrementAriaLabel={'increment'}
-        decrementAriaLabel={'decrement'}
-        onChange={onChange}
-        step={1}
-        value={0}
-      />,
-    );
-
-    await driver.enterValue('10.');
-    expect(onChange).toHaveBeenCalledWith('10');
-  });
-
   it('should have default role attribute and aria-labelledby attribute', async () => {
     const value = 10;
     const counterAriaLabelledBy = 'labelled by testy test';
--- a/src/components/Counter/Counter.spec.tsx
+++ b/src/components/Counter/Counter.spec.tsx
@@ -144,6 +144,23 @@ describe('Counter', () => {
     expect(onChange).toHaveBeenCalledWith('10');
   });
 
+  it('should sanitize a typed input', async () => {
+    const onChange = jest.fn();
+    const driver = createDriver(
+      <Counter
+        inputAriaLabel={'amount'}
+        incrementAriaLabel={'increment'}
+        decrementAriaLabel={'decrement'}
+        onChange={onChange}
+        step={1}
+        value={0}
+      />,
+    );
+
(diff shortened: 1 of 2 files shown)


ronnyrin:
not sure it should be in the library. It can also be used as price component and then points are allowed.

sagish:
In that case, I still need to control the input type from outside, otherwise, I cannot sanitize the input from our end since `onChange` is not triggered when a decimal point is typed when the input type is `number`.

PR updated.

jonathanadler:
I think it would be better to replace the `onChange` callback with `onInput`.
It should give you an option to validate the input.
Also, note that a `step` prop exists for decimals

sagish:
@jonathanadler I removed the decimals sanitization

Re `onInput`, sadly, while in `type: number` the decimal is not registered as a new input (`10.` in registered as `ev.target.value=10` so it's not useful, but only in `text`)

jonathanadler:
> @jonathanadler I removed the decimals sanitization
> 
> Re `onInput`, sadly, while in `type: number` the decimal is not registered as a new input (`10.` in registered as `ev.target.value=10` so it's not useful, but only in `text`)

But that's ok, because `10. === 10`.
Once the user adds a digit after the decimal, you'll get an update, and you can show an error

jonathanadler:
Closed by mistake 😬

sagish:
In that case `onChange` is good enough as well, thing is, you normally don't want to allow any `\D` chars on a number input, and we cannot get that with the `number` type.

@ronnyrin WDYT?

jonathanadler:
> In that case `onChange` is good enough as well, thing is, you normally don't want to allow any `\D` chars on a number input, and we cannot get that with the `number` type.
> 
> @ronnyrin WDYT?

That's not true.
`onChange` fires only when the user blurs out from the input, and `onInput` fires on any `\d` input the user enters.
regardless of which event handler we choose, the `number` type restricts the user from entering any non-digit (+ `.`) character. 
With the `text` type, you'll need to validate the value yourself on every character the user enters to the input.

Again, the `.` character is a valid number character.

sagish:
In this case, `.` cannot be validated

@ronnyrin

sagish:
@jonathanadler why not allow us to control the input type?

jonathanadler:
> @jonathanadler why not allow us to control the input type?

In order to confirm this pull request I need to understand why you need to control the input type.
At the moment I don't see a reason

sagish:
We don't want to allow a decimal point in our quantity input

jonathanadler:
But a `text` type will also allow entering a decimal dot

sagish:
A decimal point in `text` triggers `onChange` and `onInput` and therefore can be sanitized; in `number` it does not trigger any callback and therefore appear in the input without being able to do anything about it

jonathanadler:
> A decimal point in `text` triggers `onChange` and `onInput` and therefore can be sanitized; in `number` it does not trigger any callback and therefore appear in the input without being able to do anything about it

Ok, but it still allows entering a decimal point.
So your actual issue is not with entering the decimal point itself, it's sanitizing it afterwards.
And i'm telling you that if a user enters `10.` you don't need to sanitize it.

I have lot's of things to work on, so let's make this shorter.
You can either change the `onChange` to an `onInput` or we can close this PR.

sagish:
I don't think an "I don't have time for this" is an appropriate response.

I'm opting out ✌️

PR #311, opened 2020-03-30
Dropdown native select

NitzanSirota:
Rendering <select> element in mobile if shouldRenderNativeSelectOnMobile is passed with "true" value


src/components/Dropdown/Dropdown.st.css
```diff
@@ -27,6 +27,11 @@
 :import {
     -st-from: "./DropdownBase.st.css";
     -st-default: DropdownBase;
+}
+
+:import {
+    -st-from: "./DropdownNativeSelect.st.css";
+    -st-default: DropdownNativeSelect;
     -st-named: borderHoverColor;
```
jonathanadler, line 35: Is there a reason why `borderHoverColor` should be implemented in `DropdownNativeSelect` as well as the `DropdownBase`?

src/components/Dropdown/DropdownNativeSelect.st.css
```diff
@@ -0,0 +1,165 @@
...
+    background-color: transparent;
+}
+
+.root:mobile:size(medium):icon {
+    padding: 8px 36px;
+}
+
+.root:mobile:rtl {
+    /*background: value(backgroundColor) no-repeat right 13px center url("../../assets/icons/CaretDown.svg");*/
+    background-image: url("../../assets/icons/CaretDown.svg");
```
jonathanadler, line 68: Do we use this method only in `rtl`?
Didn't we say we can use an absoluted div instead?

src/components/Dropdown/DropdownNativeSelect.tsx
```diff
@@ -0,0 +1,150 @@
...
+    const { placeholder, options } = this.props;
+    return (
+      <>
+        {hasPlaceholder && (
+          <option data-hook={DATA_HOOKS.placeholderOption} value="" disabled>
+            {placeholder}
+          </option>
+        )}
+        {options.map((option, i) => (
+          <option key={i} value={option.id} disabled={!option.isSelectable}>
```
jonathanadler, line 53: Shouldn't the selected option have `selected={option.selected}`, or something like that?

src/components/Dropdown/Dropdown.st.css
NitzanSirota: actually - no 😂

src/components/Dropdown/DropdownNativeSelect.st.css
NitzanSirota: the absolute div replaces this css, forgot to delete it.
(another) nice catch

src/components/Dropdown/DropdownNativeSelect.tsx
NitzanSirota: I've read this: https://reactjs.org/docs/forms.html#the-select-tag
And I'm still not sure if it needs to be added. 
The selection(show which option is selected) is working by default with <select> so I'm not sure what is the added value

jonathanadler: the difference between using the `selected` attribute vs. not using it,
is having a controlled component vs an uncontrolled component.
In our case, this component is an uncontrolled component, so you're right that maybe in this case (until we merge that PR I made) we don't need it.
However, I think we are using the state's `selectedOption` property for the selected option.
So that the native select element will be aligned to the custom desktop component.

NitzanSirota: Just to be clear - right now we are not using the state's selected option in the native dropdown to find out which of the options is selected. We have the placeholder/ first option as defaults, and after first selection, the <select> is functioning as an uncontrolled component and managing the state by itself

jonathanadler: yeah, so I was thinking just to have it aligned with the desktop version, maybe it would be best to use it as well.

In addition, the `initialSelectedId` prop should be used for the first option selected.
And some might use it for changing the selected option.

So I guess it's best to keep it aligned so that the behaviour will be identical.

And when i'll migrate it to the controlled version i'll change both

jonathanadler:
btw, I merged a PR that is conflicting with yours 😬 sorry
It shouldn't be too hard to fix I think

NitzanSirota:
Rebased the branch and added one native storybook example

src/components/Dropdown/Dropdown.st.css
jonathanadler: So maybe we can move it back 🤔 ?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -51,10 +54,12 @@ interface State {
  * */
 export class Dropdown extends React.Component<DropdownProps, State> {
   static displayName = 'Dropdown';
+  static contextType = TPAComponentsContext;
   static defaultProps: DefaultProps = {
     placeholder: '',
     options: [],
     placement: 'bottom',
+    shouldRenderNativeSelectOnMobile: false,
```
jonathanadler, line 62: Let's change this to be `true` by default.
This prop will give consumers a backdoor

jonathanadler: And maybe let's change it to just `mobileNativeSelect` 😬 ?

src/components/Dropdown/DropdownNativeSelect.st.css
jonathanadler: Still needs to be deleted no? 🤔

src/components/Dropdown/DropdownNativeSelect.tsx
jonathanadler: What did we decide here?

src/components/Dropdown/Dropdown.tsx
NitzanSirota: that is definitely a better name. I'll change it.

NitzanSirota:
Some of the CR fixes I made we're reverted during the rebase 🤦 😢 
I'll re-commit them.

src/components/Dropdown/Dropdown.driver.ts
```diff
@@ -73,6 +85,84 @@ export const dropdownDriverFactory = (base: UniDriver): DropdownDriver => {
...
+      (await (await getDropdownNativeSelect(base)).attr(
+        DATA_ATTRIBUTES.error,
+      )) !== null,
+    getErrorMessageContent: async () => {
+      const tooltipDriver = await getTooltipDriver(baseUniDriver);
+      tooltipDriver.mouseEnter();
+      return tooltipDriver.getContentElement().innerHTML;
+    },
+    getDropdownText: async () =>
+      (await getNativeSelectElement()).selectedOptions[0].text,
```
jonathanadler, line 134: hmmm, I think this means this won't work for e2e tests.
When you called `native()` up there ☝️ the unidriver returns the environment driver.
In our tests (mocha) the native driver can fetch the `selectedOptions`, because it has access to the dom (jsdom in this case).
But in e2e tests it won't work.

I think we need to use Unidriver's `_prop` method instead:
https://github.com/wix-incubator/unidriver/blob/20c9935d2be9ca3caf717ab6235d1a4a0346e467/core/src/index.ts#L47

jonathanadler: If i'm wrong, and you'ld like to prove it, you can add an `e2e` test file, testing this new method.
If it works we're all good

NitzanSirota: It may cause problems with IE 11 e2e tests because it is not supported (and probably not polyfilled) in IE11. I'll try using _prop method

PR #312, opened 2020-03-30
few size optimizations

zhukovka:
I am working on reducing our bundle size and have noticed that in `wix-ui-tpa` there are few things that could be optimized. Namely `wix-ui-icons-common` and `lodash`
my suggestions work for `wix-ui-icons-common` but, sadly not for `lodash` (maybe add it to externals?)

Before 

Screen Shot 2020-03-30 at 10 32 55 PM

After 

Screen Shot 2020-03-30 at 10 38 05 PM

### DIFF
--- a/src/components/DotNavigation/DotNavigation.tsx
+++ b/src/components/DotNavigation/DotNavigation.tsx
@@ -3,8 +3,10 @@ import styles from './DotNavigation.st.css';
 import { RadioButton, RadioButtonKeyDownEvent } from 'wix-ui-core/radio-button';
 import { DotNavigationDataKeys, DotNavigationDataHooks } from './dataHooks';
 import classNames from 'classnames';
-import isNaN from 'lodash/isNaN';
-import isNumber from 'lodash/isNumber';
+
+const isNan = require('lodash/isNan');
+const isNumber = require('lodash/isNumber');
+
 import { TPAComponentProps } from '../../types';
 
 const MAX_SHORT_LIST_LENGTH = 5;
--- a/src/components/DotNavigation/DotNavigation.tsx
+++ b/src/components/DotNavigation/DotNavigation.tsx
@@ -4,7 +4,7 @@ import { RadioButton, RadioButtonKeyDownEvent } from 'wix-ui-core/radio-button';
 import { DotNavigationDataKeys, DotNavigationDataHooks } from './dataHooks';
 import classNames from 'classnames';
 
-const isNan = require('lodash/isNan');
+const isNaN = require('lodash/isNaN');
 const isNumber = require('lodash/isNumber');
 
 import { TPAComponentProps } from '../../types';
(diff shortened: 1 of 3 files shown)


tsconfig.json
```diff
@@ -10,7 +10,8 @@
     "baseUrl": ".",
     "paths": {
       "protractor": ["test/protractor.d.ts"]
-    }
+    },
+    "esModuleInterop": true
```
jonathanadler: Are we sure we want to add this in this?
I'm not a big typescript expert, but this might have a big effect on the lib, no?
I think you can import lodash with `*` instead, just for now.
i.e.:
```js
import * as isNumber from 'lodash/isNumber';
```

zhukovka:
Thank you so much

tsconfig.json
jonathanadler: @zhukovka 
Did you see my comment here?

zhukovka: sorry, I missed it 
my TS doesn't accept 
` import * as isNumber from 'lodash/isNumber'; `

src/components/DotNavigation/DotNavigation.tsx
```diff
@@ -3,7 +3,8 @@ import styles from './DotNavigation.st.css';
 import { RadioButton, RadioButtonKeyDownEvent } from 'wix-ui-core/radio-button';
 import { DotNavigationDataKeys, DotNavigationDataHooks } from './dataHooks';
 import classNames from 'classnames';
-import * as _ from 'lodash';
+import isNaN from 'lodash/isNaN';
+import isNumber from 'lodash/isNumber';
```
jonathanadler, line 7: @zhukovka 
Even though your solution is probably the correct one, i'd rather we do the `tsconfig` change in a different, and controlled PR.
So for now, maybe instead write:
```suggestion

const isNan = require('lodash/isNan');
const isNumber = require('lodash/isNumber');

```

zhukovka: ok

tsconfig.json
jonathanadler, line 14: @zhukovka 
But also don't forget to remove this 😁 
Thanks 🙏

zhukovka: done

src/components/DotNavigation/DotNavigation.tsx
```diff
@@ -3,7 +3,10 @@ import styles from './DotNavigation.st.css';
 import { RadioButton, RadioButtonKeyDownEvent } from 'wix-ui-core/radio-button';
 import { DotNavigationDataKeys, DotNavigationDataHooks } from './dataHooks';
 import classNames from 'classnames';
-import * as _ from 'lodash';
+
+const isNan = require('lodash/isNan');
```
jonathanadler, line 7: Note that the build is breaking.
Need to change this:
```suggestion
const isNaN = require('lodash/isNaN');
```

PR #283, opened 2020-02-16
Dropdown state change fix

jonathanadler:

PR #313, opened 2020-04-01
Add <FloatingDropdown />

kranux:


src/components/FloatingDropdown/FloatingDropdownBase.st.css
```diff
@@ -0,0 +1,83 @@
...
+    outline: none;
+}
+
+.arrowIcon {
+    color: value(inputColor);
+}
+
+.text {
+    -st-mixin: TPAText(
+        MainTextFont "{theme: 'Body-M', size: '16px'}"
```
kranux, line 35: Hi, @jonathanadler 

Docs https://zeroheight.com/7sjjzhgo2/p/05a744-05-selections/b/46537f/t/36cac4

Says that I should use `Paragraph 2`:
image

I noticed that <Dropdown> has the same wiring:
image


And uses `-st-mixin: TPAText(MainTextFont "{theme: 'Body-M'`

Anyway, where is mapping between `Paragraph 2` and `Body-M` happening?

jonathanadler: That's a great question, and I hope we'll be able to make this more clear in the near future.
Our designers use `Paragraph 2` because that's the name that appears in the editor.
But the font variable is called `Body-M`, which quite confusing 😖 
The conversion is done by this tool we use:
https://github.com/wix/wix-style-processor
(but TPA's also use this tool instead: https://github.com/wix-incubator/tpa-style-webpack-plugin, which is pretty much the same, but works as a webpack plugin, and comes with yoshi apps).

As I said, I hope to have some Stylable constants in the future, so that contributors will only need to write:
```css
:import {
    -st-from: "../../common/typography.st";
    -st-named: paragraph_2;
}

.selector {
  font: value(paragraph_2);
}
```

Or a mixin, or something like that.

I hope this answered your question 🤔 .

kranux: Yes, thanks :)

src/components/FloatingDropdown/FloatingDropdown.st.css
```diff
@@ -0,0 +1,41 @@
...
+}
+
+:import {
+    -st-from: "./FloatingDropdownBase.st.css";
+    -st-default: FloatingDropdownBase;
+}
+
+:import {
+    -st-from: "../Dropdown/Dropdown.st.css";
+    -st-named: dropdown;
```
kranux, line 13: @jonathanadler I just re-use all the stylings from local <Dropdown/> component, is that OK?

src/components/FloatingDropdown/FloatingDropdown.tsx
```diff
@@ -0,0 +1,108 @@
+import * as React from 'react';
+
+import { Dropdown as CoreDropdown } from 'wix-ui-core/dropdown';
+
+import { TPAComponentsConsumer } from '../TPAComponentsConfig';
+import styles from './FloatingDropdown.st.css';
+import { DropdownOption } from '../Dropdown/DropdownOption';
```
kranux, line 7: @jonathanadler same with Options styling and props, use what we already have locally, OK?

jonathanadler:
Hey @kranux !
Note that you have some typescript errors (you can see in the build)
So, if you could just fix them, it would be easier for me to go over the PR.
Thanks!

p.s.
Ping me when you're ready

src/components/FloatingDropdown/FloatingDropdown.tsx
```diff
@@ -0,0 +1,133 @@
...
+import { TPAComponentsConsumer } from '../TPAComponentsConfig';
+import styles from './FloatingDropdown.st.css';
+import { FloatingDropdownBase } from './FloatingDropdownBase';
+import { DATA_HOOKS } from './constants';
+import { FloatingDropdownOptionProps } from './FloatingDropdownOption';
+
+export interface FloatingDropdownProps {
+  /** Defines a string value that labels the current element. Optional. */
+  'aria-label'?: string;
```
kranux, line 14: @jonathanadler I was not 100% sure about ARIA labels, so just copied ones used in `<Dropdown />`

src/components/FloatingDropdown/constants.ts
```diff
@@ -0,0 +1,6 @@
+export const DATA_HOOKS = {
+  coreDropdown: 'core-dropdown',
+  base: 'floating-dropdown-base',
+};
+
+export const ICON_SIZE = 24;
```
kranux, line 6: I would like to re-use it inside my st.css: https://github.com/wix/wix-ui-tpa/pull/313/files#diff-4c0e7e82a5365fefb909bd5e4b709f5eR23 - can I?

jonathanadler: Theoretically it might be able to import it (like we import the `formatters.st.js` file.
But this might cause a performance hit.
I guess for now it would be best to just have duplication :( 
Btw, from the brief review I gave, the component looks great!
I would consider using the `TextButton` instead of the regular `Button` 🤔

kranux: Yes, you are right, I should use TextButton here :)

Will add it :)

Thanks! :)

kranux:
TODO: use TextButton instead of Button;

src/components/FloatingDropdown/FloatingDropdown.tsx
jonathanadler: 👍

src/components/FloatingDropdown/constants.ts
kranux: Hi, changed Base implementation to use TextButton: https://github.com/wix/wix-ui-tpa/pull/313/commits/4c7f002016dba5418f3c1dab9d34532787fa86cc

src/components/FloatingDropdown/FloatingDropdown.e2e.ts
```diff
@@ -0,0 +1,26 @@
...
+  const storyUrl = createStoryUrl({
+    kind: 'Components',
+    story: 'FloatingDropdown',
+    withExamples: true,
+  });
+  const dataHook = 'storybook-FloatingDropdown';
+
+  beforeEach(() => browser.get(storyUrl));
+
+  it('should render', async () => {
```
jonathanadler, line 18: If we don't need any other tests in this file, then we can remove the file altogether.
Our e2e tests are used when we need a DOM for testing certain stuff.
In this case, we might need it to test that the dropdown opens as expected, etc.
However, we might also achieve this with visual tests.

src/components/FloatingDropdown/FloatingDropdown.spec.tsx
```diff
@@ -0,0 +1,52 @@
...
+          <FloatingDropdown {...testProps} />,
+          floatingDropdownTestkitFactory,
+          {
+            dataHookPropName: 'data-hook',
+          },
+        ),
+      ).toBe(true);
+    });
+  });
+
```
jonathanadler, line 37: Do we have any logic we need to test here 🤔 ?

src/components/FloatingDropdown/FloatingDropdown.tsx
```diff
@@ -0,0 +1,133 @@
...
+  /** Force dropdown open. Use for visual test. Optional. Boolean. */
+  forceContentElementVisibility?: boolean;
+}
+
+type DefaultProps = Required<
+  Pick<FloatingDropdownProps, 'disabled' | 'options'>
+>;
+
+interface State {
+  selectedOption: FloatingDropdownOptionProps;
```
jonathanadler, line 38: This ones my fault:
I have a PR I haven't merged yet
It removes the state from the dropdown.
Maybe you can do it here as well.
You can see my pr here:
https://github.com/wix/wix-ui-tpa/pull/283/files

It's because we had a bug that was caused because of this internal state

src/components/FloatingDropdown/FloatingDropdown.spec.tsx
kranux: I think, it makes sense to add some tests for:

1. checking if dropdown appears after a click on enabled FloatingDropdown component 
2. does not appear for disabled
3. making sore, that selected options changes in dropdown base after selecting it.

I will add these.

kranux: @jonathanadler added these: https://github.com/wix/wix-ui-tpa/pull/313/commits/a40eef63ae3d9fca60631f5dced87ae71be3857f

src/components/FloatingDropdown/FloatingDropdown.tsx
kranux: @jonathanadler I made component stateless. 
After this change, documentation might be a bit confusing, as it does not reflect real world example of controlled component.

jonathanadler: You mean in the storybook?

kranux: Yes, in storybook, https://wix-wix-ui-tpa-pr-313.surge.sh/

src/components/FloatingDropdown/FloatingDropdown.driver.ts
```diff
@@ -0,0 +1,57 @@
...
+  getBaseSelectedValue(): Promise<string>;
+}
+
+export const floatingDropdownDriverFactory = (
+  base: UniDriver,
+): FloatingDropdownDriver => {
+  const baseUniDriver = baseUniDriverFactory(base);
+
+  const getDropdownBase = async () => {
+    return base.$$(selectorFromDataHook(DATA_HOOKS.base)).get(0);
```
jonathanadler, line 25: Just a remark:
I think you can use just `$` instead of `$$`, then you won't need the `get(0)`.
But it's really really ok like this as well

kranux: Changed, thanks :)

kranux:
@jonathanadler thanks and Happy holidays! :)

Note that we also have Holidays here, so be back only on Tuesday next week :)

jonathanadler:
> @jonathanadler thanks and Happy holidays! :)
> 
> Note that we also have Holidays here, so be back only on Tuesday next week :)

Oh so happy holidays as well!

kranux:
Thanks! :)

michalvardi:
Component review:
- Arrow is not aligned to the text
- Whats the font size linked to?
- How can i see disabled state?

kranux:
@michalvardi 
- I'm checking Arrow alignment and will update you.
- Font size is linked to Paragraph 2 and size 16px
- Disabled scenario can be seen as the last example on this page: https://wix-wix-ui-tpa-pr-313.surge.sh/?path=/story/components--floatingdropdown

src/components/FloatingDropdown/FloatingDropdownBase.st.css
```diff
@@ -0,0 +1,91 @@
...
+}
+
+.arrowIcon path {
+    fill: value(textColor);
+}
+
+.label {
+    display: flex;
+    opacity: 0.6;
+    white-space: nowrap;
```
jonathanadler, line 61: Are we sure about this?
If so, we might need an ellipsis, since we have a max-width, no?

kranux: We do not want to wrap contend. Not sure about ellipsis. Should we?

Adding: https://github.com/wix/wix-ui-tpa/pull/313/commits/f55adeccf2b747e6e920aaa67369e2d64ea1c1f5

jonathanadler: From what I see in the spec, we should have the content wrapped:
image
So I guess it would be best to revert these last commits

jonathanadler: Wait, I made a mistake, I was talking about the option.
And your commit was about the label

kranux: Note that I modified DropdownBase not dropdown popup.

jonathanadler: We should check with @michalvardi to see what she says about this

kranux: image

kranux: I was making sure this scenario works fine ^^

jonathanadler: Yeah, I realised it late 😁
But we should still check with @michalvardi .
I'm guessing we'll go with ellipsis, but we need to be sure.

kranux: Yes, I asked her :)

jonathanadler:
Thanks @kranux 
great work!

PR #315, opened 2020-04-16
<Counter/> - add hover and focus behavior

bamai:
@jonathanadler

### DIFF
--- a/src/components/Counter/Counter.st.css
+++ b/src/components/Counter/Counter.st.css
@@ -1,6 +1,6 @@
 :import {
     -st-from: "../../common/formatters.st";
-    -st-named: applyUnit, color, fallback, font, value;
+    -st-named: applyUnit, color, fallback, font, value, applyOpacity;
 }
 
 :import {
@@ -43,11 +43,16 @@
     background: color(value(DefaultBackgroundColor));
     border-width: 1px;
     border-style: solid;
-    border-color: color(value(counterColor));
+    border-color: applyOpacity(value(counterColor), 0.6);
 }
 
 .root:error {
     border-color: value(DefaultErrorColor);
+}
+
+.root:hover:not(:disabled),
+.root:focus:not(:disabled){
+    border-color: value(counterColor);
 }
 
 .root:disabled {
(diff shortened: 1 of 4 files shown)


jonathanadler:
Looks good!
Did you talk to Johanna about the opacity with the user's custom color?
Also, I think we also need to add the hover styling to the buttons.
Only instead of changing the style of the button, we should probably change the implementation to use `IconButton` instead.
@johannaasseraf wdyt?

src/components/Counter/Counter.st.css
```diff
@@ -43,34 +49,47 @@
...
     padding: 6px;
     border: none;
     cursor: pointer;
 }
 
-.btn path {
+.root .btn::icon svg path {
     fill: color(value(counterColor));
+    stroke: none;
 }
```
jonathanadler, line 79: Note that you should also be able write it like this:
```suggestion
.btn {
    -st-extends: TPAIconButton;
    -st-mixin: TPAIconButton(
        IconColor color(value(counterColor))
    );
    background: none;
    padding: 6px;
    border: none;
    cursor: pointer;
}
```
similar to what I did in the Tabs component:
https://github.com/wix/wix-ui-tpa/blob/073e8d562ad2f4cc13426ac30d58d1e17188001b/src/components/Tabs/Tabs.st.css#L102

src/components/Counter/Counter.st.css
```diff
@@ -43,34 +49,47 @@
...
-.btn[disabled] path {
+.root:disabled .btn::icon svg path {
     fill: color(value(counterDisabledColor));
+    stroke: none;
+}
+
+.root .btn[disabled]::icon svg path {
+    fill: color(value(counterDisabledColor)) !important;
+    stroke: none;
 }
```
jonathanadler, line 93: Similar to the previous comment, you can do:
```suggestion
.btn:disabled {
    -st-mixin: TPAIconButton(
        IconColor color(value(counterDisabledColor))
    );
}
```
But this means the Counter's disabled color is different than the `IconButton`'s disabled color?
Maybe we should ask Johanna or Michal about this.
It seems to me they should be the same color.

PR #317, opened 2020-04-21
Native floating dropdown dropdown

jonathanadler:


jonathanadler:
@kranux 
I've added the native element for mobile.
I made some small changes to adjust to it.

I think that on the way I might have fixed the color wiring issue 😄

PR #321, opened 2020-04-26
Radio button component

ItayTur:
tests needs to be updated


jonathanadler:
Sorry, I just merged another PR, so you'll have to `git pull --rebase origin master` and resolve the conflicts 🙏

src/components/RadioButton/README.md
```diff
@@ -0,0 +1,4 @@
+## RadioButton
+An implementation of an RadioButton for TPAs
+This Readme was auto-generated and needs to be filled manually
+
```
jonathanadler, line 4: You can remove this file

src/components/RadioButton/RadioButton.driver.ts
```diff
@@ -0,0 +1,25 @@
+import {
+  BaseUniDriver,
+  baseUniDriverFactory,
+} from 'wix-ui-test-utils/base-driver';
+import { UniDriver } from 'wix-ui-test-utils/unidriver';
+import { RADIOBUTTON_DATA_KEYS } from './datahooks';
+
+export interface RadioButtonDriver extends BaseUniDriver {
+  hasChecked(): Promise<boolean>;
+  hasDisabled(): Promise<boolean>;
```
jonathanadler, line 10: Lets change these to 
```suggestion
  isChecked(): Promise<boolean>;
  isDisabled(): Promise<boolean>;
```

src/components/RadioButton/RadioButton.e2e.ts
```diff
@@ -0,0 +1,30 @@
+import { browser } from 'protractor';
```
jonathanadler, line 1: This file can be removed if there aren't any other tests, other than the default one.
The e2e file is part of the generator, because sometimes we need a real DOM for our tests.
But if it's not needed, we can just remove.

src/components/RadioButton/RadioButton.spec.tsx
```diff
@@ -0,0 +1,66 @@
...
+  //         <RadioButton {...defProps}/>,
+  //         enzymeRadioButtonTestkitFactory,
+  //         mount,
+  //         {
+  //           dataHookPropName: 'data-hook',
+  //         },
+  //       ),
+  //     ).toBe(true);
+  //   });
+  // });
```
jonathanadler, line 65: Why is this commented out?

src/components/RadioButton/RadioButton.spec.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+    expect(await driver.hasChecked()).toBeTruthy();
+  });
+
+  it('should show disabled state', async () => {
+    const driver = createDriver(<RadioButton disabled {...defProps} />);
+
+    expect(await driver.hasDisabled()).toBeTruthy();
+  });
+
```
jonathanadler, line 37: We should probably add a test for the onChange handler.
Also, we should probably test that a disabled component does not call the onChange handler.

src/components/RadioButton/RadioButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+    const {
+      checked,
+      disabled,
+      value,
+      label,
+      name,
+      onChange,
+      ...rest
+    } = this.props;
+    const radioBtnIcon = <span className={styles.checkmark} />;
```
jonathanadler, line 58: I would move this outside of the render method, so it won't be created on each render.
I think it can even be outside the element, at the top

src/components/RadioButton/RadioButton.visual.tsx
```diff
@@ -0,0 +1,34 @@
...
+  },
+];
+
+tests.forEach(({ describe, its }) => {
+  its.forEach(({ it, props }) => {
+    storiesOf(`RadioButton/${describe}`, module).add(it, () => (
+      <RadioButtonVisual {...props} />
+    ));
+  });
+});
```
jonathanadler, line 34: I know it wasn't part of the generator, but can you please use `storybook-snapper` instead of `storiesOf`.
You can take a look at the `ShareButton` for instance as an example: https://github.com/wix/wix-ui-tpa/blob/ea1685638e03e0d9a401285453c59106b1dfb8c1/src/components/ShareButton/ShareButton.visual.tsx

Also, we should probably have additional test cases, like `disabled` and `checked`

I really need to add this to the generator...
Sorry for the hassle

src/components/RadioButton/docs/RadioButtonWiringExample.tsx
```diff
@@ -0,0 +1,41 @@
...
+class RadioButtonWiringExample extends React.Component<{}, IState> {
+  state = {
+    label: 'Click for laugh',
+    isChecked: false,
+  };
+  handleChange = event => {
+    this.setState(prevState => {
+      const label = prevState.isChecked
+        ? 'Click for laughs'
+        : 'My wife is really mad at the fact that I have no sense of direction. So I packed up my stuff and right.';
```
jonathanadler, line 19: 😄

src/components/RadioButton/RadioButton.meta.tsx
```diff
@@ -0,0 +1,14 @@
...
+RadioButtonMetadata.nonReactStrictModeCompliant = true;
+
+RadioButtonMetadata.addSim({
+  title: 'render',
+  props: {
+    label: 'label',
+    value: 'value',
+    onChange: () => {},
+  },
+});
```
jonathanadler, line 14: Let's add additional simulations, like `disabled` and `checked`.
You can pretty much copy paste and just add the necessary props

src/components/RadioButton/RadioButton.spec.tsx
ItayTur: Good point!

src/components/RadioButton/RadioButton.spec.tsx
ItayTur: im not sure what is this test checking

jonathanadler: So when the project is built, each component has testkits generated automatically.
These testkits expose the drivers we create, to our consumers, so that they can use them in their tests.
This test just checks that the testkit exists.

src/components/RadioButton/RadioButton.meta.tsx
ItayTur: for this one i would also need explanation. didn't understand what it does.

jonathanadler: Sure.
This is a test file that tests a11y and ssr compliancy.
We use a tool called ui-autotools: https://github.com/wix-incubator/ui-autotools

`addSim` is just a test suite actually.
each simulation is a different configuration of the component to test

src/components/RadioButton/RadioButton.visual.tsx
ItayTur: when you have the time to explain all tests it would be great.

src/components/RadioButton/RadioButton.spec.tsx
ItayTur: I tried to do this test like checkbox component. but couldn't. the test fail. could you take a look ?

PR #323, opened 2020-04-26
#EE-23606|dropdown| make down arrow icon align with font size

varzager:
in dropdown => make down arrow icon align with font size of text

https://jira.wixpress.com/browse/EE-23606?page=com.wixpress.vi.jira-git-commits%3Acommits-plugin


jonathanadler:
The code looks pretty simple and good.
However, the visual tests are breaking because this change somehow shrank the component's height.
Can you try and see how to fix this?
Thanks!

PR #324, opened 2020-04-28
Arrow icon css size

jonathanadler:
@varzager 
Weird applitools tests failing
So I closed and opened the pr again

I don't know if it's related, but a little voodoo can't do any harm

### DIFF
--- a/src/components/Dropdown/Dropdown.st.css
+++ b/src/components/Dropdown/Dropdown.st.css
@@ -96,6 +96,9 @@
 
 .dropdown:content-visible .dropdownBase::arrowIcon {
     transform: rotate(180deg);
+}
+
+.dropdown .dropdownBase::arrowIcon {
     height: 1.5em;
     width: 1.5em;
 }
--- a/src/components/Dropdown/Dropdown.st.css
+++ b/src/components/Dropdown/Dropdown.st.css
@@ -99,6 +99,10 @@
 }
 
 .dropdown .dropdownBase::arrowIcon {
+    position: absolute;
+    right: 8px;
+    top: 50%;
+    transform: translateY(-50%);
     height: 1.5em;
     width: 1.5em;
 }
--- a/src/components/Dropdown/Dropdown.st.css
+++ b/src/components/Dropdown/Dropdown.st.css
@@ -95,7 +95,7 @@
 }
 
 .dropdown:content-visible .dropdownBase::arrowIcon {
-    transform: rotate(180deg);
+    transform: translateY(-50%) rotate(180deg);
 }
 
 .dropdown .dropdownBase::arrowIcon {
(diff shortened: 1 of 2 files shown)


jonathanadler:
@varzager unfortunately the visual test is still breaking 😞

jonathanadler:
Oh, I just noticed that only the `mobile` tests are breaking.
Since the mobile dropdown is structured a bit differently, I guess there needs to be a special treatment there

varzager:
@jonathanadler so i decided just to add a constant element(::after) with `height:24px`(as was originally constant for arrow icon).
this will prevent any breakage of ui for anyone who changed font-size.
other options would fix for plain case, but wouldn't cover others.

=> if someone wants to get a smaller dropdown he needs to change `height` to be `height:auto/initial`. so it will become dynamic

_examples for breakage cases:_
**desktop -** consumers expect to get same height of dropdown for smaller font - but icon getting smaller will make dropdown smaller
mobile - originally icon was bigger than text and made dropdown bigger. changing padding from 8px to 10px would fix for plain case, but would affect all other case of font-size changes to bigger or smaller

src/components/Dropdown/Dropdown.st.css
```diff
@@ -95,7 +95,16 @@
 }
 
 .dropdown:content-visible .dropdownBase::arrowIcon {
-    transform: rotate(180deg);
+    transform: translateY(-50%) rotate(180deg);
+}
+
+.dropdown .dropdownBase::arrowIcon {
```
jonathanadler, line 101: Looks great, awesome work!

Note that this won't be applied to the mobile dropdown (which is rendered with a `select` element).
To fix this, I think you need to just add the correct selector, like this:
```suggestion
.dropdown .dropdownBase::arrowIcon, 
.dropdown .dropdownNativeSelect::arrowIcon {
```
(see here: https://github.com/wix/wix-ui-tpa/pull/324/files#diff-c6f15d870b6a55bc847324c839142cceR70)

jonathanadler:
Unfortunately, the mobile icon size still isn't changing according to the font-size.

@varzager Do you want to try and fix this?
I can also merge for now, and we'll deal with mobile later

jonathanadler:
@varzager 
btw, how do you override the font-size in you app?
The Dropdown doesn't have a `font-size` variable defined in its `st.css`

jonathanadler:
@NitzanSirota 
Since you developed the native Dropdown, maybe you can help us figure out how to adjust the icon's size according to font-size on mobile?

NitzanSirota:
Hi, I think I know what is the problem - I messed up the stylable hierarchy in the native dropdown.
I made the select element itself as root, and the arrowIcon is its sibling and not his descendant which ruins the stylable override (.dropdownNativeSelect::arrowIcon).

PR #325, opened 2020-04-30
fix: Dropdown - expose dropdownOption in stylable

NitzanSirota:
Made it possible to override dropdownOption classes through stylable.

### DIFF
--- a/src/components/Dropdown/Dropdown.st.css
+++ b/src/components/Dropdown/Dropdown.st.css
@@ -67,6 +67,10 @@
     -st-extends: DropdownBase;
 }
 
+.dropdownOption {
+    -st-extends: DropdownOption;
+}
+
 .dropdownNativeSelect {
     -st-extends: DropdownNativeSelect;
 }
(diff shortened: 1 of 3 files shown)


jonathanadler:
@NitzanSirota 
This fix seems as it should work,
but
1. Maybe we should add a test/example to see that it is actually working
2. It's actually kind of a "hack", since we just expose the class. A more "correct" solution, would be to expose the variables we want to allow the users to override.

NitzanSirota:
Thanks for the response. Regarding your questions: 
1. I don't think this should have a special example because it is exactly like exposing the rest of the building components such as: DropdownError, DropdownNativeSelect and the core Dropdown itself. It just allows the consumers to override settings in case the default TPA wiring is not good for their situation.
2. You are right, but this will require more research about which variables we want to expose and where to wire them inside the component. As I said in answer 1, I think this PR is more for equality between the inner components, which will allow consumers to "hack" their way. The next level is making a more clear api through stylable variables.

PR #330, opened 2020-05-11
Stylable v3 migration

NitzanSirota:


src/components/Button/README.md
```diff
@@ -48,7 +48,11 @@ In order to override the theme, use Wix Stylable extends capabilities and wrap w
...
     ``` javascript
     import Button from 'wix-ui-tpa/Button';
-    import buttonStylesExt from './ButtonExt.st.css';
+    import { st, classes } from './ButtonExt.st.css';
 
-    const ButtonExt = (props) => <Button {...props} {...buttonStylesExt('root', {}, props)}/>;
+    const ButtonExt = (props) => 
+        (<Button 
+            className={st(classes.root)} 
+            {...props}
```
argshook, line 56:
```suggestion
            {...props} 
            className={st(classes.root, props.className)} 
```

src/components/Card/README.md
```diff
@@ -44,7 +44,7 @@ In order to override the theme, use Wix Stylable extends capabilities and wrap w
 
 2. Create a component that uses it
     ``` javascript
-    import cardStylesExt from './CardExt.st.css';
+    import { st, classes } from './CardExt.st.css';
 
-    const CardExt = (props) => <Card upgrade {...props} {...cardStylesExt('root', {}, props)}/>;
+    const CardExt = (props) => <Card upgrade className={st(classes.root)} {...props}/>;
```
argshook, line 49: same comment as before, `{...props}` should go before `className`

src/components/Checkbox/Checkbox.spec.tsx
```diff
@@ -35,29 +35,29 @@ describe('Checkbox', () => {
...
-
-    expect(await driver.hasIndeterminate()).toBeTruthy();
-  });
-
-  it('should show checked state', async () => {
-    const driver = createDriver(<Checkbox checked {...defProps} />);
-
-    expect(await driver.hasChecked()).toBeTruthy();
-  });
+  // it('should show error state', async () => {
```
argshook, line 38: what's the reason to comment these tests?

if it's the only way, i'd do `it.skip`, then it would show up in stats

src/components/Checkbox/README.md
```diff
@@ -31,13 +31,13 @@ In order to override the theme, use Wix Stylable extends capabilities and wrap w
...
     import Checkbox from 'wix-ui-tpa/Checkbox';
-    import checkboxStylesExt from './CheckboxExt.st.css';
+    import { st, classes } from './CheckboxExt.st.css';
 
     const CheckboxExt = (props) => (
         <Checkbox
             onChange={() => {}}
             label="Label"
+            className={st(classes.root)} 
             {...props}
```
argshook, line 41:
```suggestion
            {...props}
            className={st(classes.root, props.className)}
```

src/components/CheckboxGroup/CheckboxGroup.spec.tsx
```diff
@@ -28,17 +28,17 @@ describe('CheckboxGroup', () => {
...
-
-    expect(await driver.isCheckboxesDisabled()).toBe(true);
-  });
-
-  it('should show error state at all checkboxes with error prop', async () => {
-    const driver = createDriver(<TestComp error />);
-
-    expect(await driver.isCheckboxesErrored()).toBe(true);
-  });
+  // it('should disable all checkboxes with disabled prop', async () => {
```
argshook, line 31: same comment as before: if there's no other way but to skip tests, then do `it.skip`

src/components/CheckboxGroup/README.md
```diff
@@ -28,13 +28,13 @@ In order to override the theme, use Wix Stylable extends capabilities and wrap w
...
     import CheckboxGroup from 'wix-ui-tpa/CheckboxGroup';
     import Checkbox from 'wix-ui-tpa/Checkbox';
-    import checkboxGroupStylesExt from './CheckboxGroupExt.st.css';
+    import { st, classes } from './CheckboxGroupExt.st.css';
 
     const CheckboxExt = (props) => (
         <CheckboxGroup
             label="Label"
+            className={st(classes.root)} 
             {...props}
```
argshook, line 37:
```suggestion
            {...props}
            className={st(classes.root, props.className)}
```

src/components/ColorPicker/ColorPickerItem/ColorPickerItem.tsx
```diff
@@ -32,11 +32,7 @@ export class ColorPickerItem extends React.Component<
 
     return (
       <RadioButton
-        {...styles(
-          'root',
-          { checked, focused: this.state.focused },
-          { ...this.props },
-        )}
+        className={st(classes.root, { checked, focused: this.state.focused })}
```
argshook, line 35:
```suggestion
        className={st(classes.root, { checked, focused: this.state.focused }, this.props.className)}
```

`className` is defined in `ColorPickerItemProps`, so should be passed

src/components/Button/Button.st.css
```diff
@@ -26,11 +26,11 @@
 
 :vars {
     /*Overrides*/
-    MainTextColor: --overidable;
-    MainBackgroundColor: --overidable;
-    MainTextFont: --overidable;
-    MainBorderColor: --overidable;
-    MainBorderWidth: --overidable;
+    MainTextColor: --overridable;
```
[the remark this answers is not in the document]
NitzanSirota: Most of the components are written the way I fixed it to be, so I guess so.
This is for wix-style-processor

src/components/Button/README.md
NitzanSirota: This is just an example of using the component, so most consumers won't accept className as a prop, they would just use it like this

src/components/Card/README.md
NitzanSirota: again, an example of extending a component

src/components/Checkbox/Checkbox.spec.tsx
NitzanSirota: Those are tests I need to fix somehow.

The problem is that we send data-attributes on ui-tpa's `CheckBox` which renders `CoreCheckBox` - before the update it was ok because it was spread, but now we need to find another solution.
Those data-attributes represents stylable states for tests, but @jonathanadler said that they don't use the stylable testkit because it may fail on consumers E2E tests.

src/components/Checkbox/README.md
NitzanSirota: again, an example of extending a component

src/components/CheckboxGroup/CheckboxGroup.spec.tsx
NitzanSirota: same problem as above - I need to fix those before merging, just not sure how yet

src/components/CheckboxGroup/README.md
NitzanSirota: again, an example of extending a component

src/components/ColorPicker/ColorPickerItem/ColorPickerItem.tsx
NitzanSirota: good catch

src/components/Button/README.md
argshook: i understand this (and some others) are just examples, but it would be good to show proper usage in examples that can avoid potential missuses.

it's easy for us, cause we know how it works, but for someone new it can be not that trivial

src/components/Checkbox/Checkbox.spec.tsx
argshook: there's the same issue with wix-style-react, so i updated wix-ui-core to
support `data-*` attributes on all components.

please try reinstalling `wix-ui-core` and let me know if it doesn't work

NitzanSirota: https://github.com/wix/wix-ui-tpa/pull/330/commits/71f7ec23ea51d09eb77601d02b157b75dc2e2e12

src/components/SocialBar/SocialBarIcon.tsx
```diff
@@ -20,19 +20,27 @@ export class SocialBarIcon extends React.Component<SocialBarIconProps> {
   static defaultProps = { socialBarTheme: 'light' };
 
   render() {
-    const { children, tooltip, socialBarTheme, ...rest } = this.props;
+    const {
+      children,
+      tooltip,
+      socialBarTheme,
+      className,
+      ...rest
```
argshook, line 28:
```suggestion
```

looks unused

src/components/Text/Text.tsx
```diff
@@ -21,22 +21,23 @@ export class Text extends React.Component<TextProps> {
...
+              className: st(
+                classes.root,
                 {
                   typography,
                   mobile,
                 },
-                rest,
+                className,
               ),
+              ['data-hook']: this.props['data-hook'],
```
argshook, line 40:
```suggestion
              'data-hook': this.props['data-hook'],
```

src/components/SocialBar/SocialBarIcon.tsx
NitzanSirota: used in line 46 (was there before this PR)

src/components/CheckboxGroup/index.tsx
```diff
@@ -1 +1,5 @@
-export * from './CheckboxGroup';
+export {
+  CheckboxGroup,
+  CheckboxGroupProps,
+  Layout as CheckboxGroupLayout,
+} from './CheckboxGroup';
```
jonathanadler, line 5: This is actually causing us issues in production at the moment.
This is a breaking change, so we should remember notifying about this change.

src/components/DotNavigation/DotNavigation.tsx
```diff
@@ -220,7 +220,7 @@ export class DotNavigation extends React.Component<
       onChange={onChange}
       onKeyDown={onKeyDown}
       aria-label={ariaLabel}
-      {...{ className: styles.extendedRadioButton }}
+      {...{ className: classes.extendedRadioButton }}
```
jonathanadler, line 223: Not related to your work, but i'm really scratching my head on why this was written this way 🤔

src/components/FloatingDropdown/FloatingDropdown.tsx
```diff
@@ -28,18 +31,24 @@ export interface FloatingDropdownProps {
   value?: string;
   /** Force dropdown open. Use for visual test. Optional. Boolean. */
   forceContentElementVisibility?: boolean;
+  /** Placement for Popover. Optional. Placement. */
+  placement?: Placement;
```
jonathanadler, line 35: Why was this prop added?

package.json
```diff
@@ -63,19 +63,19 @@
...
+    "wix-ui-core": "^3.0.0",
+    "wix-ui-icons-common": "^2.0.0",
+    "yoshi-style-dependencies": "^4.6.0"
   },
   "devDependencies": {
     "@applitools/eyes-storybook": "^2.7.28",
     "@storybook/addon-links": "^5.0.9",
     "@storybook/react": "^5.0.9",
-    "@stylable/cli": "^1.0.0",
-    "@stylable/core": "^1.1.0",
```
jonathanadler, line 78: We're using `@stylable/core` here: https://github.com/wix/wix-ui-tpa/blob/87648ef57f67d662dc7ce9d8a0de5109245ee914/.storybook/stylable-metadata-plugin.js#L9
I'm wondering if it's ok to remove it.

But since I can see the storybook is generated and working good, I can only assume that it's ok 🤷‍♂️

.wuf/generator/templates/mobile/Component/Component.tsx
```diff
@@ -35,13 +35,13 @@ export class {%ComponentName%} extends React.Component<{%ComponentName%}Props, S
     return (
       <TPAComponentsConsumer>
         {({ mobile }) => (
-          <div {...styles('root', { mobile }, rest)} data-mobile={mobile}>
-            <Text {...styles('number', { even: isEven, odd: !isEven })}>
+          <div className={st(classes.root, { mobile }, rest)} data-mobile={mobile}>
+            <Text className={st(classes.number, { even: isEven, odd: !isEven })}>
```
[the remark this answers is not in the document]
NitzanSirota: I remember it can be used on non-root elements if it has a state that needs to be sent.
@barak007 What is the convention? on a non-root stateful element, should I use the style function with the states or should I use the cssStates function?

src/components/FloatingDropdown/FloatingDropdown.tsx
NitzanSirota: Because of typescript changes, we had to send a placement prop to the core dropdown. I thought it should be passed like the rest of the props and not to be permanent.

.wuf/generator/templates/mobile/Component/Component.tsx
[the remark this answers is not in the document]
jonathanadler: In any case, the template shouldn't have this form of stylable state.
It's not something we need for a new component, and it's not something we want to encourage.
But you can ignore this comment.
We need to do lots of cleaning in the generator template.
We'll do it in another PR

package.json
NitzanSirota: wix-ui-tpa@1.10.2 /Users/nitzansi/Projects/wix-ui-tpa
├─┬ @stylable/cli@3.5.2
│ ├── @stylable/core@3.5.2 
│ ├─┬ @stylable/node@3.5.2
│ │ ├── @stylable/core@3.5.2  deduped
│ │ └─┬ @stylable/module-utils@3.5.2
│ │   └── @stylable/core@3.5.2  deduped
│ └─┬ @stylable/optimizer@3.5.2
│   └── @stylable/core@3.5.2  deduped
├─┬ @stylable/dom-test-kit@3.5.2
│ └── @stylable/core@3.5.2  deduped
├─┬ @ui-autotools/a11y@2.7.3
│ └─┬ @ui-autotools/node-utils@1.5.3
│   └─┬ @stylable/node@1.4.0
│     └── @stylable/core@1.4.0 
├─┬ wix-ui-core@3.0.34
│ └─┬ yoshi-stylable-dependencies@4.78.0
│   ├─┬ @stylable/jest@3.5.0
│   │ └── @stylable/core@3.5.2  deduped
│   ├─┬ @stylable/module-utils@3.5.0
│   │ └── @stylable/core@3.5.2  deduped
│   └─┬ @stylable/node@3.5.1
│     └── @stylable/core@3.5.2  deduped
└─┬ yoshi@4.80.3
  └─┬ yoshi-common@4.80.3
    ├── @stylable/core@1.4.0 
    └─┬ @stylable/node@1.4.0
      └── @stylable/core@1.4.0  deduped

NitzanSirota: This is the npm ls tree for @stylable/core.
To be honest, I'm not sure if we need to install it explicitly because so many packages are relying on it. 
Maybe we should do it to prevent versions mess

src/components/FloatingDropdown/FloatingDropdown.tsx
jonathanadler: I see.
Too bad the wix-ui-core Dropdown doesn't have a default `auto` placement.
Should our default be `auto` instead of `bottom` 🤔  or is `bottom` the behaviour we're expecting?

package.json
jonathanadler: Yeah, I guess we can leave it, because it's working ok.

jonathanadler: I mean, we can leave the PR as it is now

jonathanadler:
@NitzanSirota 
I created a fix for the `CheckboxGroupLayout` issue in master.
So it's now creating a conflict here.

Since this is a breaking change, I think it would be best to change the `Layout`'s name like you did.

In addition, I think we don't need to rename it each time `CheckboxGroupLayout` => `Layout` and then `Layout` => `CheckboxGroupLayout` again.

I think we can just rename it,

wdyt?

NitzanSirota:
> @NitzanSirota
> I created a fix for the `CheckboxGroupLayout` issue in master.
> So it's now creating a conflict here.
> 
> Since this is a breaking change, I think it would be best to change the `Layout`'s name like you did.
> 
> In addition, I think we don't need to rename it each time `CheckboxGroupLayout` => `Layout` and then `Layout` => `CheckboxGroupLayout` again.
> 
> I think we can just rename it,
> 
> wdyt?

Let me see if I got you straight - You want to rename the exported interface to be renamed to`CheckboxGroupLayout`, and not exporting it like `export {Layout as CheckboxGroupLayout}`.
Right? 
If so, Then yes, I think it's the best solution :)

jonathanadler:
> > @NitzanSirota
> > I created a fix for the `CheckboxGroupLayout` issue in master.
> > So it's now creating a conflict here.
> > Since this is a breaking change, I think it would be best to change the `Layout`'s name like you did.
> > In addition, I think we don't need to rename it each time `CheckboxGroupLayout` => `Layout` and then `Layout` => `CheckboxGroupLayout` again.
> > I think we can just rename it,
> > wdyt?
> 
> Let me see if I got you straight - You want to the exported interface to be renamed to`CheckboxGroupLayout`, and not exporting it like `export {Layout as CheckboxGroupLayout}`.
> Right?
> If so, Then yes, I think it's the best solution :)

Exactly 😁

package.json
argshook: this reminds me, that `@ui-autotools` should probably also be upgraded

PR #335, opened 2020-05-20
Counter - a11y fixes

NitzanSirota:
https://jira.wixpress.com/browse/EE-24013

### DIFF
--- a/src/components/Counter/Counter.driver.ts
+++ b/src/components/Counter/Counter.driver.ts
@@ -8,11 +8,13 @@ import { tooltipDriverFactory } from 'wix-ui-core/dist/src/components/tooltip/To
 import { Simulate } from 'react-dom/test-utils';
 
 export interface CounterDriver extends BaseUniDriver {
+  changeInputFocus(shouldFocus: boolean): Promise<void>;
   enterValue(val: string): Promise<void>;
   getCounterAriaLabel(): Promise<string>;
   getCounterAriaLabellledby(): Promise<string>;
   getErrorMessageContent(): Promise<void>;
   hasCounterComponentError(): Promise<boolean>;
+  isInputHasAriaLive(): Promise<boolean>;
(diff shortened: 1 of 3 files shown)


NitzanSirota:
I think it's a behavior that should come built-in. This way, whenever you increase/decrease while focusing on the plus/minus the screen reader reads the new value.
I don't see any case that some consumers wouldn't want this behavior.

I add the aria-live property only when the input is not focused to prevent the SR to read it twice when you are focused on the input.

Maybe I should change the state from `isFocused` to `isInputFocused` - I think it caused some confuision.

jonathanadler:
> I think it's a behavior that should come built-in. This way, whenever you increase/decrease while focusing on the plus/minus the screen reader reads the new value.
> I don't see any case that some consumers wouldn't want this behavior.
> 
> I add the aria-live property only when the input is not focused to prevent the SR to read it twice when you are focused on the input.
> 
> Maybe I should change the state from `isFocused` to `isInputFocused` - I think it caused some confuision.

🤔 I'm not sure about what you're saying.
Usually, inputs should work out of the box, and no aria-live should be added.
aria-live is usually used on notifications, alerts and elements of that type.

That's why I really feel it's only your use case.

Did @NeilWix guide you about this issue?

NitzanSirota:
It's there for the case that you change the input's value when you are not focused on it(plus/minus) which is not your usual input behavior. Also, this behavior should be the same for everyone who uses this component.
And Alon the a11y QA guided me through this.
Examples:
https://codepen.io/WW3/pen/wJPwMB

NitzanSirota:
this change was triggered because of this JIRA: https://jira.wixpress.com/browse/EE-24013

jonathanadler:
@NitzanSirota 
I see...
However I don't really understand the codepen, it doesn't really seem to work.
But if Neil wrote it, so I guess it's legit 🤷‍♂️

PR #344, opened 2020-06-02
Origin picker

jonathanadler:


src/components/Picker/Picker.tsx
```diff
@@ -0,0 +1,65 @@
...
+  arrowsSize?: string;
+}
+
+interface DefaultProps {
+  arrowsSize: string;
+  'data-hook': string;
+}
+
+/** Picker */
+export class Picker extends React.Component<PickerProps> {
```
jonathanadler, line 22: I think `Picker` is a bit too general.
Let's call it `CalendarPicker`

src/components/Picker/Picker.tsx
```diff
@@ -0,0 +1,65 @@
...
+    } = this.props;
+
+    return (
+      <div
+        {...styles('root', {}, rest)}
+        data-hook={this.props['data-hook']}
+        {...this.getDataAttributes()}
+      >
+        <IconButton
+          {...styles('arrow', {})}
```
jonathanadler, line 52: This should be
```suggestion
          {styles.arrow}
```

src/components/Picker/Picker.tsx
```diff
@@ -0,0 +1,65 @@
...
+          icon={<ChevronLeft height={arrowsSize} width={arrowsSize} />}
+          as="a"
+        />
+        <Text className={styles.value}>{value}</Text>
+        <IconButton
+          {...styles('arrow', {})}
```
jonathanadler, line 58: This should be
```suggestion
          {styles.arrow}
```

src/components/Picker/Picker.visual.tsx
```diff
@@ -0,0 +1,34 @@
...
+  }
+}
+
+const tests = [
+  {
+    describe: 'basic',
+    its: [
+      {
+        it: 'default',
+        props: {},
```
jonathanadler, line 22: We should pass a `label` for this test

src/components/Picker/README.md
```diff
@@ -0,0 +1,4 @@
+## Picker
+An implementation of an Picker for TPAs
+This Readme was auto-generated and needs to be filled manually
+
```
jonathanadler, line 4: You can delete this file

src/components/Picker/Picker.e2e.ts
```diff
@@ -0,0 +1,27 @@
...
+  const dataHook = 'storybook-Picker';
+
+  beforeEach(() => browser.get(storyUrl));
+
+  it('should render', async () => {
+    const driver = pickerTestkitFactory({ dataHook });
+    await waitForVisibilityOf(await driver.element(), 'Cannot find Picker');
+    expect((await driver.element()).isDisplayed()).toBe(true);
+  });
+});
```
jonathanadler, line 27: I think this file isn't needed, so you can delete it

src/components/Picker/Picker.driver.ts
```diff
@@ -0,0 +1,19 @@
+import {
+  BaseUniDriver,
+  baseUniDriverFactory,
+} from 'wix-ui-test-utils/base-driver';
+import { PICKER_DATA_KEYS } from './dataHooks';
+import { UniDriver } from 'wix-ui-test-utils/unidriver';
+
+export interface PickerDriver extends BaseUniDriver {
+  isArrowsSizeSet(arrowsSized): Promise<boolean>;
```
jonathanadler, line 9: We need to add `clickNext` and `clickPrev` or something like that, and also test the callbacks in the spec file

src/components/Picker/docs/PickerWiringExample.tsx
```diff
@@ -0,0 +1,14 @@
...
+export const PickerWiringExample = () => {
+  return (
+    <Picker
+      value={'October 2020'}
+      previousClickHandler={() => {}}
+      nextClickHandler={() => {}}
+      {...styles('root', {})}
+    />
+  );
+};
```
jonathanadler, line 14: Let's have a bit more options to this example.
For instance add a state for the value, so that we'll have some options.
ie:
```suggestion
const months = ['January', 'February', ... , 'December'];
export const PickerWiringExample = () => {
  const [value, setValue] = React.useState(months.length / 2);
  return (
    <Picker
      value={'October 2020'}
      onPrev={() => setValue(Math.max(value - 1, 0))}
      onNext={() => setValue(Math.min(value + 1, months.length - 1))}
      disableNext={value === months.length - 1}
      disablePrev={value === 0}
      {...styles('root', {})}
    />
  );
};
```

src/components/Picker/Picker.tsx
```diff
@@ -0,0 +1,65 @@
+import * as React from 'react';
+import { Text } from '../Text';
+import { PICKER_DATA_HOOKS, PICKER_DATA_KEYS } from './dataHooks';
+import { ReactComponent as ChevronLeft } from '../../assets/icons/ChevronLeft.svg';
+import { ReactComponent as ChevronRight } from '../../assets/icons/ChevronRight.svg';
+import { IconButton } from '../IconButton';
+import styles from './Picker.st.css';
+
+export interface PickerProps {
```
jonathanadler, line 9: I think we'll also need to add:
```
disablePrev?: boolean;
disableNext?: boolean;
```

jonathanadler: or
```
disable?: 'prev' | 'next' | 'both';
```

src/components/Picker/Picker.tsx
```diff
@@ -0,0 +1,65 @@
...
+  value: any;
+  previousClickHandler(): void;
+  nextClickHandler(): void;
+  arrowsSize?: string;
```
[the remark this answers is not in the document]
ItayTur: that's absolutely right. The problem is that the SVG's height and width can be overriden by react props. I didn't find a way to do it by stylable.

src/components/Picker/Picker.tsx
ItayTur: The picker should be general. i don't see the behavior and style change for any use case.

src/components/Picker/Picker.visual.tsx
ItayTur: what is label is used for ? also, would you like it to be with storybook-snapper ?

src/components/Picker/Picker.tsx
ItayTur: good idea ! what do you think about calling them prevDisabled and nextDisabled ? because they are booleans so adjectives would fit better then verbs

src/components/Picker/Picker.driver.ts
ItayTur: I followed the TPA Dropdown click test. but I don't know if it's enough, let me know what you think

ItayTur:
@jonathanadler 
good notes! 
I followed them and left my own clarifications and questions.
About the array, index and onChange props. that will cause the component to be less dynamic. because for a component that don't use an array the picker won't work. the calendar is a good example: in this case we would want the picker to be infinite. So, the main component holds a date object and clicking the picker arrows adds or subtract 1 month from it. The picker value gets the month from the object through the props. 
let me know what you think about that and the rest of the updates.

src/components/Picker/Picker.tsx
```diff
@@ -0,0 +1,75 @@
...
+    return (
+      <div
+        data-hook={this.props['data-hook']}
+        {...styles('root', {}, rest)}
+        {...this.getDataAttributes()}
+      >
+        <IconButton
+          className={styles.arrow}
+          icon={<ChevronLeft />}
+          as="a"
```
jonathanadler, line 58: this shouldn't be used `as` an anchor.
It's not the correct usage here.
You can leave this without the `as` prop, for the default `button`

src/components/Picker/Picker.tsx
```diff
@@ -0,0 +1,75 @@
...
+          onClick={() => (prevDisabled ? null : onPrev())}
+          disabled={prevDisabled}
+          data-hook={PICKER_DATA_HOOKS.Prev}
+        />
+        <Text className={styles.value}>{value}</Text>
+        <IconButton
+          className={styles.arrow}
+          as="a"
```
jonathanadler, line 66: same here

jonathanadler:
We've noticed the picker grows with the text, which makes it jumpy.
So I've consulted with @michalvardi and we decided to have a width of `12.6em` and a min-width of `207px` for the value's element

PR #345, opened 2020-06-02
Calendar Event

jonathanadler:


src/components/DotNavigation/DotNavigation.tsx
```diff
@@ -311,10 +311,10 @@ export class DotNavigation extends React.Component<
...
     const { length, ...rest } = this.props;
-    let short = length <= MAX_SHORT_LIST_LENGTH;
+    const short = length <= MAX_SHORT_LIST_LENGTH;
 
     return _isNumber(length) && length > 0 ? (
-      <div {...styles('root', {short}, rest)} {...this._getDataAttributes()}>
+      <div {...styles('root', { short }, rest)} {...this._getDataAttributes()}>
         {length <= MAX_SHORT_LIST_LENGTH
           ? this.renderShortVersion()
           : this.renderLongVersion()}
```
jonathanadler, line 320: Note that you have this lint correction in all your PR's
So i'll add this fix to master, and then remove these changes and pull rebase

src/components/Event/Event.e2e.ts
```diff
@@ -0,0 +1,27 @@
...
+  const dataHook = 'storybook-Event';
+
+  beforeEach(() => browser.get(storyUrl));
+
+  it('should render', async () => {
+    const driver = eventTestkitFactory({ dataHook });
+    await waitForVisibilityOf(await driver.element(), 'Cannot find Event');
+    expect((await driver.element()).isDisplayed()).toBe(true);
+  });
+});
```
jonathanadler, line 27: This can be erased

src/components/Event/Event.st.css
```diff
@@ -0,0 +1,143 @@
...
+    CurrentAllDayBackground: color(fallback(value(AllDayBackground), value(DefaultAllDayBackground)));
+    CurrentAllDayOpacity: fallback(value(AllDayOpacity), value(DefaultAllDayOpacity));
+    CurrentTopPadding: fallback(value(TopPadding),value(DefaultTopPadding));
+    CurrentRightPadding: fallback(value(RightPadding),value(DefaultRightPadding));
+    CurrentLeftPadding: fallback(value(LeftPadding),value(DefaultLeftPadding));
+    CurrentBottomPadding: fallback(value(BottomPadding),value(DefaultBottomPadding));
+    CurrentRTLTopPadding: fallback(value(RTLTopPadding),value(DefaultRTLTopPadding));
+    CurrentRTLRightPadding: fallback(value(RTLRightPadding),value(DefaultRTLRightPadding));
+    CurrentRTLLeftPadding: fallback(value(RTLLeftPadding),value(DefaultRTLLeftPadding));
+    CurrentRTLBottomPadding: fallback(value(RTLBottomPadding),value(DefaultRTLBottomPadding));
```
jonathanadler, line 61: did you mean:
```suggestion
    DefaultFontStyle: fallback(value(FontStyle), font(value(DefaultFontStyle)));
    DefaultTimeColor: color(fallback(value(TimeColor), value(DefaultTimeColor)));
    DefaultTitleColor: color(fallback(value(TitleColor), value(DefaultTitleColor)));
    DefaultAllDayBackground: color(fallback(value(AllDayBackground), value(DefaultAllDayBackground)));
    DefaultAllDayOpacity: fallback(value(AllDayOpacity), value(DefaultAllDayOpacity));
    DefaultTopPadding: fallback(value(TopPadding),value(DefaultTopPadding));
    DefaultRightPadding: fallback(value(RightPadding),value(DefaultRightPadding));
    DefaultLeftPadding: fallback(value(LeftPadding),value(DefaultLeftPadding));
    DefaultBottomPadding: fallback(value(BottomPadding),value(DefaultBottomPadding));
    DefaultRTLTopPadding: fallback(value(RTLTopPadding),value(DefaultRTLTopPadding));
    DefaultRTLRightPadding: fallback(value(RTLRightPadding),value(DefaultRTLRightPadding));
    DefaultRTLLeftPadding: fallback(value(RTLLeftPadding),value(DefaultRTLLeftPadding));
    DefaultRTLBottomPadding: fallback(value(RTLBottomPadding),value(DefaultRTLBottomPadding));
```

src/components/Event/Event.st.css
```diff
@@ -0,0 +1,143 @@
...
+    AllDayBackground:  --overridable;
+    AllDayOpacity: --overridable;
+    TopPadding: --overridable;
+    RightPadding: --overridable;
+    BottomPadding: --overridable;
+    LeftPadding: --overridable;
+    RTLTopPadding: --overridable;
+    RTLRightPadding: --overridable;
+    RTLBottomPadding: --overridable;
+    RTLLeftPadding: --overridable;
```
jonathanadler, line 44: I'm not sure the paddings should be overridden
@michalvardi ?

jonathanadler: In any case, we also prefer css solutions to dealing with `rtl` ie `flex` and `::before` and `::after`

src/components/Event/Event.st.css
```diff
@@ -0,0 +1,143 @@
...
+.root {
+    -st-states: isMultiday, isSelected, isRightToLeft;
+    position: relative;
+    padding: value(CurrentTopPadding) value(CurrentRightPadding) value(CurrentBottomPadding) value(CurrentLeftPadding);
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
+}
+
+.root:isRightToLeft {
```
jonathanadler, line 73: If we really need to pass an `rtl` state from javascript, then we use the `TPAComponentsConsumer`.
You can see many examples for using this in the code.
But before we use this, we prefer having a `direction` agnostic approach.
Meaning, using flex, and other solutions (`::before`, `::after`).

src/components/Event/Event.st.css
```diff
@@ -0,0 +1,143 @@
...
+    position: relative;
+    direction: rtl;
+    display: table;
```
jonathanadler, line 76: I'm not sure using `table` is necessary here.
It seems to be a bit of an overkill imo

src/components/Event/Event.st.css
```diff
@@ -0,0 +1,143 @@
...
+    table-layout: fixed;
+    float: right;
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
```
jonathanadler, line 81: No need to add the ellipsis again (it's already defined on the `.root`)

src/components/Event/Event.st.css
```diff
@@ -0,0 +1,143 @@
...
+
+.root:isRightToLeft .time{
+    padding-right: 0;
+    padding-left: 2px;
+    display: table-cell;
+}
+
+.root:isMultiday {
+    background: applyOpacity(value(CurrentAllDayBackground), value(CurrentAllDayOpacity));
+}
```
jonathanadler, line 143: So in general, I think you can use `flex` instead of table
I think you'll get the desired outcome pretty much out of the box

src/components/Event/Event.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { EVENT_DATA_HOOKS, EVENT_DATA_KEYS } from './dataHooks';
+import styles from './Event.st.css';
+
+export interface EventProps {
+  time?: string;
+  title: string;
+  isTimeShown?: boolean;
+  isMultiday?: boolean;
+  isRightToLeft?: boolean;
+  isSelected?: boolean;
```
jonathanadler, line 12: we usually prefer using:
```suggestion
  showTime?: boolean;
  multiday?: boolean;
  selected?: boolean;
```
No need to have an `isRightToLeft` prop.
We use the `TPAComponentsConsumer`  for this.
You can see an example here: https://github.com/wix/wix-ui-tpa/blob/1f5ce19adcb485b4b8a43f044bfd1ec0b044ceda/src/components/Pagination/Pagination.tsx#L25

src/components/Event/Event.visual.tsx
```diff
@@ -0,0 +1,34 @@
...
+  {
+    describe: 'basic',
+    its: [
+      {
+        it: 'default',
+        props: {},
+      },
+    ],
+  },
+];
```
jonathanadler, line 26: Let's add more use cases to take screenshots of 
and preferably use `storybook-snapper` like you did with the cell component

src/components/Event/README.md
```diff
@@ -0,0 +1,4 @@
+## Event
+An implementation of an Event for TPAs
+This Readme was auto-generated and needs to be filled manually
+
```
jonathanadler, line 4: This can be removed

src/components/Event/docs/EventWiringExample.tsx
```diff
@@ -0,0 +1,23 @@
...
+      <div>
+        <h4>All Day</h4>
+        <Event
+          {...styles('root')}
+          isMultiday
+          time="20:30"
+          title="Blink 182 Concert"
+        />
+      </div>
+    </div>
```
jonathanadler, line 21: Even though it seems to be working, the correct stylable syntax would be:
```suggestion
    <div {...styles('root')}>
      <div>
        <h4>Regular</h4>
        <Event className={styles.event} time="20:30" title="Blink 182 Concert" />
      </div>
      <div>
        <h4>All Day</h4>
        <Event
          className={styles.event}
          isMultiday
          time="20:30"
          title="Blink 182 Concert"
        />
      </div>
    </div>
```

src/components/Event/docs/EventWiringExample.st.css
```diff
@@ -0,0 +1,19 @@
...
+            AllDayBackground '"--allDayBackground"',
+            FontStyle '"--fontStyle"',
+            AllDayOpacity '"unit(fallback(zeroAsTrue(--allDayOpacity), 0.2), "")"'
+    );
+}
+
+.flexContainer {
+    display: flex;
+    justify-content: space-around;
+}
```
jonathanadler, line 19: In accordance with what I wrote in `EventWiringExample.tsx`
The correct way to write this would be:
```suggestion
.root {
    display: flex;
    justify-content: space-around;
}

.event {
    -st-mixin: TPAEvent(
            TimeColor '"--timeColor"',
            TitleColor '"--titleColor"',
            AllDayBackground '"--allDayBackground"',
            FontStyle '"--fontStyle"',
            AllDayOpacity '"unit(fallback(zeroAsTrue(--allDayOpacity), 0.2), "")"'
    );
}
```

src/components/Event/Event.st.css
ItayTur: I meant the first one. the Current gets the fallback: if we didn't get the override for the style then we use the default. why ?

src/components/Event/Event.st.css
ItayTur: This is not dealing with the `rtl`, it's just dealing the padding in the `rtl` case.

src/components/Event/Event.st.css
ItayTur: Flex was the first thing i try, and it is the better practice. It's just that the shadowing requirement force the component to be relative positioned because the `:before` element needs to be absolute in order to be on top of the event. so  flex didn't work. I actually wanted to talk to you about it. You have an idea what could be done to achieve shadowing with no abstract ? I think that if we could figure this one out flex would work

src/components/Event/Event.st.css
ItayTur: what I said above...

src/components/Event/Event.st.css
ItayTur: for multi day ? multi day just have a background color there is no layout here

src/components/Event/Event.st.css
jonathanadler: I meant that usually our fallback variables are called `Default...` and not `Current...`

src/components/Event/Event.st.css
jonathanadler: what do you mean by `shadowing`?

src/components/Event/Event.st.css
jonathanadler: It's kind of the same thing 😛 
But what I meant is that we prefer to avoid having specific solutions for `ltr` and `rtl`.

In addition, we use the `TPAComponentsConsumer` to pass the `rtl` state to stylable, when needed

src/components/Event/Event.st.css
jonathanadler: and do we really need all these paddings?

src/components/Event/Event.st.css
jonathanadler: but regarding `multiday` if it's only the background that changes, i'm not sure we need it in the component.
I think it's a specific feature that you guys need, and that you can override the bg color from your app

src/components/Event/docs/EventWiringExample.tsx
ItayTur: event doesn't have className in it's props. it should have ? anyway, i updated the names

src/components/Event/Event.st.css
ItayTur: on hover and when selected there should be shadow div. like black but transparent. you can view it here: https://app.abstract.com/projects/3c6113bf-5996-4c28-aaa3-14ab8b599eb4/branches/master/commits/latest/files/7997d7f2-7bd7-4fe3-9e22-017d7c436e04/layers/330883A3-2E9F-45E5-9AFE-DEE02851C9B2?collectionId=117a41e8-de09-499c-b98a-daa1409657c2&collectionLayerId=8ceb6c0c-c21a-42b4-bc70-48e7aeedbb37

src/components/Event/Event.st.css
ItayTur: I followed the checkbox for conventions. I don't mind change it, just tell me to do so.
About the paddings I'll fix it.

src/components/Event/Event.st.css
ItayTur: flex didn't work with the absolute position of the :before element.
I took the idea from google, but if you have a better solution it would be great.

ItayTur: to drop the multiday ? it is just the background i agree, but it's part of the requirements ron gave me....

src/components/Event/Event.st.css
```diff
@@ -0,0 +1,130 @@
...
+:vars {
+    FontStyle: --overridable;
+    TimeColor: --overridable;
+    TitleColor: --overridable;
+    AllDayBackground:  --overridable;
+    AllDayOpacity: --overridable;
+    TopPadding: --overridable;
+    RightPadding: --overridable;
+    BottomPadding: --overridable;
+    LeftPadding: --overridable;
```
jonathanadler, line 36: Okay, so after consulting with our designers, the overridable variables should be like in this table https://zeroheight.com/7sjjzhgo2/p/48969b-calendar/t/34d1d4 :

* textFont
* timeColor
* titleColor
* bgColor
* hoverOverlayColor
* selectedOverlayColor
* fulldayBgColor

No padding is needed here, it should be fixed in the component

src/components/Event/Event.st.css
jonathanadler: So this should be changed to behave similar to this:
https://codepen.io/Jonnybordo/pen/KKVpRGX

src/components/Event/Event.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { TPAComponentsConsumer } from '../TPAComponentsConfig';
+import { Text } from '../Text';
+import { EVENT_DATA_HOOKS, EVENT_DATA_KEYS } from './dataHooks';
+import styles from './Event.st.css';
+
+export interface EventProps {
+  time?: string;
+  title: string;
+  showTime?: boolean;
+  multiday?: boolean;
```
jonathanadler, line 11: multiday isn't needed
You can override the `bgColor` in your app

src/components/Event/Event.tsx
```diff
@@ -0,0 +1,64 @@
+import * as React from 'react';
+import { TPAComponentsConsumer } from '../TPAComponentsConfig';
+import { Text } from '../Text';
+import { EVENT_DATA_HOOKS, EVENT_DATA_KEYS } from './dataHooks';
+import styles from './Event.st.css';
+
+export interface EventProps {
```
jonathanadler, line 7:
```suggestion
import styles from './Event.st.css';
import { TPAComponentProps } from '../../types';

export interface EventProps extends TPAComponentProps {
```
You should add this.
Sorry I thought it was part of the generator

src/components/Event/Event.tsx
```diff
@@ -0,0 +1,64 @@
...
+}
+
+/** Event */
+export class Event extends React.Component<EventProps> {
+  static displayName = 'Event';
+  static defaultProps: DefaultProps = {
+    showTime: true,
+    multiday: false,
+    selected: false,
+    'data-hook': EVENT_DATA_HOOKS.Event,
```
jonathanadler, line 29: We actually don't want a default value here, because then consumers will start using it, and we won't be able to change `EVENT_DATA_HOOKS.Event`, or else we'll break their app

src/components/Event/Event.visual.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+visualize('Event', () => {
+  story('render', () => {
+    snap('default', <Event {...defaultProps} />);
+    snap('time', <Event time="23:23" {...defaultProps} />);
+    snap('timeShown', <Event showTime {...defaultProps} />);
+    snap('multiday', <Event multiday {...defaultProps} />);
+    snap('selected', <Event selected {...defaultProps} />);
+  });
+});
```
jonathanadler, line 17: Awesome!

src/components/Event/Event.tsx
```diff
@@ -0,0 +1,64 @@
...
+    showTime: true,
+    fullday: false,
+    selected: false,
+  };
+
+  getDataAttributes() {
+    const { fullday, selected, showTime } = this.props;
+
+    return {
+      [EVENT_DATA_KEYS.IsMultiday]: fullday,
```
jonathanadler, line 35:
```suggestion
      [EVENT_DATA_KEYS.IsFullDay]: fullday,
```

src/components/Event/dataHooks.ts
```diff
@@ -0,0 +1,6 @@
+export enum EVENT_DATA_KEYS {
+  IsTimeShown = 'data-time-shown',
+  IsMultiday = 'data-fullday',
```
jonathanadler, line 3:
```suggestion
  IsFullDay = 'data-fullday',
```

PR #347, opened 2020-06-11
Calendar-Popover

ItayTur:


src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,100 @@
...
+  arrowTop: string;
+  withShadow: boolean;
+}
+
+/** Popover */
+export class Popover extends React.Component<PopoverProps> {
+  static displayName = 'Popover';
+  static defaultProps: DefaultProps = {
+    title: '',
+    'data-hook': POPOVER_DATA_HOOKS.Popover,
```
jonathanadler, line 32: as in the other PR's
it's best not to add a default data-hook.
It will encourage consumers to use this instead of adding their own, which means this will be a binding contract, and we prefer to avoid any unnecessary api

src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,100 @@
...
+import { TPAComponentsConsumer } from '../TPAComponentsConfig';
+import { POPOVER_DATA_HOOKS, POPOVER_DATA_KEYS } from './dataHooks';
+import styles from './Popover.st.css';
+
+export interface PopoverProps {
+  title?: string;
+  onClose(): void;
+  withArrow?: boolean;
+  rightArrow?: boolean;
+  arrowTop?: string;
```
jonathanadler, line 14: I think we should also accept `number` for the `arrowTop`

jonathanadler: In addition, i'm still unsure about this API.
I feel that have one prop, such as `arrowPosition` would be a bit more intuitive.
Something like:
```js
type ArrowPosition = {
  side: 'left' | 'right';
  top: string | number;
}
...
arrowPosition: ArrowPosition;
```

or something like that

src/components/index.ts
```diff
@@ -36,3 +36,4 @@ export { Tags } from './Tags';
 export { FloatingDropdown } from './FloatingDropdown';
 
 export { CheckboxGroup, CheckboxGroupProps, Layout };
+export { Popover } from './Popover';
```
jonathanadler, line 39: I think we can remove this file, since I don't think we should let users use this outside of the calendar for now.

src/components/Popover/Popover.driver.ts
```diff
@@ -0,0 +1,35 @@
...
+      return (await base.attr(POPOVER_DATA_KEYS.WithShadow)) === 'true';
+    },
+    async isRightArrow() {
+      return (await base.attr(POPOVER_DATA_KEYS.RightArrow)) === 'true';
+    },
+    async withTitle() {
+      return (await base.attr(POPOVER_DATA_KEYS.Title)) === 'title';
+    },
+    async withArrowTop() {
+      return (await base.attr(POPOVER_DATA_KEYS.ArrowTop)) === '50px';
```
jonathanadler, line 32: why is this hardcoded to `50px`?
This should be something like:
```suggestion
    async hasArrowTop(value: string | number) {
      return (await base.attr(POPOVER_DATA_KEYS.ArrowTop)) === value;
```

or maybe just:
```suggestion
    async getArrowTop() {
      return base.attr(POPOVER_DATA_KEYS.ArrowTop);
```

jonathanadler: We might not have talked about it, 
but the drivers are also exported for the consumers to be used as testkits.
So the API of the driver is also a contract we have with our consumers

src/components/Popover/Popover.driver.ts
```diff
@@ -0,0 +1,35 @@
...
+    },
+    async withShadow() {
+      return (await base.attr(POPOVER_DATA_KEYS.WithShadow)) === 'true';
+    },
+    async isRightArrow() {
+      return (await base.attr(POPOVER_DATA_KEYS.RightArrow)) === 'true';
+    },
+    async withTitle() {
+      return (await base.attr(POPOVER_DATA_KEYS.Title)) === 'title';
+    },
```
jonathanadler, line 30: I think in the case of drivers, it's better to have names like `hasArrow` or `hasShadow`.
That way we get tests like this:
```js
const hasArrow = await driver.hasArrow();
expect(hasArrow).toBe(true);
```

src/components/Popover/Popover.st.css
```diff
@@ -0,0 +1,162 @@
...
+
+/*Overrides*/
+:vars {
+    BorderColor: --overridable;
+    Radius: --overridable;
+    IconColor: --overridable;
+    TopPadding: --overridable;
+    RightPadding: --overridable;
+    BottomPadding: --overridable;
+    LeftPadding: --overridable;
```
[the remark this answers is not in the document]
ItayTur: it is going to be a library component, so giving this ability could help when specific design is needed. but yes i see what you mean. I'll drop it

src/components/Popover/Popover.st.css
```diff
@@ -0,0 +1,162 @@
...
+.root:rtl .close {
+    left: value(CurrentIconLeftPadding);
+}
+
+.root:rtl {
+    text-align: right;
+}
+
+.root:rtl .container {
+    float: right;
```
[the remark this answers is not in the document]
ItayTur: you right

src/components/Popover/Popover.st.css
```diff
@@ -0,0 +1,162 @@
...
+
+.close {
+    position: absolute;
+    top: value(CurrentIconTopPadding);
+    right: value(CurrentIconRightPadding);
+    -st-mixin: TPAIconButton(IconColor value(CurrentIconColor));
+}
+
+.root:rtl .close {
+    left: value(CurrentIconLeftPadding);
```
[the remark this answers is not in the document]
ItayTur: why

src/components/Popover/Popover.tsx
ItayTur: to keep in the props ?

src/components/index.ts
ItayTur: how will the calendar could use it than ?

src/components/Popover/Popover.tsx
ItayTur: i think about arrowTop: number and side: sidesEnum

PR #349, opened 2020-06-25
Dropdown | Making a11y adjustments

NitzanSirota:
Add aria-activedescendant, replace `aria-haspopup` with `role=listbox`, add `contentId` to the core dropdown.

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -14,6 +14,8 @@ import { DropdownNativeSelect } from './DropdownNativeSelect';
 import { deprecationLog, wrap, unwrap } from '../../common/deprecationLog';
 import { IDOMid } from 'wix-ui-core/dist/es/src';
 
+const uniqueId = require('lodash/uniqueId');
+
 export enum DROPDOWN_ALIGNMENT {
   center = 'center',
 }
@@ -44,7 +46,6 @@ interface DefaultProps {
   placement: Placement;
   upgrade: boolean;
   mobileNativeSelect: boolean;
-  optionsContainerId: string;
 }
 
 interface State {
@@ -58,6 +59,7 @@ interface State {
  * Single selection dropdown.
  * */
 export class Dropdown extends React.Component<DropdownProps, State> {
+  private readonly contentId: string;
   static displayName = 'Dropdown';
   static contextType = TPAComponentsContext;
   static defaultProps: DefaultProps = {
@@ -66,11 +68,12 @@ export class Dropdown extends React.Component<DropdownProps, State> {
     placement: 'bottom',
     mobileNativeSelect: false,
     upgrade: false,
-    optionsContainerId: 'dropdown-options-container',
   };
+
   constructor(props) {
     super(props);
     wrap('Button');
+    this.contentId = props.optionsContainerId ? props.optionsContainerId : uniqueId('dropdown-options-container_');
   }
 
   componentDidMount(): void {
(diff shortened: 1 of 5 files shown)


src/components/Dropdown/Dropdown.tsx
```diff
@@ -61,6 +66,7 @@ export class Dropdown extends React.Component<DropdownProps, State> {
     placement: 'bottom',
     mobileNativeSelect: false,
     upgrade: false,
+    optionsContainerId: 'dropdown-options-container',
```
NitzanSirota, line 69: I'm not fully complete with the default value, but it has to be here to give the options dom id in order to put aria-activedescendant.
Do you think we should add something random here?
@jonathanadler

src/components/Dropdown/Dropdown.tsx
```diff
@@ -104,11 +112,28 @@ export class Dropdown extends React.Component<DropdownProps, State> {
...
+      selectedOption,
+      isOpen: false,
+      ariaActivedescendant: null,
+    });
     onChange &&
       onChange(this.props.options.find(({ id }) => selectedOption.id === id));
   };
 
+  private readonly onOptionHover = (option: DropdownOptionProps & IDOMid) => {
+    const ariaActivedescendant = option ? option._DOMid : null;
```
jonathanadler, line 125: I feel like I'm missing something.
no one will ever pass `_DOMid` to the option, because it doesn't exist in the `DropdownOptionProps` interface.
Why not just use the option's `id` property?

src/components/Dropdown/Dropdown.tsx
jonathanadler, line 69: Why no call this `id` 🤔 ?

jonathanadler: yeah, we should add a random suffix
or add a static `_index` property that keeps count or something like this.
I think that, soon, when we publish `v2`, we can just require our consumers to use this `id` prop

src/components/Dropdown/Dropdown.tsx
NitzanSirota: this ensures us that the option will have a domId ->https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/dropdown-content/DropdownContent.tsx#L150
You can also checkout the onOptionHover function there.

NitzanSirota: I used it that way because that what I saw they are doing in ui-core and also this is the function's signature.

src/components/Dropdown/Dropdown.tsx
NitzanSirota: it's a very specific id for the options container who is generated inside ui-core. You can check how here : https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/dropdown-content/DropdownContent.tsx#L217 and https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/dropdown-content/DropdownContent.tsx#L150

NitzanSirota:
How do you think I should implement the random default id that will also be testable?
And I'll check why the a11y check is failing, sometimes it is just flaky 🤷

NitzanSirota:
a11y tests are really failing. I'll consult with Neil about that.
Is there any way to ignore specific tests if he tells me that it is fine by him?

jonathanadler:
Ok, so I guess just add the random id generator & fix the tests
Other than that, looks cool-ish.
I actually think that the wix-ui-core Dropdown isn't good, and that we should change its implementation.
but maybe next quarter 🤷

NitzanSirota:
It definitely has some problems. I think the next dropdown iteration should be done with Neil in the design phase.

jonathanadler:
> It definitely has some problems. I think the next dropdown iteration should be done with Neil in the design phase.

fo sho

NitzanSirota:
Added the suffix to the default id.
Ignored the component in a11y tests.

PR #356, opened 2020-07-10
Update Dropdown.tsx

zhukovka:
recompute selected option if selected option is changed from next props

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -89,7 +89,7 @@ export class Dropdown extends React.Component<DropdownProps, State> {
   }
 
   static getDerivedStateFromProps(nextProps, state) {
-    if (state.selectedOption && nextProps.initialSelectedId !== state.selectedOption.id || !state.selectedOption) {
+    if (!state.selectedOption || nextProps.initialSelectedId !== state.selectedOption.id) {
       return {
         selectedOption:
           nextProps.options.find(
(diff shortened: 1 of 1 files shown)


src/components/Dropdown/Dropdown.tsx
```diff
@@ -89,15 +89,17 @@ export class Dropdown extends React.Component<DropdownProps, State> {
   }
 
   static getDerivedStateFromProps(nextProps, state) {
+    if (state.selectedOption && nextProps.initialSelectedId !== state.selectedOption.id || !state.selectedOption) {
```
jonathanadler, line 92:
```suggestion
    if (!state.selectedOption || nextProps.initialSelectedId !== state.selectedOption.id) {
```

src/components/Dropdown/Dropdown.tsx
```diff
@@ -89,15 +89,17 @@ export class Dropdown extends React.Component<DropdownProps, State> {
...
+      return {
+        selectedOption:
+          nextProps.options.find(
+            option => option.id === nextProps.initialSelectedId,
+          ) || null,
+      };
+    }
     if (state.selectedOption) {
       return null;
     }
```
jonathanadler, line 102: In this case, we should always return null, and not only if `state.selectedOption`

shlomitc:
is this PR still relevant?

jonathanadler:
Closing for now

PR #364, opened 2020-08-02
Feat/checkbox box and more

GuyPerkal:
I added a wrapper box for the checkbox, and an optional price prop. I thought about calling the price property 'suffix' instead, but in the [UI docs for the radio button](https://zeroheight.com/7sjjzhgo2/p/05a744-05-selections/b/743582) they explicitly say the suffix has to be a price.

I also fixed a bug with the disabled prop not being propagated to CoreCheckbox.

### DIFF
--- a/src/components/Checkbox/Checkbox.st.css
+++ b/src/components/Checkbox/Checkbox.st.css
@@ -16,6 +16,7 @@
 :vars {
     /*Defaults*/
     DefaultTextColor: color-5;
+    DefaultBorderColor: color-5;
     DefaultBoxBorderColor: color-5;
     DefaultIconColor: color-8;
     DefaultBoxColor: color-1;
@@ -26,12 +27,14 @@
 :vars {
     /*Overrides*/
     TextColor: --overridable;
+    BorderColor: --overridable;
     BoxBorderColor: --overridable;
     IconColor: --overridable;
     BoxColor: --overridable;
 }
 
 :vars {
+    CurrentBorderColor: color(fallback(value(BorderColor), value(DefaultBorderColor)));
     CurrentBoxBorderColor: color(fallback(value(BoxBorderColor), value(DefaultBoxBorderColor)));
     CurrentIconColor: color(fallback(value(IconColor), value(DefaultIconColor)));
     CurrentTextColor: color(fallback(value(TextColor), value(DefaultTextColor)));
@@ -91,7 +94,7 @@
     height: 16px;
 
     border: 1px solid;
-    border-color: applyOpacity(value(CurrentTextColor), 0.6);
+    border-color: applyOpacity(value(CurrentBorderColor), 0.6);
     border-radius: 1px;
 
     font-size: 13px;
--- a/src/components/Checkbox/Checkbox.st.css
+++ b/src/components/Checkbox/Checkbox.st.css
@@ -43,6 +43,8 @@
     border: 1px solid applyOpacity(value(CurrentBorderColor), 0.6);
     margin-bottom: 16px;
     cursor: pointer;
+    width: 100%;
+    box-sizing: border-box;
 }
 
 .root {
@@ -54,6 +56,7 @@
     box-sizing: border-box;
 
     font: value(Font);
+    flex: 1;
 }
 
 .root * {
@@ -62,6 +65,7 @@
 
 .root::childContainer {
     display: flex;
+    flex: 1;
 }
 
 .root::box {
@@ -103,6 +107,14 @@
 .label {
     display: inline-block;
     color: value(CurrentTextColor);
+}
+
+.price {
+    margin-left: 5px;
+}
+
+.box .price {
+    margin-left: auto;
 }
 
 /* STATES */
--- a/src/components/Checkbox/Checkbox.st.css
+++ b/src/components/Checkbox/Checkbox.st.css
@@ -49,6 +49,7 @@
     box-sizing: border-box;
 
     font: value(Font);
+    transition: background-color 0.2s linear, border-color 0.2s linear;
 }
 
 .root * {
@@ -81,11 +82,34 @@
 
 .root:box {
     display: flex;
-    padding: 16px;
     background-color: color(value(CurrentBoxColor));
     border: 1px solid applyOpacity(value(CurrentBoxBorderColor), 0.6);
-    cursor: pointer;
     box-sizing: border-box;
+    position: relative;
+}
+
+.root:box::after {
+    content: '';
+    width: 100%;
+    height: 100%;
+    position: absolute;
+    top: -1px;
+    left: -1px;
+    display: block;
+    background-color: applyOpacity(value(CurrentIconColor), 0.1);
+    border: 1px solid color(value(CurrentIconColor));
+    pointer-events: none;
+    opacity: 0;
+    transition: opacity 0.2s linear;
+}
+
+.root:box::childContainer {
+    padding: 16px;
+    padding-left: 0;
+}
+
+.root:box::box {
+    padding-left: 16px;
 }
 
 .icon {
@@ -100,6 +124,7 @@
     font-size: 13px;
     line-height: 1.2;
     vertical-align: middle;
+    transition: border-color 0.2s linear;
 }
 
 .icon svg {
@@ -134,11 +159,7 @@
 }
 
 /* STATES */
-.root:box:disabled {
-    cursor: default;
-}
-
-.root:box:hover, .root:box:focus-within {
+.root:hover:box, .root:focus-within:box {
     border-color: applyOpacity(value(CurrentBoxBorderColor), 1);
 }
 
@@ -146,20 +167,11 @@
     border-color: applyOpacity(value(CurrentBoxBorderColor), 0.6);
 }
 
-.root:box:checked {
-    background-color: applyOpacity(value(CurrentIconColor), 0.1);
-    border-color: value(CurrentIconColor);
+.root:box:checked::after {
+    opacity: 1;
 }
 
-.root:box:checked:not(:disabled):hover, .root:box:checked:not(:disabled):focus-within {
-    border-color: applyOpacity(value(CurrentIconColor), 0.6);
-}
-
-.root:box:not(:disabled):hover .icon {
-    border-color: applyOpacity(value(CurrentBoxBorderColor), 1);
-}
-
-.root:hover .icon {
+.root:hover .icon, .root:focus-within .icon {
     border-color: applyOpacity(value(CurrentBorderColor), 1);
 }
 
--- a/src/components/Checkbox/Checkbox.st.css
+++ b/src/components/Checkbox/Checkbox.st.css
@@ -59,7 +59,12 @@
 
 .root::box {
     line-height: 0;
-    margin-right: 12px;
+    display: flex;
+}
+
+.root::box:after {
+    content: '';
+    width: 12px;
 }
 
 .core {
@@ -109,9 +114,14 @@
 }
 
 .label.suffixed {
-    margin-right: 5px;
     flex: 1;
     white-space: nowrap;
+    display: flex;
+}
+
+.label.suffixed:after {
+    content: '';
+    width: 5px;
 }
 
 .suffix {
--- a/src/components/Checkbox/Checkbox.st.css
+++ b/src/components/Checkbox/Checkbox.st.css
@@ -62,7 +62,7 @@
     display: flex;
 }
 
-.root::box:after {
+.root::box::after {
     content: '';
     width: 12px;
 }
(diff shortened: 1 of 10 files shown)


jonathanadler:
@GuyPerkal 
sorry I didn't tell you earlier.
You need to sign your commits, since this is a public repo.
See here for more: https://github.com/wix/wix-style-react/blob/master/docs/contribution/CREATE_PR.md

GuyPerkal:
Commits are now signed.

GuyPerkal:
Commits are now signed.

jonathanadler:
> Commits are now signed.

Awesome, thanks!

Now we just need you to make a new (signed) commit so that the build will run 😁 

As you suspected, the `price` prop mentioned in zeroheight was just an example, and it shouldn't have been titled that way there.
And it is really specific to have a `price` prop in a checkbox component.

So yeah, let's change it to `suffix`

src/components/Checkbox/Checkbox.st.css
```diff
@@ -95,7 +109,44 @@
...
+    border-color: value(CurrentIconColor);
+}
+
+.box.checked:not(.disabled):hover, .box.checked:not(.disabled):focus-within {
+    border-color: applyOpacity(value(CurrentIconColor), 0.6);
+}
+
+.box:not(.disabled):hover .icon {
+    border-color: applyOpacity(value(CurrentBorderColor), 1);
+}
```
jonathanadler, line 144: The correct syntax for using Stylable states is like this:
```css
.root:disabled {
  ...
}
```

So you'll need to adjust this code accordingly.
Not that the state is configured on the `.root` class, and I think it won't work on other classes.

So for instance I think it should be written like this:
```css
.root:disabled.box {
  ...
}
```

src/components/Checkbox/Checkbox.st.css
```diff
@@ -95,7 +109,44 @@
...
+
+.box:last-of-type {
+    margin-bottom: 0;
+}
```
jonathanadler, line 148: It seems this code should be in the CheckboxGroup component's stylesheet.
We usually don't want the component itself affecting its surrounding.

src/components/Checkbox/Checkbox.tsx
```diff
@@ -17,6 +18,8 @@ export interface CheckboxProps extends TPAComponentProps {
   indeterminate?: boolean;
   error?: boolean;
   name?: string;
+  box?: boolean;
```
jonathanadler, line 21: In order to align with other components (radioButton for one), instead of having a `box` prop, please use a `theme` prop.
You can consult your fellow teammate @mariakwix on the prop she's adding for the RadioButton component

src/components/Checkbox/Checkbox.tsx
```diff
@@ -72,40 +75,53 @@ export class Checkbox extends React.Component<CheckboxProps> {
...
-        checkedIcon={iconContent}
-        uncheckedIcon={iconContent}
-        indeterminateIcon={iconContent}
-        indeterminate={indeterminate}
-        checked={checked}
-        onChange={onChange}
-        name={name}
+      <div
+        className={`${box && classes.box} ${checked &&
+          classes.checked} ${disabled && classes.disabled} `}
```
jonathanadler, line 94: In Stylable, the root element should receive the `.root` class.
and Stylable states should be used instead of regular classes (see the `root` below).
So since the dom structure has changed here, the stylesheet structure should change as well
(and again, I would like you to talk to @mariakwix and figure out how to solve this together, since the same solution should appear in the `RadioButton` component she's working on - the best approach would be to have a single abstraction/component with these themes, so that both `RadioButton` and `Checkbox` could use it without code duplication)

jonathanadler: Note that this DOM change, also caused the `Checkbox`'s tests to fail (you can see in the status check at the bottom of the PR page)

src/components/Checkbox/docs/examples.ts
```diff
@@ -19,3 +19,19 @@ export const exampleWithDisabled = `
...
+<Checkbox box checked onChange={val => console.log(val)} label="A checkbox!" />
+`;
+
+export const exampleWithDisabledCheckedBox = `
+<Checkbox box checked disabled onChange={val => console.log(val)} label="I'm a disabled, checked box 😶" />
+`;
+
+export const exampleWithSuffix = `
+<Checkbox box suffix="$1,000" onChange={val => console.log(val)} label="An expansive checkbox." />
+`;
```
jonathanadler, line 37: Great examples 👍 
Let's also add an example, where the container is small, and so the content will wrap as described here: https://zeroheight.com/7sjjzhgo2/p/05a744-05-selections/t/76a987

src/components/Checkbox/docs/index.story.tsx
```diff
@@ -68,6 +68,30 @@ export default {
...
+              title: 'With disabled, checked box',
+              source: examples.exampleWithDisabledCheckedBox,
+            },
+          ].map(code),
+          ...[
+            {
+              title: 'With suffix',
+              source: examples.exampleWithSuffix,
+            },
+          ].map(code),
```
jonathanadler, line 94: I see that this code was written this way before your PR, but instead of having multiple 
```js
.map(code)
```

let's just have one array of examples, and one invocation of `map`.
ie
```js
...[
      {
        title: 'Checked',
        source: examples.exampleWithChecked,
      },
      ...
      {
        title: 'With box',
        source: examples.exampleWithBox,
      },
      ...
  ].map(code),
```

or even:
```js
const codeExamples = [
      {
        title: 'Checked',
        source: examples.exampleWithChecked,
      },
      ...
      {
        title: 'With box',
        source: examples.exampleWithBox,
      },
      ...
];

...

...codeExamples.map(code)
```

src/components/Checkbox/Checkbox.st.css
```diff
@@ -44,6 +56,7 @@
     box-sizing: border-box;
 
     font: value(Font);
+    flex: 1;
```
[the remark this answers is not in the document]
GuyPerkal: The .box container is `display: flex`, I added this so that the suffix will be aligned to the right.

src/components/Checkbox/Checkbox.tsx
GuyPerkal: While it's possible to share some CSS between the RadioButton's and Checbox's box, I'm not sure it will be worth the added complexity of passing the state along as arguments (selected, disabled, hovered). There isn't any logic in the box, it's just some CSS.

src/components/Checkbox/Checkbox.tsx
```diff
@@ -9,6 +9,9 @@ import { st, classes } from './Checkbox.st.css';
 interface OnChangeEvent extends React.ChangeEvent<HTMLInputElement> {
   checked: boolean;
 }
+
+type Theme = 'box';
```
jonathanadler, line 13: We use this convention:
```suggestion
export enum CheckboxTheme {
  Default = 'default',
  Box = 'box',
};
```
Then we use it in the rest of the files (tests, storybook, etc.)

src/components/Checkbox/Checkbox.st.css
```diff
@@ -52,10 +52,30 @@
...
     display: flex;
+    flex: 1;
+    flex-wrap: wrap;
+    align-items: center;
 }
 
 .root::box {
-    line-height: 20px;
+    line-height: 0;
+    margin-right: 12px;
```
jonathanadler, line 62: Note that we need to support `rtl` preferably out-of-the-box.
So instead of using `margin-right` and such, we prefer to us `::after` and give it the `width` we expect the `margin` to be.

src/components/Checkbox/Checkbox.tsx
```diff
@@ -72,40 +77,51 @@ export class Checkbox extends React.Component<CheckboxProps> {
...
+          indeterminate={indeterminate}
+          checked={checked}
+          onChange={onChange}
+          name={name}
+          disabled={disabled}
+        >
           <div
             data-hook={CHECKBOX_DATA_HOOKS.LabelWrapper}
-            className={classes.label}
+            className={`${classes.label} ${suffix && classes.suffixed}`}
```
jonathanadler, line 116: We use `classnames` in this library, so you can write this as:
```suggestion
            className={classnames(classes.label, {[classes.suffixed]:  suffix })}
```
and of course you'll need to add an import

GuyPerkal:
@jonathanadler done

src/components/Checkbox/Checkbox.st.css
```diff
@@ -27,6 +27,7 @@
     TextColor: --overridable;
     BorderColor: --overridable;
     IconColor: color-8;
+    BoxColor: color-1;
```
jonathanadler, line 30: I just noticed that you're not using `BoxColor` anywhere.
Why is that?

src/components/Checkbox/docs/index.story.tsx
```diff
@@ -86,7 +101,7 @@ export default {
               params: {
                 colors: [
                   {
-                    label: 'Icon Color',
+                    label: 'Icon/Box Color',
```
jonathanadler, line 104: Icon and box colors aren't supposed to be the same.
We'll need the two separate colors to be customizable.

src/components/Checkbox/Checkbox.st.css
GuyPerkal: You're right, I forgot about it, I'll add it.

src/components/Checkbox/docs/index.story.tsx
GuyPerkal: Ah OK, I thought because they both use color-5 it should be the same var.

GuyPerkal:
@jonathanadler everything should now be correctly configurable.

src/components/Checkbox/Checkbox.st.css
```diff
@@ -16,31 +16,33 @@
...
     DefaultIconColor: color-8;
+    DefaultBoxColor: color-1;
     DisabledColor: color-3;
     Font: font("{theme: 'Body-M', size: '16px', lineHeight: '24px'}");
 }
 
 :vars {
     /*Overrides*/
     TextColor: --overridable;
-    BorderColor: --overridable;
```
jonathanadler, line 28: We can't remove or rename existing variables, because consumers might already be using them, so this might break their projects.
I agree that the naming here is confusing and bad, but we'll have to leave it like that

src/components/Checkbox/Checkbox.st.css
```diff
@@ -64,7 +91,7 @@
     height: 16px;
 
     border: 1px solid;
-    border-color: applyOpacity(value(CurrentBorderColor), 0.6);
+    border-color: applyOpacity(value(CurrentTextColor), 0.6);
```
jonathanadler, line 94: This is also a breaking change

src/components/Checkbox/Checkbox.st.css
GuyPerkal: Is it OK if I bring back the overridable BorderColor but keep it off the storybook? I think it should be deprecated.

jonathanadler: For the moment it's better to leave it in the storybook.
I'll raise this issue in our team, and we have our deprecation process for this.
But for now, just leave the api as it was + add the addition api for the box theme

GuyPerkal:
Alright, I brought back the overridable BorderColor.

src/components/Checkbox/Checkbox.st.css
```diff
@@ -95,9 +118,49 @@
...
+    border-color: applyOpacity(value(CurrentIconColor), 0.6);
+}
+
+.root:box:not(:disabled):hover .icon {
+    border-color: applyOpacity(value(CurrentBoxBorderColor), 1);
+}
+
 .root:hover .icon {
-    border-color: applyOpacity(value(CurrentBorderColor), 1);
+    border-color: applyOpacity(value(CurrentBoxBorderColor), 1);
```
jonathanadler, line 163: Shouldn't this stay `CurrentBorderColor`?
It's not under the `box` state

GuyPerkal: Right, I changed it back.

jonathanadler:
Great work!

jonathanadler:
Hey, 
So I have some comments from @johannaasseraf, our designer:

> Nice work! 👏🏻
> 
> * Add 0.2s transition between states
> * Make sure the selected state color is an overlay on top of the box background, as specified here: https://zeroheight.com/7sjjzhgo2/p/05a744-05-selections/t/97d023
> * Box non-selected state needs to have hover+focus state similar to checkbox border (currently none) https://zeroheight.com/7sjjzhgo2/p/05a744-05-selections/t/108969 
> * Box selected state shouldn't have different hover state (currently getting lighter on hover)

GuyPerkal:
Pushed fixes.

jonathanadler:
https://zeroheight.com/7sjjzhgo2/p/05a744-05-selections/t/32079e

PR #370, opened 2020-08-10
feat: Spinner

rustemk13:


jonathanadler:
@rustemk13 Sorry didn't get to go over this today
I'll do it tomorrow morning.
Just one important thing I totally forgot!
We need to sign our commits in order to contribute to public repos.

So please add a signature to your commits
You can find a guide here: https://github.com/wix/wix-style-react/blob/master/docs/contribution/CREATE_PR.md

src/components/Spinner/Spinner.driver.ts
```diff
@@ -0,0 +1,16 @@
...
+import { UniDriver } from 'wix-ui-test-utils/unidriver';
+
+export interface SpinnerDriver extends BaseUniDriver {
+  isMobile(): Promise<boolean>;
+}
+
+export const spinnerDriverFactory = (base: UniDriver): SpinnerDriver => {
+  return {
+    ...baseUniDriverFactory(base),
+    isMobile: async () => (await base.attr('data-mobile')) === 'true',
```
jonathanadler, line 14: The `isMobile` prop isn't needed for this component.
So you can just remove it

src/components/Spinner/Spinner.spec.tsx
```diff
@@ -0,0 +1,51 @@
...
+    const driver = createDriver(<Spinner />);
+    expect(await driver.exists()).toBe(true);
+  });
+
+  it('should use mobile design', async () => {
+    const driver = createDriver(
+      TPAComponentsWrapper({ mobile: true })(<Spinner />),
+    );
+    expect(await driver.isMobile()).toBe(true);
+  });
```
jonathanadler, line 25: Again, I think this can be removed, since there's no special design for mobile.
In fact, all references to mobile can be removed for this component I guess

src/components/Spinner/Spinner.st.css
```diff
@@ -0,0 +1,60 @@
...
+
+:vars {
+    UsedPathColor: color(fallback(value(PathColor), value(DefaultPathColor)));
+}
+
+.root {
+    -st-states: mobile, centered, static;
+
+    animation: rotate 2s linear infinite;
+    z-index: 2;
```
jonathanadler, line 22: Not sure this is needed.
It might interfere with other consumers use of zIndex.
So I think we can remove this, and if a consumer needs to add a `z-index` they can add it using a className which is passed to the component

src/components/Spinner/Spinner.st.css
```diff
@@ -0,0 +1,60 @@
...
+.root:centered {
+    position: absolute;
+    left: 50%;
+}
+
+.root:static, .root:static .circle {
+    animation: none;
+}
+
+.root:mobile {}
```
jonathanadler, line 39: Again, can be removed

src/components/Spinner/Spinner.st.css
```diff
@@ -0,0 +1,60 @@
...
+}
+
+.circle {
+    stroke: value(UsedPathColor);
+    animation: dash 1.5s ease-in-out infinite;
+}
+
+.root:centered {
+    position: absolute;
+    left: 50%;
```
jonathanadler, line 32: is `left: 50%` enough?
Maybe we need to add `transform: translateX(-50%);`?

src/components/Spinner/Spinner.tsx
```diff
@@ -0,0 +1,54 @@
...
+  SPINNER_TYPES,
+} from './types';
+import { st, classes } from './Spinner.st.css';
+import { TPAComponentsConsumer } from '../TPAComponentsConfig';
+
+export class Spinner extends React.Component<SpinnerProps, SpinnerState> {
+  static displayName = 'Spinner';
+  static defaultProps: SpinnerDefaultProps = {
+    type: SPINNER_TYPES.regular,
+    width: 100,
```
jonathanadler, line 15: By looking in the zeroheight spec, I think the default width should be `50px`
Also, maybe instead of `width` we can call it `diameter`? since it's only one prop

src/components/Spinner/Spinner.tsx
```diff
@@ -0,0 +1,54 @@
...
+              cx="25"
+              cy="25"
+              r="20"
+              fill="none"
+              strokeWidth={type === SPINNER_TYPES.regular ? 4 : 1}
+              className={classes.circle}
+            />
+          </svg>
+        )}
+      </TPAComponentsConsumer>
```
jonathanadler, line 51: Again, mobile can be removed:

```suggestion
        <svg
          viewBox="0 0 50 50"
          className={st(
            classes.root,
            { centered: !!isCentered, static: !!isStatic },
            className,
          )}
          data-mobile={mobile}
          style={{
            width: `${width}px`,
            height: `${width}px`,
            top: isCentered ? `calc(50% - ${width / 2}px)` : 0,
            margin: isCentered ? `0 0 0 -${width / 2}px` : 0,
          }}
          data-hook={this.props['data-hook']}
        >
          <circle
            cx="25"
            cy="25"
            r="20"
            fill="none"
            strokeWidth={type === SPINNER_TYPES.regular ? 4 : 1}
            className={classes.circle}
          />
        </svg>
```

src/components/Spinner/Spinner.visual.tsx
```diff
@@ -0,0 +1,9 @@
+import * as React from 'react';
+import { visualize, story, snap } from 'storybook-snapper';
+import { Spinner } from './';
+
+visualize('Spinner', () => {
+  story('simple', () => {
+    snap('default props', <Spinner isStatic />);
```
jonathanadler, line 7: 👍 !

src/components/Spinner/docs/examples.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const regular = `
+<Spinner />
+`;
+
+export const slim = `
+<ExampleWithContextProps mobile={true}>
+  <Spinner type="slim" />
+</ExampleWithContextProps>
+`;
```
jonathanadler, line 11: Again, mobile can be removed:
```suggestion
export const slim = `
<Spinner type="slim" />
`;
```

src/components/Spinner/Spinner.st.css
jonathanadler: btw, do we also want to center it vertically?

src/components/Spinner/Spinner.driver.ts
rustemk13: Removed mobile from all cases inside component

src/components/Spinner/Spinner.st.css
rustemk13: Removed

src/components/Spinner/Spinner.st.css
rustemk13: Fixed, now top and left are setted inside jsx, in accordance to diameter prop

src/components/Spinner/Spinner.tsx
rustemk13: Fixed

PR #372, opened 2020-08-13
feat: Modal

rustemk13:


src/components/Modal/Modal.e2e.ts
```diff
@@ -0,0 +1,32 @@
...
+  const dataHook = 'e2e-storybook-modal-wrapper';
+
+  beforeEach(() => browser.get(storyUrl));
+
+  it('should render', async () => {
+    const driver = modalTestkitFactory({ dataHook });
+    await waitForVisibilityOf(await driver.element(), 'Cannot find Modal');
+
+    expect(await driver.isModalShowed()).toBe(false);
+    await driver.clickOnOpenButton();
```
jonathanadler, line 23: so instead of using the driver here you would do:
```suggestion
    element(by.css('[data-hook=e2e-storybook-modal-open-btn]')).click();
```

jonathanadler: and import `element`, and `by` from `protractor`

src/components/Modal/Modal.spec.tsx
```diff
@@ -0,0 +1,59 @@
...
+          <Modal isOpen rootElement={document.body} focusTrap={false} />,
+          modalTestkitFactory,
+          {
+            dataHookPropName: 'data-hook',
+          },
+        ),
+      ).toBe(true);
+    });
+  });
+
```
jonathanadler, line 44: Maybe we can add a test to see that the focus trap works?

src/components/Modal/Modal.st.css
```diff
@@ -0,0 +1,91 @@
+:import {
+    -st-from: "../../common/formatters.st";
+    -st-named: color;
+}
```
jonathanadler, line 4: I think this isn't used, so it can be removed

src/components/Modal/Modal.st.css
```diff
@@ -0,0 +1,91 @@
...
+
+:vars {}
```
jonathanadler, line 6: This also can be removed

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+export class Modal extends React.Component<ModalProps, ModalState> {
+  static displayName = 'Modal';
+  static defaultProps: ModalDefaultProps = {
+    isOpen: false,
+    closeOnClickOutside: true,
+    minHeight: '150px',
+    maxHeight: '620px',
+    minWidth: 'auto',
+    maxWidth: '580px',
```
jonathanadler, line 17: So I think these props shouldn't be here, but actually in the Dialog component, since the `Modal` component shouldn't really care about style (except maybe for the background overlay color, etc)
In addition, instead of using javascript for these properties, consumers of the component can use css/stylable for this, just by passing a className, and giving a min/max height/width in css

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,159 @@
...
+    focusTrap: true,
+  };
+
+  el = null;
+  focusTrapInstance = null;
+
+  constructor(props: ModalProps) {
+    super(props);
+
+    this.el = document.createElement('div');
```
jonathanadler, line 27: This line won't work in SSR, because document doesn't exist.
So it should probably be moved to `componentDidMount`.
**However**, I don't think we need to create an extra `div`, maybe we can just use a ref?

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  componentDidUpdate(prevProps: ModalProps) {
+    if (this.props.isOpen !== prevProps.isOpen && this.focusTrapInstance) {
+      if (this.props.isOpen) {
+        try {
+          this.focusTrapInstance.activate();
+        } catch (err) {
+          console.error(err);
+        }
```
jonathanadler, line 51: Is the try & catch really needed?
Maybe, instead of catching and printing an error, we can just remove the try & catch.
And if there will be errors, then will find them quicker this way

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,159 @@
...
+      } else {
+        this.focusTrapInstance.deactivate();
+      }
+    }
+  }
+
+  onClose = () => {
```
jonathanadler, line 58: We prefer to use `_onClose` for internal methods

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,159 @@
...
+            <CloseIcon />
+          </button>
+          <div
+            className={`${classes.content} ${this.props.contentClassName ||
+              ''}`}
+            data-hook="tpa-modal-content"
+          >
+            {this.props.children}
+          </div>
+        </section>
```
jonathanadler, line 151: So all this part should actually be in the Dialog component.
So it would look like this:

```tsx
<section className={...} ...>
  {this.props.children}
</section>
```

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,159 @@
...
+            >
+              {ReactDOM.createPortal(this.renderModal(mobile, rtl), this.el)}
+            </div>
+          );
+        }}
+      </TPAComponentsConsumer>
+    );
+  }
+
+  renderModal(mobile, rtl) {
```
jonathanadler, line 98: Again, this should be `_renderModal`

src/components/Modal/Modal.visual.tsx
```diff
@@ -0,0 +1,9 @@
+import * as React from 'react';
+import { visualize, story, snap } from 'storybook-snapper';
+import { Modal } from './';
+
+visualize('Modal', () => {
+  story('simple', () => {
+    snap('default props', <Modal isOpen rootElement={document.body} />);
```
jonathanadler, line 7: We should probably add some content to the Modal, or else there won't be much to visualize 😁

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,159 @@
...
+            },
+            () => {
+              if (this.props.onClose) {
+                this.props.onClose();
+              }
+            },
+          );
+        }, 200);
+      },
+    );
```
jonathanadler, line 77: Don't forget that if we use `setTimeout` we need to be able to `clearTimeout` in order to avoid exceptions and memory leaks.
Also, I think instead of putting all this logic in the close method, I would do something like this:
```tsx
_onClose = () => {
  this.setState({ isCloseInProgress: true });
}
```

and add to `componentDidUpdate`:
```tsx
componentDidUpdate(prevProps: ModalProps) {
  ...
  if (!prevProps.isCloseInProgress && this.props.isCloseInProgress) {
    clearTimeout(this._closingTimeout);
    this._closingTimeout = setTimeout(() => {
      this.setState({isCloseInProgress: false});  
    }, 200);
  } else if (prevProps.isCloseInProgress && !this.props.isCloseInProgress) {
    if (this.props.onClose) {
      this.props.onClose();
    }
  }
}
```

and also always we add a `clearTimeout` to `componentWillUnMount`

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,159 @@
...
+    return (
+      <TPAComponentsConsumer>
+        {({ mobile, rtl }) => {
+          return (
+            <div
+              data-mobile={mobile}
+              data-rtl={rtl}
+              data-hook={this.props['data-hook']}
+            >
+              {ReactDOM.createPortal(this.renderModal(mobile, rtl), this.el)}
```
jonathanadler, line 90: I think `this.el` can be the `rootElement` in this case

src/components/Modal/types.ts
```diff
@@ -0,0 +1,30 @@
...
+  isOpen: boolean;
+  rootElement: HTMLElement;
+  onClose?(): void;
+  contentClassName?: string;
+  closeOnClickOutside?: boolean;
+  minHeight?: string;
+  maxHeight?: string;
+  minWidth?: string;
+  maxWidth?: string;
+  closeButtonRef?: RefObject<any>;
```
jonathanadler, line 14: So these should probably not be part of the `Modal`'s proptypes, but in the `Dialog`'s

src/components/Modal/Modal.tsx
rustemk13: Yep, focus trap plugin generate some extra errors is different cases, f.ex., if inside modal we haven't any focusable element. So I think try/catch is needed here.

jonathanadler: Oh I see.
So in that case, let's remove the `console.log`

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_372/

jonathanadler:
We managed to continue your work and merge this component.
We are now working on adding the Dialog component.
Thanks for the contribution!

PR #376, opened 2020-08-27
Added the capability to use icon inside a badge

GBT3101:


src/assets/icons/Camera.svg
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="12px" height="8px" viewBox="0 0 12 8" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <title>Camera</title>
+    <g id="09-Icons" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+        <g id="09-Icons-/-Video-/-Solid-/-S" transform="translate(-6.000000, -8.000000)" fill="#000000">
+            <g id="Camera" transform="translate(6.000000, 8.000000)">
+                <path d="M9,8 L0,8 L0,0 L9,0 L9,2.5 L12,1 L12,7 L9,5.5 L9,8 Z" id="Combined-Shape"></path>
+            </g>
+        </g>
+    </g>
```
jonathanadler, line 10: Please remove all the `id`'s

src/components/Badge/Badge.driver.ts
```diff
@@ -25,5 +26,6 @@ export const badgeDriverFactory = (base: UniDriver): BadgeDriver => {
     isPrimary: () => hasPriority(base, BADGE_PRIORITY.primary),
     isLight: () => hasPriority(base, BADGE_PRIORITY.light),
     isDefault: () => hasPriority(base, BADGE_PRIORITY.default),
+    getIcon: () => base.attr('.icon'),
```
jonathanadler, line 29: Using the class name `.icon` won't necessary work, since it depends on the hashing that's used on the compiled bundle.
So instead of classes please use a `data-` hook (like here: https://github.com/wix/wix-ui-tpa/pull/376/files#diff-da702d7b76d4fd862bd1a8e985077d14R12)

src/components/Badge/Badge.st.css
```diff
@@ -27,20 +27,39 @@
 }
 
 .root {
-    -st-states: priority(enum(default, light, primary));
+    -st-states: withIcon, priority(enum(default, light, primary));
```
jonathanadler, line 30: It seems you're not using the `withIcon` state.
So you can just remove it

src/components/Badge/Badge.tsx
```diff
@@ -22,14 +23,17 @@ class Badge extends React.Component<BadgeProps> {
...
-    const { priority, children, className } = this.props;
+    const { priority, children, className, icon } = this.props;
     return (
       <div
         data-priority={priority}
         className={st(classes.root, { priority }, className)}
         data-hook={this.props['data-hook']}
       >
-        {children}
+        <div style={{ display: 'flex' }}>
```
jonathanadler, line 33: We prefer to not use `style` attributes in our components.
So please use a class.
btw, is this `div` really needed, or is the root `div` enough?

jonathanadler (commented):
Great work!

Just left some comments.
Also, please add a visual test (with an icon) here: https://github.com/wix/wix-ui-tpa/blob/master/src/components/Badge/Badge.visual.tsx

Thanks!

src/components/Badge/Badge.driver.ts
```diff
@@ -13,10 +13,15 @@ const hasPriority = async (
...
 
+const hasIcon = async (base: UniDriver): Promise<boolean> => {
+  return !!base.$(`[data-hook=badge-icon] svg`);
+};
+
 export interface BadgeDriver extends BaseUniDriver {
   isPrimary(): Promise<boolean>;
   isLight(): Promise<boolean>;
   isDefault(): Promise<boolean>;
+  getIcon(): Promise<boolean>;
```
jonathanadler, line 24: so I think we should change this to `hasIcon` cause we're not really returning the icon, but a boolean

jonathanadler (commented):
Nice!
Just left a really really small comment.
I'm also waiting for the designers to approve
and after that we can merge 💪 
Great work
thanks!

src/components/Badge/Badge.driver.ts
GBT3101: You are right, changed it, I had to change the inner "hasIcon" function because it would shadow it, changed the inner one to "isIconExists"

jonathanadler (approved):
Great!
I just need to get a confirmation from the designers and then we can merge 💪 
Thanks for the contribution!

PR #377, opened 2020-08-30
Calendar event a11y

ItayTur:


src/components/Event/Event.driver.ts
```diff
@@ -9,6 +9,10 @@ export interface EventDriver extends BaseUniDriver {
   isTimeShown(): Promise<boolean>;
   isSelected(): Promise<boolean>;
   isFullday(): Promise<boolean>;
+  hasOnClick(): Promise<boolean>;
```
jonathanadler, line 12: Instead of `hasOnClick` I would create a method:
```ts
    isButton(): Promise<boolean>;
```

and then I would implement it like this:
```ts
isButton: async () => {
  return (await base._prop('tagName')).toLowerCase() === 'button';
}
```

src/components/Event/Event.spec.tsx
```diff
@@ -36,6 +36,28 @@ describe('Event', () => {
     expect(await driver.isTimeShown()).toBe(true);
   });
 
+  it('should have onClick', async () => {
+    const driver = createDriver(<Event onClick={() => {}} {...defaultProps} />);
+    expect(await driver.hasOnClick()).toBe(true);
+  });
```
jonathanadler, line 42: See my comments in the `driver` file

src/components/Event/Event.driver.ts
jonathanadler: btw, see this: https://github.com/wix-incubator/unidriver/blob/958388bd0de4f648972847451a0574fbd9d127d7/core/src/index.ts#L47

src/components/Event/Event.st.css
```diff
@@ -66,6 +83,10 @@
     height: 100%;
 }
 
+.root:disabled {
+    cursor: not-allowed;
+}
```
jonathanadler, line 88: I'm not sure we want this at the moment.
So please remove the cursor

src/components/Event/Event.st.css
```diff
@@ -42,7 +47,7 @@
 }
 
 .root {
-    -st-states: selected, fullday, rtl;
+    -st-states: selected, fullday, rtl, disabled;
```
jonathanadler, line 50: I don't see that the `disabled` state is used anywhere.
If it's not used it can be removed

src/components/Event/Event.tsx
```diff
@@ -26,40 +35,76 @@ export class Event extends React.Component<EventProps> {
...
   getDataAttributes() {
-    const { fullday, selected, showTime } = this.props;
+    const { fullday, selected, showTime, onClick, disabled } = this.props;
 
     return {
       [EVENT_DATA_KEYS.IsFullDay]: fullday,
       [EVENT_DATA_KEYS.IsSelected]: selected,
       [EVENT_DATA_KEYS.IsTimeShown]: showTime,
+      [EVENT_DATA_KEYS.OnClick]: Boolean(onClick),
+      [EVENT_DATA_KEYS.Disabled]: disabled,
```
jonathanadler, line 52: I think that in these cases we don't need data- attributes, because we can actually read them straight from the dom

src/components/Event/Event.tsx
```diff
@@ -26,40 +35,76 @@ export class Event extends React.Component<EventProps> {
...
+          classes.root,
+          { fullday, selected, rtl, disabled },
+          className,
+        )}
+      >
+        {timeComponent}
+        <Text className={classes.title}>{title}</Text>
+        <div className={classes.overlay} />
+      </div>
+    );
```
jonathanadler, line 93: For readability, I prefer moving these function outside of the `render` method, and just use them as instance methods:
```js
  _getEventProps: () => {
   ...
  },
  
  _getEvent: () => {
    ...
  }
```

src/components/Event/Event.visual.tsx
```diff
@@ -13,5 +13,9 @@ visualize('Event', () => {
     snap('timeShown', <Event showTime {...defaultProps} />);
     snap('fullday', <Event fullday {...defaultProps} />);
     snap('selected', <Event selected {...defaultProps} />);
+    snap('clickable', <Event onClick={() => {}} {...defaultProps} />);
+    snap('disabled', <Event onClick={() => {}} disabled {...defaultProps} />);
+    snap('aria-has-popup', <Event aria-has-popup {...defaultProps} />);
+    snap('aria-expanded', <Event aria-expanded {...defaultProps} />);
```
jonathanadler, line 19: Don't think these are needed

jonathanadler (commented):
Great work!

Left some small comments.

Thanks!

jonathanadler (approved):
Excellent!

PR #378, opened 2020-08-30
feat: add disabled and tooltip props to ColorPickerItem

alexgoncharwix:
https://zeroheight.com/695j65912/p/07a775-add-to-cart-on-gallery/b/19a09e

### DIFF
--- a/src/components/ColorPicker/ColorPickerItem/ColorPickerItem.driver.ts
+++ b/src/components/ColorPicker/ColorPickerItem/ColorPickerItem.driver.ts
@@ -2,8 +2,7 @@ import {
   BaseUniDriver,
   baseUniDriverFactory,
 } from 'wix-ui-test-utils/base-driver';
-import { StylableUnidriverUtil, UniDriver } from 'wix-ui-test-utils/unidriver';
-import * as style from './ColorPickerItem.st.css';
+import { UniDriver } from 'wix-ui-test-utils/unidriver';
 import { tooltipDriverFactory } from 'wix-ui-core/dist/src/components/tooltip/Tooltip.driver';
 import { radioButtonDriverFactory } from 'wix-ui-core/dist/src/components/radio-button/RadioButton.driver';
 import { Simulate } from 'react-dom/test-utils';
@@ -21,7 +20,6 @@ export interface ColorPickerItemDriver extends BaseUniDriver {
 export const colorPickerItemDriverFactory = (
   base: UniDriver,
 ): ColorPickerItemDriver => {
-  // const stylableDOMUtil = new StylableUnidriverUtil(style);
   const baseUniDriver = baseUniDriverFactory(base);
   const getTooltipDriver = async () => {
     const element = (await baseUniDriver.element()).querySelector(
@@ -39,27 +37,18 @@ export const colorPickerItemDriverFactory = (
   return {
     ...baseUniDriverFactory(base),
     isDisabled: async () => {
-      /*console.log(
-        "Elem",
-        await base.$(`[data-hook="${colorPickerItemDataHook}"]`).value()
-      );*/
       return (await getCoreRadioButtonDriver()).isDisabled();
       return (
         (await base
           .$(`[data-hook="${colorPickerItemDataHook}"]`)
           .attr('disabled')) !== null
       );
-      // return !!(await stylableDOMUtil.getStyleState(base, "isCrossedOut"));
     },
     getTooltipText: async () => {
       const tooltipDriver = await getTooltipDriver();
       return tooltipDriver.getTooltipText();
-
-      /*const tooltipDriver = await getTooltipDriver();
-      tooltipDriver.mouseEnter();
-      return tooltipDriver.getContentElement().innerHTML;*/
     },
-    isTooltipExist: async () => {
+    hasTooltip: async () => {
       const tooltipDriver = await getTooltipDriver();
       return tooltipDriver.exists();
     },
--- a/src/components/ColorPicker/ColorPickerItem/ColorPickerItem.driver.ts
+++ b/src/components/ColorPicker/ColorPickerItem/ColorPickerItem.driver.ts
@@ -2,7 +2,7 @@ import {
   BaseUniDriver,
   baseUniDriverFactory,
 } from 'wix-ui-test-utils/base-driver';
-import { UniDriver } from 'wix-ui-test-utils/unidriver';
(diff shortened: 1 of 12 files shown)


src/components/ColorPicker/ColorPickerItem/ColorPickerItem.driver.ts
```diff
@@ -0,0 +1,67 @@
...
+import { Simulate } from 'react-dom/test-utils';
+import {
+  colorPickerItemDataHook,
+  colorPickerItemTooltipDataHook,
+} from '../dataHooks';
+
+export interface ColorPickerItemDriver extends BaseUniDriver {
+  isDisabled(): Promise<boolean>;
+  getTooltipText(): Promise<string>;
+  isTooltipExist(): Promise<boolean>;
```
jonathanadler, line 18: Maybe?
```suggestion
  hasTooltip(): Promise<boolean>;
```

src/components/ColorPicker/ColorPickerItem/ColorPickerItem.driver.ts
```diff
@@ -0,0 +1,67 @@
...
+      return tooltipDriver.getTooltipText();
+
+      /*const tooltipDriver = await getTooltipDriver();
+      tooltipDriver.mouseEnter();
+      return tooltipDriver.getContentElement().innerHTML;*/
+    },
+    isTooltipExist: async () => {
+      const tooltipDriver = await getTooltipDriver();
+      return tooltipDriver.exists();
+    },
```
jonathanadler, line 65: Probably some leftover commented code

src/components/ColorPicker/ColorPickerItem/ColorPickerItem.st.css
```diff
@@ -57,5 +57,21 @@
...
     border-radius: value(colorPickerBorderRadius);
     border: 1px solid #C9C9C9;
+    position: relative;
+    overflow: hidden;
 }
 
+.root:isCrossedOut .radioInner::before {
+    content: '';
+    position: absolute;
+    border-top: 1px solid white;
```
jonathanadler, line 67: Note that the `border-top` color here, should be the same as the as the border there ☝️  (#c9c9c9)

jonathanadler (commented):
Wow, awesome work!

I left some teeny tiny comments.

Also, the `MobileTooltip` you wrote is very nice!
I just couldn't find any spec for this (in zeroheight or something like that)

Could you add a `mobile` example to the storybook, so that we can play with it?
(you can wrap your example with the `TPAComponentsProvider` - I believe you can find existing examples in other stories)

Nice work!

src/components/ColorPicker/ColorPickerItem/ColorPickerItem.tsx
```diff
@@ -2,11 +2,19 @@ import * as React from 'react';
...
+  colorPickerItemTooltipDataHook,
+} from '../dataHooks';
+import { Tooltip } from '../../Tooltip';
+import { MobileTooltip } from './MobileTooltip';
+import { TPAComponentsConsumer } from '../../TPAComponentsConfig';
 
 export interface ColorPickerItemProps extends RadioButtonProps {
   className?: string;
   key?: number;
+  disabled?: boolean;
```
jonathanadler, line 16: After talking to @johannaasseraf I understand that the item isn't actually disabled.
This means we should probably change the name of the prop to `unavailable` because we still want the item to be selectable.
We might add a `disabled` prop in the future, which will really set the input to be `disabled` and the item won't be selectable.

You can then change the stylable state to `unavailable` as well

alexgoncharwix:
Here is zeroheight requirements for tooltip in mobile mode: https://zeroheight.com/695j65912/p/07a775-add-to-cart-on-gallery/b/19a09e ("On mobile, show tooltip while holding touch on the color. (this is approved by Neil from the a11y team, but lets check how it feels). When releasing give a 1 sec delay before hiding the tooltip.")
Is there any centralised store of requirements for all component? I want to ask our UX to add these rules to common store.

src/components/ColorPicker/ColorPickerItem/ColorPickerItem.tsx
alexgoncharwix: After the very last discussion we decided that we currently need `disabled` state (cannot be selected, but with tooltip). In future we plan to have addition `unavailable` state (or we can call it `crossedOut`) which will be selectable but with crossed out style.

alexgoncharwix: @johannaasseraf Please review.

johannaasseraf: I just didn't really want the disabled state to be connected to the crossed-out/unavailable design. If you want it can be a separate prop that is turned on for our case. Makes sense?

alexgoncharwix: Absolutely. So to recap. Two props: `disabled` and `isCrossedOut`. All good?

johannaasseraf: Makes sense to me.. What do you say @jonathanadler ?

alexgoncharwix: `unavailable ` or `isCrossedOut`  -- both variants are good in my opinion. Don't know which one to pick out.

johannaasseraf: I'm ok with crossedout, sounds less specific for a certain vertical

jonathanadler: yeah lets go with the crossed-out one 👍

jonathanadler:
> Here is zeroheight requirements for tooltip in mobile mode: https://zeroheight.com/695j65912/p/07a775-add-to-cart-on-gallery/b/19a09e ("On mobile, show tooltip while holding touch on the color. (this is approved by Neil from the a11y team, but lets check how it feels). When releasing give a 1 sec delay before hiding the tooltip.")
> Is there any centralised store of requirements for all component? I want to ask our UX to add these rules to common store.

What do you mean by store of requirements?
Desing-wise we use the zeroheight link

jonathanadler:
Excellent!
I'll publish a new version tomorrow

PR #379, opened 2020-08-30
fix(Checkbox):  Add transition to checkbox label color on switch to d…

GuyPerkal:
…isabled.

### DIFF
--- a/src/components/Checkbox/Checkbox.st.css
+++ b/src/components/Checkbox/Checkbox.st.css
@@ -141,6 +141,7 @@
 .label {
     display: inline-block;
     color: value(CurrentTextColor);
+    transition: color 0.2s linear;
 }
 
 .label.suffixed {


jonathanadler:
Hey Guy,
Thanks for the contribution!

I need to first check this change is approved by our designers.
Not sure if it's something that was defined for the library.

jonathanadler:
Hey @GuyPerkal 
I checked with the designers and we got an approval 👌 
Only thing missing is adding the same transition (same timing) to the checkmark as well.
Thanks!

GuyPerkal:
Well, it's actually not the same thing - the checkmark is from wix-ui-core's checkbox, and it can't be transitioned in the CSS since it's only added to the DOM when the checkbox is checked.

jonathanadler:
> Well, it's actually not the same thing - the checkmark is from wix-ui-core's checkbox, and it can't be transitioned in the CSS since it's only added to the DOM when the checkbox is checked.

I think it's actually on our side, here: https://github.com/wix/wix-ui-tpa/blob/04b4b3c9f249be49c5bf2687d51b0b9701226948/src/components/Checkbox/Checkbox.tsx#L65
We can use classes instead of conditionally rendering the icons.
Do you think you have time to add this?
"No" is also an acceptable answer ☺️

GuyPerkal:
I missed _renderIcon, but I'm still not sure this is possible using only CSS - since there are 2 icons, indeterminate and checked, we can't keep both on the DOM and animate the opacity since then they'll both take up space - if we animate the indeterminate state icon's opacity, it will show up under the transparent checked icon.

jonathanadler:
> I missed _renderIcon, but I'm still not sure this is possible using only CSS - since there are 2 icons, indeterminate and checked, we can't keep both on the DOM and animate the opacity since then they'll both take up space - if we animate the indeterminate state icon's opacity, it will show up under the transparent checked icon.

I see, you're right 🤔 

So I'll add this to our backlog, and we'll need to think it through a bit more.

Thanks!

PR #380, opened 2020-08-30
Picker - aria-label props & role=status

ItayTur:


src/components/Picker/Picker.driver.ts
```diff
@@ -10,6 +10,8 @@ export interface PickerDriver extends BaseUniDriver {
   clickOnPrev(): Promise<void>;
   hasDisablePrev(): Promise<boolean>;
   hasDisableNext(): Promise<boolean>;
+  hasNextAriaLabel(nextAriaLabel): Promise<boolean>;
+  hasPrevAriaLabel(prevAriaLabel): Promise<boolean>;
```
jonathanadler, line 14: I think instead of having a boolean method, we can just return the ariaLabel:
```suggestion
  getNextAriaLabel(nextAriaLabel): Promise<string | undefined>;
  getPrevAriaLabel(prevAriaLabel): Promise<string | undefined>;
```
and then just return 
```js
return base.$(`[data-hook=my-next-button]`).attr('aria-label')
```

src/components/Picker/Picker.spec.tsx
```diff
@@ -7,6 +7,7 @@ import { pickerDriverFactory } from './Picker.driver';
 import { Picker } from './';
 import { pickerTestkitFactory } from '../../testkit';
 import { pickerTestkitFactory as enzymePickerTestkitFactory } from '../../testkit/enzyme';
+import { Preview } from 'wix-ui-icons-common';
```
jonathanadler, line 10: I think line can be removed

src/components/Picker/Picker.tsx
```diff
@@ -13,26 +13,40 @@ export interface PickerProps extends TPAComponentProps {
...
+      nextAriaLabel,
+      prevAriaLabel,
+    } = this.props;
 
     return {
       [PICKER_DATA_PROPS.NextDisabled]: nextDisabled,
       [PICKER_DATA_PROPS.PrevDisabled]: prevDisabled,
+      [PICKER_DATA_PROPS.PrevDisabled]: prevDisabled,
+      [PICKER_DATA_PROPS.NextAriaLabel]: nextAriaLabel,
+      [PICKER_DATA_PROPS.PrevAriaLabel]: prevAriaLabel,
```
jonathanadler, line 49: These aren't needed

src/components/Picker/Picker.tsx
```diff
@@ -43,14 +57,18 @@ export class Picker extends React.Component<PickerProps> {
...
       className,
+      ...rest
     } = this.props;
 
     return (
       <div
         data-hook={this.props['data-hook']}
         className={st(classes.root, {}, className)}
         {...this.getDataAttributes()}
+        {...rest}
```
jonathanadler, line 71: Do we need to pass all the `rest` to the `div`?

src/components/Picker/Picker.tsx
```diff
@@ -59,15 +77,19 @@ export class Picker extends React.Component<PickerProps> {
           disabled={prevDisabled}
           data-hook={PICKER_DATA_HOOKS.Prev}
           title={PICKER_DATA_HOOKS.Prev}
+          aria-label={prevAriaLabel}
         />
-        <Text className={classes.value}>{value}</Text>
+        <Text className={classes.value}>
+          <span role="status">{value}</span>
+        </Text>
```
jonathanadler, line 84: I would actually prefer we create a `role` prop in the `Text` component, so we won't have unneeded spans and divs

src/components/Picker/Picker.visual.tsx
```diff
@@ -13,5 +13,11 @@ visualize('Picker', () => {
     snap('default', () => <Picker {...defaultProps} />);
     snap('prev disabled', () => <Picker prevDisabled {...defaultProps} />);
     snap('next disabled', () => <Picker nextDisabled {...defaultProps} />);
+    snap('next aria label', () => (
+      <Picker nextAriaLabel="next" {...defaultProps} />
+    ));
+    snap('prev aria label', () => (
+      <Picker prevAriaLabel="prev" {...defaultProps} />
+    ));
```
jonathanadler, line 21: These aren't needed actually, cause they don't change anything visually.
If anything I would add these props to all the other visual tests by default

src/components/Picker/dataHooks.ts
```diff
@@ -6,4 +6,6 @@ export enum PICKER_DATA_HOOKS {
 export enum PICKER_DATA_PROPS {
   NextDisabled = 'data-disabled-next',
   PrevDisabled = 'data-disabled-prev',
+  NextAriaLabel = 'data-next-aria-label',
+  PrevAriaLabel = 'data-prev-aria-label',
```
jonathanadler, line 10: Again, I don't think this is needed

jonathanadler (commented):
Great work 👍 
I've added some comments but looks great in general

src/components/Picker/Picker.tsx
ItayTur: i just thought that it will allow consumers add props that they need, but thinking about it, typescript will not allow it anyway.

jonathanadler: that's cause the IconButton extends ButtonNext, so the ...rest is for the rest of ButtonNext props.
In this case it's not needed

ItayTur: the aria-label prop, how's holding it ?

jonathanadler (approved):
Great work!

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_380/

PR #385, opened 2020-09-02
Better look for badge with icon, with more examples

GBT3101:


src/components/Badge/Badge.st.css
```diff
@@ -54,6 +54,11 @@
     width: 2px;
 }
 
+.icon::before {
+    content: '';
+    margin: 0 -1px;
+}
+
```
jonathanadler, line 61: 🤔  I looked at the storybook, but it doesn't seem to work
Also, if for instance we're in `ltr`, the `margin-right: -1px` won't move the icon left, but move what's after the icon to the left.

I think we should just replace the `root`'s padding with a `before` and `after` instead.

jonathanadler (commented):
Left a comment
Great examples!

src/components/Badge/Badge.st.css
GBT3101: Are you sure? this is why I did it on both sides

jonathanadler (approved):

PR #387, opened 2020-09-03
FloatingDropdown: import missing styles from Dropdown

mariusju:
apparently these need to be imported explicitly

### DIFF
--- a/src/components/FloatingDropdown/FloatingDropdown.tsx
+++ b/src/components/FloatingDropdown/FloatingDropdown.tsx
@@ -4,6 +4,13 @@ import { Dropdown as CoreDropdown } from 'wix-ui-core/dropdown';
 
 import { DropdownOption } from '../Dropdown/DropdownOption';
 import { TPAComponentsConsumer } from '../TPAComponentsConfig';
+
+/* TODO
+ * The import below is a workaround. FloatingDropdown and Dropdown
+ * components should be refactored according to this thread
+ * https://github.com/wix/wix-ui-tpa/pull/387 (or merged into a single component)
+ */
+import '../Dropdown/Dropdown.st.css';
 import { st, classes } from './FloatingDropdown.st.css';
 import { FloatingDropdownBase } from './FloatingDropdownBase';
 import { DATA_HOOKS } from './constants';
@@ -72,7 +79,7 @@ export class FloatingDropdown extends React.Component<FloatingDropdownProps> {
   _generateCoreOptions() {
     const { options } = this.props;
 
-    const renderableOptions = options.map(option => ({
+    const renderableOptions = options.map((option) => ({
       ...option,
       render: () => <DropdownOption {...option} />,
     }));


src/components/FloatingDropdown/FloatingDropdown.tsx
```diff
@@ -4,6 +4,7 @@ import { Dropdown as CoreDropdown } from 'wix-ui-core/dropdown';
 
 import { DropdownOption } from '../Dropdown/DropdownOption';
 import { TPAComponentsConsumer } from '../TPAComponentsConfig';
+import '../Dropdown/Dropdown.st.css';
```
jonathanadler, line 7: Why is that 🤔 ?
It doesn't make much sense, since Stylable works with scoped classNames

mariusju: Solely importing Dropdown.st.css in the FloatingDropdown.st.css file does not include it in the css bundle. 
If a site has a regular dropdown somewhere then these styles are bundled, but if not, then FloatingDropdown is rendered with empty classes. 
Importing it explicitly assures that these styles are always bundled.

@barak007 has more details on this issue, he helped me to debug it.

shlomitc: It doesn't seem like the right solution and something is a bit odd.
In this component, we import `Dropdown` directly from `wix-ui-core`, that's kind of fine although we have a `Dropdown` component already wrapped with styles. not sure what was the reason for it though.

In that case, I would expect to see some `-st-extends: Dropdown` inside the `FloatingDropdown.st.css`, instead of importing the stylesheet in the `tsx` file.

My guess that the missing extend is what's causing the issue.

jonathanadler: @shlomitc I think that's cause it's not extending the component, it's just sharing the stylesheet

jonathanadler: But yeah, the better solution would be refactoring the structure of the two components.

jonathanadler:
@mariusju You'll need to sign your commits for the PR build to run.
Please follow the guide here: https://github.com/wix/wix-style-react/blob/master/docs/contribution/CREATE_PR.md

src/components/FloatingDropdown/FloatingDropdown.tsx
barak007: The problem here is that we are using the `Dropdown` component from `wix-ui-core` but we want to reuse the styling from the `Dropdown` from `wix-ui-tpa` (and not use the component as the root). The way it works is by importing a class called `dropdown` form the `wix-ui-tpa Dropdown` to the `FloatingDropdown`. That will cause the same className to be applied to both (when used in the same build which has it's own pros and cons). But when you build a project that **dose not** use the `wix-ui-tpa Dropdown` Stylable knows that the it's stylesheet is not in use in `js` files so no one can apply classes from that file to the DOM then it ***exclude it safely** from the output css. so what we did is to tell the bundler that we are using this stylesheet in this component so it dose not optimized it out.

The current way dose not duplicate css between the two component instead the css is sheared.

*This is a semi bug semi unexpected behavior which we have a solution in long waiting related [PR](https://github.com/wix/stylable/pull/906)  

It's kind of a long explanation of what we did to solve an issue. but it's not explaining how we got here.

I don't think that the structure of the components and the css is correct here and this is why we ask these questions the idea that we use css from a `component` and not use the component is not right. 
I think that if there was a separation between structural css and styling then it would be more acceptable to reuse styling.

like `dropdown-coolstyle.st.css` provided from the core component that both component can use.

I can add hours of reading time here if more explanation required I will add it.

shlomitc: So let's start with this quick hack, but ideally, we should refactor the entire Dropdown styles and reuse them in `Dropdown` and `FloatingDropdown`.

Let's add a big **TODO** here with a link to this thread and decide that the next time we are touching this component we'll need a refactor

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_387/

jonathanadler (approved):

PR #405, opened 2020-09-10
feat(TextArea): add 'onBlur' property

GuyPerkal:


src/components/TextArea/TextArea.tsx
```diff
@@ -25,6 +25,7 @@ export interface TextAreaProps extends TPAComponentProps {
   theme?: TextAreaTheme;
   errorDescription?: string;
   onChange(event: React.ChangeEvent<HTMLTextAreaElement>): void;
+  onBlur(): void;
```
jonathanadler, line 28:
```suggestion
  onBlur()?: void;
```

jonathanadler (commented):
Nice 👍 

We just need to add a test, for future proofing

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_405/

src/components/TextArea/TextArea.driver.ts
```diff
@@ -2,6 +2,7 @@ import {
   BaseUniDriver,
   baseUniDriverFactory,
 } from 'wix-ui-test-utils/base-driver';
+import { hasAttribute } from 'wix-ui-test-utils/protractor';
```
jonathanadler, line 5: This was probably an accident by webstorm

src/components/TextArea/TextArea.driver.ts
```diff
@@ -75,5 +77,8 @@ export const textAreaDriverFactory = (base: UniDriver): TextAreaDriver => {
     async success() {
       return isSuccess(base);
     },
+    async hasAttribute(attr: string) {
+      return (await textArea(base).attr(attr)).length > 0;
+    },
```
jonathanadler, line 82: I don't think we need this method.
I think you're not using it, and it's very generic, and in any case consumers using the driver can do:
```js
(await driver.element()).getAttribute('my attribute');
```

Also, just for future reference, this could have been written like this:
```js
    async hasAttribute(attr: string) {
      return !!(await base.attr(attr));
    },
```

jonathanadler (commented):
Great,
just some small comments left

src/components/TextArea/TextArea.driver.ts
GuyPerkal: Yeah sorry about this, I didn't meant to commit it.

GuyPerkal:
OK, the test is fixed.

src/components/TextArea/TextArea.driver.ts
```diff
@@ -2,6 +2,7 @@ import {
   BaseUniDriver,
   baseUniDriverFactory,
 } from 'wix-ui-test-utils/base-driver';
+import { Simulate } from 'react-dom/test-utils';
```
jonathanadler, line 5: I'm assuming what I suggested didn't work?

This isn't fit for unidriver's though.
The unidriver should be environment agnostic, so we shouldn't use tools specific for react-dom/jsdom.

I guess we'll have to move this logic to the test itself.
So you'll need to expose the `textarea` function and move the simulation to the test itself.

src/components/TextArea/TextArea.driver.ts
```diff
@@ -75,5 +77,9 @@ export const textAreaDriverFactory = (base: UniDriver): TextAreaDriver => {
     async success() {
       return isSuccess(base);
     },
+    async blur() {
+      const element = await textArea(base).getNative();
+      Simulate.blur(element);
+    },
```
jonathanadler, line 83:
```suggestion
    async getTextareaElement() {
      return textArea(base).getNative();
    },
```

src/components/TextArea/TextArea.spec.tsx
```diff
@@ -141,4 +141,19 @@ describe('TextArea', () => {
...
+    const onBlur = jest.fn();
+    const driver = createDriver(
+      <TextArea
+        ariaLabel="test"
+        value="Test"
+        onChange={onChange}
+        onBlur={onBlur}
+      />,
+    );
+    await driver.blur();
```
jonathanadler, line 156: So in this case you'll have to write

```suggestion
    const textarea = await driver.getTextareaElement();
    Simulate.blur(textarea);
```

src/components/TextArea/TextArea.driver.ts
GuyPerkal: OK, I thought this might be acceptable since I saw other drivers use Simulate (the Checkbox driver, for example).

jonathanadler (approved):

src/components/TextArea/TextArea.driver.ts
jonathanadler: You're right.
We really need to fix that.

Issue #401, opened 2020-09-10
ActionsMenuLayout A11Y fixes


jonathanadler:
## ActionsMenu Item
See ws example here: https://www.w3.org/TR/2016/WD-wai-aria-practices-1.1-20161214/examples/menu-button/menu-button-1/menu-button-1.html

* should expose a `focus` method - internally will have a ref to the button and invoke the buttons focus (should it also be focusable if not clickable?)
* onClick should also fire when focused on an item and pressed `space` or `enter`

## ActionsMenu
* `aria-labelledby` and `aria-label` props
* navigation with keyboard arrows

jonathanadler:
updated the spec

bordi:
https://github.com/wix/wix-ui-tpa/pull/406

PR #406, opened 2020-09-10
<ActionsMenuLayout /> - add keyboard arrow navigation 

bordi:
This PR implements this issue: https://github.com/wix/wix-ui-tpa/issues/401
As far as this component is meant to be in `Popover` element we can focus `item` on `cdm` phase

### DIFF
--- a/src/components/ActionsMenuLayout/ActionsMenuLayout.tsx
+++ b/src/components/ActionsMenuLayout/ActionsMenuLayout.tsx
@@ -12,6 +12,7 @@ export interface ActionsMenuLayoutProps extends TPAComponentProps {
   focusedIndex?: number;
   /** a11y */
   'aria-labeledby'?: string;
+  'aria-label'?: string;
 }
 
 /** ActionsMenuLayout */
@@ -20,31 +21,25 @@ export class ActionsMenuLayout extends React.Component<ActionsMenuLayoutProps> {
   static Divider = ActionsMenuLayoutDivider;
 
   static displayName = 'ActionsMenuLayout';
+
   state = {
     focusedIdx: 0,
   };
 
-  liRefs = [];
+  liRefs: {
+    [key: number]: {
+      liRef: React.RefObject<HTMLLIElement>;
+    };
+  } = {};
 
   componentDidMount() {
     const { focusedIndex } = this.props;
 
     if (focusedIndex) {
-      this.focusItem(focusedIndex);
+      this._focusItem(focusedIndex);
     } else {
-      this.focusItem(this.state.focusedIdx);
+      this._focusItem(this.state.focusedIdx);
     }
-  }
-
-  focusItem(idx: number) {
-    this.setState(
-      {
-        focusedIdx: idx,
-      },
-      () => {
-        this.liRefs[idx].liRef?.current.focus();
-      },
-    );
   }
 
   getDataAttributes(mobile: boolean) {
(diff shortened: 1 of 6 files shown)


src/components/ActionsMenuLayout/ActionsMenuLayout.tsx
```diff
@@ -5,9 +5,13 @@ import { ActionsMenuLayoutItem, Alignment } from './Item/ActionsMenuLayoutItem';
 import { ACTIONS_MENU_DATA_KEYS } from './dataHooks';
 import { ActionsMenuLayoutDivider } from './ActionsMenuLayoutDivider/ActionsMenuLayoutDivider';
 import { TPAComponentProps } from '../../types';
+import { KEYS } from '../../common/keyCodes';
 
 export interface ActionsMenuLayoutProps extends TPAComponentProps {
   alignment?: Alignment;
+  focusedIndex?: number;
```
bordi, line 12: I've added a declarative option to set a focused item instead of exposing ref outside because of `React.clone` added a mess with refs

src/components/ActionsMenuLayout/ActionsMenuLayout.tsx
```diff
@@ -16,30 +20,102 @@ export class ActionsMenuLayout extends React.Component<ActionsMenuLayoutProps> {
...
+    if (e.key === KEYS.ArrowUp || e.key === KEYS.ArrowUpIE) {
+      e.preventDefault(); // remove scroll on space
+      this.focusItem(this.getNextItemIdx(-1));
+    }
+  };
+
+  addRef(ref, index: number) {
+    if (ref) {
+      if (
+        this.liRefs.find(_ref => {
```
bordi: here is a bit of "magic" bc the list of refs might be not normalized

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_406/

src/components/ActionsMenuLayout/ActionsMenuLayout.tsx
```diff
@@ -16,30 +20,98 @@ export class ActionsMenuLayout extends React.Component<ActionsMenuLayoutProps> {
...
     return (
       <TPAComponentsConsumer>
         {({ mobile }) => (
           <ul
+            onKeyDown={this._onKeyDown}
             className={st(classes.root, { mobile }, className)}
             {...this.getDataAttributes(mobile)}
             data-hook={this.props['data-hook']}
             role="menu"
-            tabIndex={-1}
```
jonathanadler, line 36: I think we should leave the `tabindex="-1"`
and also add an onfocus handler to the `ul`,
which will be something like this:
```js
_onFocus: () => {
    const { focusedIdx } = this.state;
    const { focusedIndex } = this.props;

    this.focusItem(typeof focusedIndex !== 'undefined' ? focusedIndex : focusedIdx);
}
```

src/components/ActionsMenuLayout/ActionsMenuLayout.tsx
jonathanadler: I understand 🤔 
This is implementation is fine, but question:
doesn't React handle this automatically?
I mean, React calls the ref only on when it mounts, and unmounts (with `null`)
So this might work out by itself, no?

src/components/ActionsMenuLayout/ActionsMenuLayout.tsx
```diff
@@ -16,30 +20,98 @@ export class ActionsMenuLayout extends React.Component<ActionsMenuLayoutProps> {
...
+        this.liRefs.find(_ref => {
+          return _ref?.props.content === ref.props.content;
+        })
+      ) {
+        return;
+      }
+
+      this.liRefs[
+        index === 0 || this.liRefs[index - 1] ? index : index - 1
+      ] = ref;
```
jonathanadler, line 90: Not sure I follow this part.
Thinking out loud: would it be easier to use a map instead of an array?
So that `this.liRefs` will look like this:
```js
{
  '0': firstRef,
  '1': secondRef,
  ...
}
```
and then just have 
```js
this.liRefs[index] = ref
```

jonathanadler (commented):
Nice! Great code!

I left some really minor comments.
But two thing I would really like:
1. It's not related to the PR itself, but can you change all the "private" methods to start with `_` 🙏 
It's just cause the autodocs we use differentiate between methods with and without the `_` prefix
2. We have to have some tests for this. I think we can add a test like this:
```js
driver.focus();
sendKeys('space');
expect(firstOnClick).toBeCalled();
sendKeys('down');
sendKeys('space');
expect(secondOnClick).toBeCalled();
// and also test the loop
sendKeys('down');
sendKeys('space');
expect(firstOnClick).toBeCalled(); // looped to beginning
```
wdyt?
and maybe also a test for `'up'`

src/components/ActionsMenuLayout/ActionsMenuLayout.tsx
bordi: I guess I can play around with it. But the `children` of `Layout` component can be a `divider` also. 
So, the map will be something like this:
```
{
  '0': firstRef,
  '1': secondRef,
  '3': secondRef,
  ...
}
```

and we should convert maybe to normalized array with `Object.values`. Wdyt?

src/components/ActionsMenuLayout/ActionsMenuLayout.tsx
bordi: Yeah, nice catch. Will add!

src/components/ActionsMenuLayout/ActionsMenuLayout.tsx
bordi: But maybe with map code will be easier to read. will try to change it and play around with

jonathanadler (approved):
Awesome!
I'll approve, but we just need the test and I think that's it
Thanks!

src/components/ActionsMenuLayout/ActionsMenuLayout.tsx
jonathanadler: yeah, I forgot about the divider 🤦

mbutsykin:
🥇

PR #409, opened 2020-09-13
Feat/dialog

sivanharel:


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_409/

src/components/Dialog/Dialog.st.css
```diff
@@ -0,0 +1,72 @@
...
+:vars {
+    CurrentCloseButtonColor: color(fallback(value(CloseButtonColor), value(DefaultCloseButtonColor)));
+}
+
+.root {
+    -st-states: mobile;
+}
+
+.root:mobile {
+}
```
jonathanadler, line 31: This can be removed I guess

src/components/Dialog/Dialog.st.css
```diff
@@ -0,0 +1,72 @@
...
+
+.closeButtonWrapper {
+    position: absolute;
+    top: 20px;
+    right: 20px;
+}
+
+:global([dir="rtl"]) .closeButtonWrapper,
+:global(.rtl) .closeButtonWrapper {
```
jonathanadler, line 55: So actually for `rtl` cases, we use the `rtl` prop that's passed through the context.
Just like the `mobile` prop.
and then we can use it as a Stylable state.
You can see an example here: https://github.com/wix/wix-ui-tpa/blob/291f717d564cffd107f9d194ba9feca3f6d35c9b/src/components/Pagination/Pagination.tsx#L26

src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,133 @@
...
+  /** Defines a string value that labels the dialog element. Optional. */
+  'aria-label'?: string;
+  /** Identifies the element that labels the dialog element. Optional. */
+  'aria-labelledby'?: string;
+  /** Gives the dialog an accessible description by referring to the dialog content that describes the primary message or purpose of the dialog. Optional. */
+  'aria-describedby'?: string;
+  /** Defines a string value that labels the close button element. Optional. */
+  'close-button-aria-label'?: string;
+  /** Identifies the element that labels the close button element. Optional. */
+  'close-button-aria-labelledby'?: string;
```
jonathanadler, line 32: I'm not sure about the naming here.
Usually we use `aria-label` and such, in order to align with native html.
Since `aria-label` and `data-` attributes exist in html and React, we prefer to use them as is.
the close-button-aria-label, however, isn't a native html attribute.
So in this case I think we can wither have the prop named: `closeButtonAriaLabel`
or have a prop: `closeButtonAria` prop that accepts an object:
```ts
{
  label: string,
  labelledby: string,
}
```

wdyt?

src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+  _addEscEventListener() {
+    if (this.props.isOpen) {
+      document.addEventListener('keyup', this._onEscKeyup);
+    }
+  }
+
+  _removeEscEventListener() {
+    document.removeEventListener('keyup', this._onEscKeyup);
+  }
```
jonathanadler, line 56: I'm thinking maybe this logic should be in the Modal, no? wdyt?
Also, I guess we'll need this to **not** work in mobile.
In these cases, you can use the context like this:
```jsx
class MyClass extends React.Component {
  static contextType = MyContext;
  render() {
    let value = this.context;
    /* render something based on the value */
  }
}
```

src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,133 @@
...
+          <div
+            className={st(classes.root, { mobile }, className)}
+            data-hook={this.props['data-hook']}
+            data-mobile={mobile}
+            role="dialog"
+            aria-modal="true"
+            aria-label={ariaLabel}
+            aria-labelledby={ariaLabelledBy}
+            aria-describedby={ariaDescribedBy}
+          >
```
jonathanadler, line 107: This should probably be inside the Modal.
The `role="dialog"` should be the Modal's child, and not the other way around.

So I would do something like this:
```tsx
<Modal ...>
  <div role="dialog" ...> // or use <dialog .../> if you can
    ...
  </div>
</Modal>
```

src/components/Modal/Modal.driver.ts
```diff
@@ -6,22 +6,23 @@ import { UniDriver } from 'wix-ui-test-utils/unidriver';
 import { MODAL_DATA_HOOKS } from './dataHooks';
 
 export interface ModalDriver extends BaseUniDriver {
-  getModalStage(): UniDriver;
+  getModalContent(): UniDriver;
   isModalShowed(): Promise<boolean>;
```
jonathanadler, line 10: I know it's not part of your code, but maybe you can change `isModalShowed` to `isShown` or just `isOpen` 😬 ?

jonathanadler (commented):
Great work!!

I left some comments, but it looks really good.

Please also ask @michalvardi to review.

Thanks 🙏

src/components/Dialog/Dialog.tsx
sivanharel: I also wasn't sure what will be a good naming for it.I change for "closeButtonAriaLabel"

src/components/Modal/Modal.driver.ts
sivanharel: sure

src/components/Dialog/Dialog.tsx
sivanharel: yes, maybe I should move it for the modal.
But then should it added only when "closeOnClickOutside" is true? This is the reason that I put it in the dialog cause it seems like an alternative for the X button.
WDYT?

jonathanadler: 🤔  good question
I guess the safest thing to do is add a `closeOnEsc` prop.
How do WSR modal's work? do they always close on esc?

sivanharel: Yes sounds like a good idea.
In WSR the Modal uses ReactModal which has a prop "shouldCloseOnEsc" that by default true.
I think I will do the same in our Modal :)

jonathanadler: Hmmm, the only thing is that usually it's best not to have `true` as a default value.
So maybe we should use a prop like `disableEsc` or `withoutEsc`?
wdyt?

jonathanadler (approved):

src/components/Dialog/Dialog.st.css
```diff
@@ -0,0 +1,69 @@
...
+    top: 50%;
+    left: 50%;
+    transform: translate(-50%, -30%);
+
+    background: #fff;
+}
+
+.root:mobile .contentWrapper {
+    height: 100%;
+    transform: translate(-50%, -50%);
```
jonathanadler, line 47: I didn't notice this before, but maybe it's better to not have translations in mobile?
```suggestion
    top: 0;
    left: 0;
    width: 100%;
    height: 100%;
    transform: none;
```

jonathanadler (commented):
One tiny comment I missed before

PR #414, opened 2020-09-29
Stylable Partial mixins

jonathanadler:
Doesn't work out of the box yet.
Need to play with it a bit more

### DIFF
--- a/src/components/Card/Card.st.css
+++ b/src/components/Card/Card.st.css
@@ -50,7 +50,7 @@
     width: 100%;
     height: 100%;
     background-color: value(cardBGColor);
-    -st-partial-mixin: withBorder(
+    -st-mixin: withBorder(
             BorderColor value(borderColor),
             BorderWidth value(borderWidth)
     );
@@ -137,14 +137,14 @@
 .root > .infoContainer {
     box-sizing: border-box;
     background-color: value(infoBGColor);
-    -st-partial-mixin: withDynamicPadding(
+    -st-mixin: withDynamicPadding(
             PaddingSize value(infoPadding),
             BorderWidth value(borderWidth)
     );
 }
 
 .root:stacked > .infoContainer {
-    -st-partial-mixin: withDynamicPadding(
+    -st-mixin: withDynamicPadding(
             PaddingSize value(stackedInfoPadding),
             BorderWidth value(borderWidth)
     ), withDynamicSidePadding(
@@ -154,7 +154,7 @@
 }
 
 .root:mobile > .infoContainer {
-    -st-partial-mixin: withDynamicPadding(
+    -st-mixin: withDynamicPadding(
             PaddingSize value(mobileInfoPadding),
             BorderWidth value(borderWidth)
     );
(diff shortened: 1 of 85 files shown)


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_414/

src/components/Card/Card.st.css
```diff
@@ -50,7 +50,7 @@
     width: 100%;
     height: 100%;
     background-color: value(cardBGColor);
-    -st-mixin: withBorder(
+    -st-partial-mixin: withBorder(
```
shlomitc, line 53: I think that partial should be used only in the case of extending component and not every mixins.
In this example, this is the class:
```
.withBorder {
    border-style: solid;
    border-color: value(BorderColor);
    border-width: value(BorderWidth);
}
```

when using partial-mixin, it will drop the `border-style` property because it is not affected by the variables.

so this one should be a regular `-st-mixin`

src/components/Card/Card.st.css
```diff
@@ -137,14 +137,14 @@
 .root > .infoContainer {
     box-sizing: border-box;
     background-color: value(infoBGColor);
-    -st-mixin: withDynamicPadding(
+    -st-partial-mixin: withDynamicPadding(
```
shlomitc, line 140: regular -st-mixin

src/components/Card/Card.st.css
```diff
@@ -137,14 +137,14 @@
...
             PaddingSize value(infoPadding),
             BorderWidth value(borderWidth)
     );
 }
 
 .root:stacked > .infoContainer {
-    -st-mixin: withDynamicPadding(
+    -st-partial-mixin: withDynamicPadding(
```
shlomitc, line 147: regular -st-mixin

src/components/Card/Card.st.css
```diff
@@ -154,7 +154,7 @@
 }
 
 .root:mobile > .infoContainer {
-    -st-mixin: withDynamicPadding(
+    -st-partial-mixin: withDynamicPadding(
```
shlomitc, line 157: regular -st-mixin

src/components/Pagination/Pagination.st.css
```diff
@@ -43,7 +43,7 @@
 }
 
 .pageButtonRules {
-  -st-mixin: buttonTransition;
+  /*-st-mixin: buttonTransition;*/
```
shlomitc, line 46: ??

shlomitc (changes requested):
small fixes.
Also - let's mention it in the contribution docs (for whoever uses mixins in a component).
Once stylable will update their docs, we can point there as well

src/components/Card/Card.st.css
jonathanadler: Correct 👍

src/components/Pagination/Pagination.st.css
jonathanadler: It seems there's an issue with nested mixins like these inside a stylesheet that we partially mixin.
So the Example Pagination component uses `-st-partial-mixin`, and this break the build.

I need to investigate more, it might be a bug with the partial mixins or something on our side.
Didn't get to investigate it yet.

I left this comment out as a reminder

shlomitc:
we decided to ditch the partial mixins, so I'm closing this one as well

PR #425, opened 2020-10-06
add Modal files to npm package output

jonathanadler:


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_425/

shlomitc:
Why is this change required?

jonathanadler:
> Why is this change required?

I didn't want consumers to use the Modal directly.
So I added `"!Modal.d.ts"` to the `files` array.

But this also removed the `dist/src/components/Modal/Modal.js` file, so the Dialog doesn't work for consumers.

shlomitc:
You can simply add it in a sub folder of src/components instead.
Also, I think we should support entry points from the src/index.ts file instead of cherry picking (for better esm support)

jonathanadler:
> You can simply add it in a sub folder of src/components instead.
You mean move the `Modal` folder so that it won't be under src/components?
> Also, I think we should support entry points from the src/index.ts file instead of cherry picking (for better esm support)
Yeah, this is a task we have in our backlog
(by cherry picking you mean the `import-path` plugin?)

shlomitc:
Exactly

jonathanadler:
@shlomitc 
So it seems that I can't move the Modal folder, because of the `.wuf` configuration.
It doesn't identify components that aren't under `src/components`.

Since this is a quite urgent PR I'll merge it like this, and hopefully we'll rearrange it afterwards

jonathanadler:
> @shlomitc
> So it seems that I can't move the Modal folder, because of the `.wuf` configuration.
> It doesn't identify components that aren't under `src/components`.
> 
> Since this is a quite urgent PR I'll merge it like this, and hopefully we'll rearrange it afterwards

Wait, my bad.
I have a solution

PR #428, opened 2020-10-10
(a11y )Fix arrow navigation in ColorPicker

alexgoncharwix:


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_428/

jonathanadler:
Hey @alexgoncharwix 

Can you please explain what's the purpose for this PR?

Also, I see that for some reason your last commit isn't signed

src/components/ColorPicker/ColorPicker.st.css
```diff
@@ -11,8 +11,15 @@
 }
 
 .root {
-    display: inline-flex;
+    display: inline-block;
     margin: -4px;
+    border: 0;
+    padding: 0.01em 0 0 0;
+    min-width: 0;
```
jonathanadler, line 18: What are these two for?

src/components/ColorPicker/ColorPicker.tsx
```diff
@@ -7,26 +7,38 @@ import {
 } from 'wix-ui-core/src/components/radio-button/RadioButton';
 import { colorPickerDataHook } from './dataHooks';
 import { TPAComponentProps } from '../../types';
+import { generateKey } from '../../common/random';
 
 interface ColorPickerProps extends TPAComponentProps {
   /** A callback to invoke on a value change */
   onChange?(event: RadioButtonChangeEvent | RadioButtonClickEvent): void;
+  /** aria-label - Accessibility */
+  'aria-label'?: string;
```
jonathanadler, line 16: let's also add an `aria-labelledby` prop

src/components/ColorPicker/ColorPicker.tsx
```diff
@@ -37,10 +49,12 @@ export class ColorPicker extends React.Component<ColorPickerProps> {
               onChange,
               key: index,
               className: classes.item,
+              name: this.groupName,
```
jonathanadler, line 52: I think it would be better to add a `name` prop
If it is undefined then we can use the aria-label as the name.
meaning:

```tsx
name: this.props.name || this.props['aria-label'] || this._groupName
```

Also, please use `_groupName` instead of `groupName`

src/components/ColorPicker/docs/ColorPickerFocusTabExample.tsx
```diff
@@ -0,0 +1,52 @@
...
+  );
+}
+
+export class ColorPickerFocusTabExample extends React.Component {
+  render = () => (
+    <div className={classes.wrapper}>
+      <ColorPickerStateful className={classes.withFocusHighlight} />
+      <ColorPickerStateful className={classes.withFocusHighlight} />
+      <button className={classes.withFocusHighlight}>Select me</button>
+      <button className={classes.withFocusHighlight}>Select me 2</button>
```
jonathanadler, line 49: I'm not sure these buttons are necessary here

src/components/ColorPicker/ColorPicker.st.css
alexgoncharwix: This is a reset for fieldset.
https://gist.github.com/ultrox/84935bf1ce4fc938f61c2142b9a9c1f2

src/components/ColorPicker/docs/ColorPickerFocusTabExample.tsx
alexgoncharwix: This is the last block in the Settings section. After the last ColorPicker in this block the focus goes to browser's url input. It may be confusing when playing with focus behaviour so I added extra buttons.

src/components/ColorPicker/ColorPicker.tsx
alexgoncharwix: Added.

src/components/ColorPicker/ColorPicker.tsx
alexgoncharwix: Done.

src/components/ColorPicker/docs/ColorPickerFocusTabExample.tsx
jonathanadler: I think it's best to remove unrelated components in this tab, since it might cause some confusion. 
Shift-tab should still work, even when going passed the second ColorPIcker, 
and since there are 2 ColorPickers I think this example is pretty clear

jonathanadler (approved):
Just left one comment, regarding the buttons in the example.
But awesome work!

jonathanadler (approved):

PR #433, opened 2020-10-12
feat(Counter): Add `stopPropagation` prop

GuyPerkal:


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_433/

jonathanadler:
@GuyPerkal 
Seems a bit intrusive
Can you explain why this is needed?

GuyPerkal:
We have a clickable container that has counters in it, and we'd like to prevent counters from trigger the container's click event. It's not a must to add this prop since it's possible to wrap the counter with an element that stops propagation, but I figured it's nice to have and would be less hacky.

GuyPerkal:
Sure, you can see it here - https://gilada.wixsite.com/sunny-falafel/mainpage

Add an item to the cart and then open the cart, the item in the cart is clickable and has a counter in it. Currently the propagation is stopped by wrapping with a div, otherwise clicking the counter would open the dish modal.

jonathanadler:
> Sure, you can see it here - https://gilada.wixsite.com/sunny-falafel/mainpage
> 
> Add an item to the cart and then open the cart, the item in the cart is clickable and has a counter in it. Currently the propagation is stopped by wrapping with a div, otherwise clicking the counter would open the dish modal.

I understand...

Usually, we don't expect to have an interactive element like this, inside of a clickable container.
I'm not sure this is something we would want to maintain in the library.

GuyPerkal:
No problem, closing this PR.

PR #434, opened 2020-10-14
Fix/autotools config

argshook:
relevant for https://github.com/wix/wix-ui-tpa/pull/423

in #423 doing `npm run autotools` does not work, throws error:

```
wix-ui-tpa/node_modules/wix-style-react/dist/es/src/FontUpgrade/FontUpgrade.scss:1

.root {
^

SyntaxError: Unexpected token '.'
```

this is because new component `AddItem` tries to require things from
`wix-style-react`. This makes node go through whole `wix-style-react`
tree and eventually find `import { styles } from './FontUpgrade.scss'`.

without a require hook which is able to handle such file,
`autotools-sanity` fails.


jonathanadler (approved):

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_434/

PR #440, opened 2020-10-20
POC - use CSS vars for component customization

shlomitc:
This PR proposes how to use CSS vars for customizing our component.
The goal is to get rid of the `-st-mixin` approach which causes bloated bundles.

I suggest two solutions:
1. Expose props that apply inline CSS vars
**Pros**: simple, declarative, documented API OOTB.
**Cons**: not working with the style processor CSS function

2. Extending stylesheets using stylable and CSS Vars
**Pros**: it works.
**Cons**: coupled to stylable, more complex to extend a component and import the variables, no stylable auto-completion for css vars imports

I really like option 1 but if the style processor functions aren't working, then we should not go in that direction until we find a suitable replacement for the runtime style-processor

@jonathanadler - I couldn't pass the CSS var the style-processor CSS function (e.g. `color(var(--BadgeBgColor))`), I'll be happy if you can give a look and verify that it works

### DIFF
--- a/src/components/Badge/Badge.st.css
+++ b/src/components/Badge/Badge.st.css
@@ -61,8 +61,8 @@
 }
 
 .root:priority(default) {
-    background-color: value(badgeBgColor);
-    color: value(badgeTextColor);
+    background-color: var(--BadgeBgColor, value(badgeBgColor));
+    color: var(--BadgeTextColor, value(badgeTextColor));
     border: none;
 }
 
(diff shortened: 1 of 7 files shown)


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_440/

shlomitc:
"CSS Vars" is a good idea but the Viewer Platform is still far from supporting it for old browsers
Also, there are some limitations that might require a lot more effort in supporting vars.

Also, runtime props => vars is not something that can work with the current style-processor mechanism.

if someday we drop it in favor of something else + have full support for CSS vars, this might be a good direction.

PR #442, opened 2020-10-22
fix(infra) - update the success and error colors to accessible colors

sivanharel:


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_442/

jonathanadler (approved):
And these are all the places that we use these colors?

sivanharel:
> And these are all the places that we use these colors?

I am checking with Michal if we should change the LikeButton checked color as well or not.
Beside that everything was changed.

jonathanadler:
> > And these are all the places that we use these colors?
> 
> I am checking with Michal if we should change the LikeButton checked color as well or not.
> Beside that everything was changed.

Excellent

sivanharel:
> > > And these are all the places that we use these colors?
> > 
> > 
> > I am checking with Michal if we should change the LikeButton checked color as well or not.
> > Beside that everything was changed.
> 
> Excellent

Ok, she said to keep it as is. So I will merge the PR :)

PR #456, opened 2020-10-30
<Badge/> - expose a more performant API for customization

shlomitc:
This PR is a follow up to this initial POC: https://github.com/wix/wix-ui-tpa/pull/452
1. Expose the `badgeOverrides` public API for performant customization with backward compatibility
2. Allow overriding component variation by eliminating the `-st-state` in favor of a raw class)
3. Improve docs

@sivanharel - bullet 2 is a best practice to apply to all of our components and could be relevant in WSR to make theming easier with shorter selectors (cc @EgoziE )


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_456/

src/components/Badge/docs/BadgeExtendedExample.st.css
```diff
@@ -1,12 +1,19 @@
...
         BadgeBgColor '"color(--badgeBgColor)"',
         BadgeTextColor '"color(--badgeTextColor)"',
-        BadgeBorderColor '"color(--badgeBorderColor)"'
     );
-}
\ No newline at end of file
+}
+
+.mixOverrides {
+    -st-mixin: badgeOverrides(
```
shlomitc, line 15: @jonathanadler - what do you think of the API? Think that this should become some kind of a convention for all the components we will create

src/components/Badge/Badge.tsx
```diff
@@ -26,8 +26,13 @@ class Badge extends React.Component<BadgeProps> {
     const { priority, children, className, icon } = this.props;
     return (
       <div
+        className={st(
+          classes.root,
+          { withIcon: !!icon },
+          classes[`priority-${priority}`],
+          className,
+        )}
```
jonathanadler, line 34: So, if I understand correct, what we're actually saying here, is that we want to forbid the use of "variant" states, and allow only booleans.
I wonder though, how can we enforce this?
Some kind of lint rule?

jonathanadler (commented):
I think this is awesome
The only thing I'm "concerned" about, is that it will be difficult to enforce.
Or maybe not enforce, but, we'll need to find an easy way for contributors to understand this approach.
I'm thinking maybe we should separate the stylesheet to two: `Badge.st.css` & `overrides.st.css` (also in the generator).
Maybe this will make it clearer?
It will affect extension as well:
```
:import {
  -st-from: "path/to/Badge/overrides.st.css";
  -st-named: tbd;
}
```

src/components/Badge/Badge.tsx
shlomitc: I just think that we've using stylable wrong all along.
state are there to represents, well, logic states. styles variations are class variations and not states.

so I don't think we need a rule, but rather pass the knowledge and fix ourselves

shlomitc:
> I'm thinking maybe we should separate the stylesheet to two: `Badge.st.css` & `overrides.st.css` (also in the generator).
> 
> ```
> :import {
>   -st-from: "path/to/Badge/overrides.st.css";
>   -st-named: tbd;
> }
> ```

I'm afraid this will not be possible in a way that preserves the existing styling mechanism.
If we separate it to a different file, the css specificity will be different and we will be able to achieve only one of the options:
- only mix the entire file (current behavior, so we did nothing)
- only mix the overrides (the new behavior, but a breaking change)

The only way to go with your approach will be also separating the `priority` themes from the root file. meaning that we will treat each one as a separate variant: `<Badge className={primaryBadge}/>` or `<Badge className={secondaryBadge}/>`.
And then maybe (didn't check yet) the consumer could do `<Badge className={st(primaryBadge, styleParams}/>`. We might be able to do it for new components, but not for the existing ones.

PR #458, opened 2020-11-01
perf(Icons): use tslint import-blacklist to prevent incorrect imports

sivanharel:


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_458/

shlomitc (approved):

tslint_base.json
```diff
@@ -16,6 +16,10 @@
         "singleQuote": true,
         "trailingComma": "all"
       }
+    ],
+    "import-blacklist": [
+      true,
+      [".*\\/assets\\/icons$"]
```
jonathanadler, line 22: I guess we should also add `".*\\/assets\\/icons/social$"`.

Do we also need a rule for implicitly using `index` or `index.tsx` in the path?

jonathanadler (approved):

tslint_base.json
sivanharel: > I guess we should also add `".*\\/assets\\/icons/social$"`.
>
> Do we also need a rule for implicitly using `index` or `index.tsx` in the path?

Yes, I will add it as well.

PR #461, opened 2020-11-02
Text field suffix

sivanharel:


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_461/

src/components/TextField/TextField.tsx
```diff
@@ -38,50 +74,89 @@ export class TextField extends React.Component<TextFieldProps> {
...
+    return hasSuffix ? (
+      <div className={classes.suffixWrapper}>
+        <div className={classes.gap} />
+        {shouldShowClearButton && (
+          <IconButton
+            className={classes.clearButton}
+            data-hook={DATA_HOOKS.CLEAR_BUTTON}
+            aria-label={clearButtonAriaLabel}
+            aria-labelledby={clearButtonAriaLabelledby}
+            onClick={() => onClear && this.props.onClear()}
```
jonathanadler, line 115: I would do
```suggestion
            onClick={onClear ? () => this.props.onClear() : undefined}
```
cause there's no need to pass a function at all if the callback doesn't exist

src/components/TextField/TextField.tsx
```diff
@@ -38,50 +74,89 @@ export class TextField extends React.Component<TextFieldProps> {
...
+    const shouldShowClearButton = withClearButton && value && !disabled;
+
+    const hasSuffix =
+      shouldShowCustomSuffix ||
+      shouldShowErrorIcon ||
+      shouldShowSuccessIcon ||
+      shouldShowClearButton;
+
+    const shouldAddLeftPaddingToCustomSuffix =
+      !shouldShowErrorIcon && !shouldShowSuccessIcon && shouldShowClearButton;
```
jonathanadler, line 104: I think instead of using these states to understand if there needs to be padding or not, I would give all the icons a className like `suffixIcon` and go with a pure css solution like this:
```
.suffixWrapper .suffixIcon {
  padding-right: number;
}

.suffixWrapper .suffixIcon::last-child {
  padding-right: none;
}
```
Haven't tested this of course, so I'm not 100% sure this will work

jonathanadler: btw, how does this work with `rtl`?

src/components/TextField/TextField.tsx
```diff
@@ -104,18 +179,41 @@ export class TextField extends React.Component<TextFieldProps> {
...
+  success,
+  successIcon,
+}) => {
+  let statusIcon = null;
+
+  if (errorMessage && error) {
+    statusIcon = <ErrorSuffix errorMessage={errorMessage} />;
+  } else if (successIcon && success) {
+    statusIcon = <SuccessIcon data-hook={DATA_HOOKS.SUCCESS_ICON} />;
+  }
```
jonathanadler, line 216: Note that in the [original implementation](https://github.com/wix/wix-ui-tpa/blob/master/src/components/TextField/TextField.tsx#L65), the `success` status had precedent, cause it didn't have the `else`.

So it should either be flipped:
```tsx
   if (successIcon && success) {
    statusIcon = <SuccessIcon data-hook={DATA_HOOKS.SUCCESS_ICON} />;
  } else if (errorMessage && error) {
    statusIcon = <ErrorSuffix errorMessage={errorMessage} />;
  }
```
or:
```tsx
  if (errorMessage && error) {
    statusIcon = <ErrorSuffix errorMessage={errorMessage} />;
  }
  if (successIcon && success) {
    statusIcon = <SuccessIcon data-hook={DATA_HOOKS.SUCCESS_ICON} />;
  }
```

jonathanadler (commented):
Wow, nice job!

I left some tiny comments.
And also two things:
1. we still have the icon size issue, no? I'm not sure it should be 8px (the X icon)
2. in the storybook playground, let's maybe add a predefined `suffix` icon?

src/components/TextField/docs/examples.ts
```diff
@@ -24,6 +24,67 @@ export const errorWithErrorMessage = `
...
+              error
+              errorMessage="TestError"
+              suffix={<Calendar />}
+              onChange={this._onChange}
+          />
+        );
+      }
+    }
+`;
+
```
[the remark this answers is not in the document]
sivanharel: yes, we could. But do you think we should add all of the suffix's examples for this theme or one is enough?

src/components/TextField/TextField.tsx
sivanharel: I know that it was the opposite but there was a bug: when you passed all of the props I mean on this case: <TextField successIcon success error errorMessage /> the border of the input was colored red (error border) but the suffix icon was a success icon.
So I fixed it:)
Screen Shot 2020-11-04 at 10 06 36

src/components/TextField/TextField.tsx
sivanharel: I refactored it and fix the icons spaces

jonathanadler (approved):
Awesome!

PR #463, opened 2020-11-03
Infra: improve generator initial boilerplate and style override mechanism

shlomitc:
Following this PR: https://github.com/wix/wix-ui-tpa/pull/456

### DIFF
--- a/.wuf/generator/templates/desktop/Component/Component.st.css
+++ b/.wuf/generator/templates/desktop/Component/Component.st.css
@@ -25,6 +25,6 @@
   background-color: value(defaultBgColor);
 }
 
-.root.styleParams { 
+.root.overrideStyleParams { 
   background-color: value(bgColor);
 }
--- a/.wuf/generator/templates/desktop/Component/Component.st.css
+++ b/.wuf/generator/templates/desktop/Component/Component.st.css
@@ -22,5 +22,9 @@
 }
 
 .root {
+  background-color: value(defaultBgColor);
+}
+
+.styleParams { 
   background-color: value(bgColor);
 }
--- a/.wuf/generator/templates/desktop/Component/Component.st.css
+++ b/.wuf/generator/templates/desktop/Component/Component.st.css
@@ -1,18 +1,26 @@
 :import {
     -st-from: "../../common/formatters.st";
-    -st-named: color;
+    -st-named: color, fallback;
 }
 
-/*Defaults*/
+/* Defaults - default params */
 :vars {
-    /*Defaults params goes here*/
+  defaultBgColor: color(color-1);
 }
 
-/*Overrides*/
+/* Overrides - deceleration of overridable parameters by the consumer */
 :vars {
-    /*Deceleration of overridable placeholders*/
+  /*
+    Main background color
+  */
+  BgColor: --overridable;
+}
+
+/* Overrides */
+:vars {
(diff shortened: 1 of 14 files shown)


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_463/

.wuf/generator/templates/desktop/Component/Component.st.css
```diff
@@ -1,18 +1,30 @@
...
 }
 
-/*Overrides*/
+/* Overrides - deceleration of overridable parameters by the consumer */
 :vars {
-    /*Deceleration of overridable placeholders*/
+  /*
+    Main background color
+  */
+  BgColor: --overridable;
```
jonathanadler, line 16: Do you maybe want to use the autodocs Stylable plugin "syntax" of adding defaults?
https://github.com/wix/wix-ui-tpa/blob/48aa71d1a6cd07f139561d4b559a80754e72977c/src/components/TextArea/TextArea.st.css#L51

And maybe we don't even need to split the `defaults` and the `overridables`?
Just have it:
```css
  /*
    Main background color
    @default color-1
  */
  BgColor: color(color-1);
```

.wuf/generator/templates/desktop/Component/Component.st.css
```diff
@@ -1,18 +1,30 @@
...
 }
 
 .root {
-    /*css goes here*/
+  background-color: value(defaultBgColor);
 }
+
+.styleParams { 
+  background-color: value(bgColor);
+}
```
jonathanadler, line 30: even though the TPA's use the `styleParams` to refer to the variables coming from the settings
I'm not sure that in this context it's completely understood

I mean, if this wasn't `css` I think I would go with `applyStyles` or something like that
but then `.applyStyles{ ... }` is a bit weird

idk, mainly thinking out loud

jonathanadler:
Also, we probably need to think of how to document this, and let developers know that this is the "new" way to override styles.
We'll maybe need some indicator for each component

.wuf/generator/templates/desktop/Component/Component.st.css
shlomitc: Other options can be paramsOverrides or overrideParams.

I want a name that is short yet explicit enough to explain the intent.

shlomitc:
> Also, we probably need to think of how to document this, and let developers know that this is the "new" way to override styles.
> We'll maybe need some indicator for each component

We can have a single doc (reflected in the storybook as well) that explains the pattern and point to it from each story page

.wuf/generator/templates/desktop/Component/Component.st.css
shlomitc: I'll need to experiment it further so I'm keeping it out of this current PR

jonathanadler (approved):

jonathanadler:
> > Also, we probably need to think of how to document this, and let developers know that this is the "new" way to override styles.
> > We'll maybe need some indicator for each component
> 
> We can have a single doc (reflected in the storybook as well) that explains the pattern and point to it from each story page

Should this be part of this PR?

shlomitc:
image

jonathanadler (approved):
Inspiring 👏 👏 👏

shlomitc:
merged as part of #467

PR #464, opened 2020-11-05
release: v2.9.2

jonathanadler:


package.json
```diff
@@ -2,7 +2,7 @@
   "private": false,
   "name": "wix-ui-tpa",
   "sideEffects": false,
-  "version": "2.9.1",
+  "version": "2.9.2",
```
sivanharel, line 5: Since this version includes a new feature shouldn't it be a minor version instead of a patch?

jonathanadler: That's true 😄 
But since it's only one small change, and mainly since Shlomi's PR refers to a `2.10.0`version ([here](https://github.com/wix/wix-ui-tpa/pull/463/files#diff-52296615debf3994195e43d91c94e02bfdc615a27620e0dbc72140b3de98dde1R64))
I preferred using a patch for now

jonathanadler: We can change this though, and Shlomi's PR

sivanharel (approved):

PR #470, opened 2020-11-10
Image - implement MVP with basic props

nitayneeman:
This PR implements a minimal Image component supporting basic props which also exposes the `mediaItem` as a short-term solution.


src/components/Image/Image.st.css
```diff
@@ -0,0 +1,18 @@
+:import {
```
nitayneeman, line 1: I wonder if to remove this file (but we'll probably need it soon)

src/components/Image/docs/ImageWiringExample.tsx
```diff
@@ -0,0 +1,12 @@
+import * as React from 'react';
```
nitayneeman, line 1: I wonder if to remove also this one (with ImageWiringExample.st.css) until we've a settings panel support?

src/components/Image/Image.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  render() {
+    const { className, src, mediaItem, ...imageProps } = this.props;
+
+    return (
+      <div
+        className={st(classes.root, className)}
+        data-hook={this.props['data-hook']}
+      >
+        {mediaItem ? (
```
nitayneeman, line 36: Maybe we should log an error when both `src` & `mediaItem` are empty? Because these props are optional and cannot be required (otherwise we can have invalid use-case)...

Another option is logging a warning that one of them must be with value when both are empty

src/components/Image/Image.driver.ts
```diff
@@ -0,0 +1,13 @@
...
+  BaseUniDriver,
+  baseUniDriverFactory,
+} from 'wix-ui-test-utils/base-driver';
+import { UniDriver } from 'wix-ui-test-utils/unidriver';
+
+export interface ImageDriver extends BaseUniDriver {}
+
+export const imageDriverFactory = (base: UniDriver): ImageDriver => {
+  return {
+    ...baseUniDriverFactory(base),
```
nitayneeman, line 11: Do you think we should expose the methods of MediaImage's driver at the moment knowing this is a temporary solution until we've the final spec? 🤔

image

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_470/

src/components/Image/Image.tsx
jonathanadler: I'm not sure it's needed at the moment.
I think we can expect our consumers to pass at least one of them.

src/components/Image/Image.tsx
```diff
@@ -0,0 +1,44 @@
...
+  static displayName = 'Image';
+
+  render() {
+    const { className, src, mediaItem, ...imageProps } = this.props;
+
+    return (
+      <div
+        className={st(classes.root, className)}
+        data-hook={this.props['data-hook']}
+      >
```
jonathanadler, line 35: This isn't critical at all, but I wonder if the wrapping `div` is necessary 🤔 ?
It probably depends if `MediaImage` accepts the props, and also if it might be useful for design in the future.

src/components/Image/Image.tsx
```diff
@@ -0,0 +1,44 @@
...
+        {mediaItem ? (
+          <MediaImage {...imageProps} mediaPlatformItem={mediaItem} />
+        ) : (
+          <img {...imageProps} src={src} />
```
jonathanadler, line 39: Should we use wix-ui-core's `Image` component?

src/components/Image/README.md
```diff
@@ -0,0 +1,4 @@
+## Image
+
+Image is a component to literally display an image - whether a regular with full path or an item from the media manager.
+
```
jonathanadler, line 4: This file can be removed

src/components/Image/docs/ImageWiringExample.tsx
jonathanadler: I guess we can remove this for now

src/components/Image/docs/index.story.tsx
```diff
@@ -0,0 +1,93 @@
...
+              example: <ImageWiringExample />,
+              rawSource: ImageWiringExampleRaw,
+              rawCSSSource: ImageWiringExampleCSSRaw,
+              params: {
+                colors: [],
+                fonts: [],
+                numbers: [],
+              },
+            }),
+          ],
```
jonathanadler, line 88: I think we can remove this or now

jonathanadler (commented):
Looks great!

Just left some minor comments

src/components/Image/Image.driver.ts
nitayneeman: Yeah, `click` is redundant. 
I think we can implement `getSrc`, `getAlt` and `getLoadStatus` for now.
Regarding `isLoaded` - seems like it's not implemented by the driver of ui-core's Image - not sure we actually need this if we already have `getLoadStatus`?

src/components/Image/Image.meta.tsx
```diff
@@ -0,0 +1,17 @@
...
+  title: 'render',
+  props: {} as ImageProps,
+});
+
+ImageMetadata.addSim({
+  title: 'render with alt',
+  props: {
+    alt: 'some alternative text',
+  } as ImageProps,
+});
```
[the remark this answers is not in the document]
nitayneeman: They are not critical specifically for `alt` (it should work even without them) - do you think we should add them anyway?

src/components/Image/Image.tsx
nitayneeman: It seems `MediaImage` doesn't accept `data-hook` and `className`...
I guess we'll need the wrapper until we fix `MediaImage` in ui-core (if we decide to)

src/components/Image/Image.tsx
nitayneeman: That's actually a very good idea, mainly to benefit from its driver

src/components/Image/Image.driver.ts
sivanharel: I think you are right and the 3 methods that you wrote are enough. It seems like the getLoadStatus includes the (isLoaded, isError, and the isLoading). If I understood correctly :)

jonathanadler: but what about the `error` state?
We might want to support this, no?
ie:
```tsx
<img onError={this.props.onError} ...>
```
I'm assuming the `MediaImage` component also supports it

src/components/Image/Image.tsx
```diff
@@ -0,0 +1,61 @@
...
+            {...imageProps}
+            {...dimensions}
+            mediaPlatformItem={mediaItem}
+          />
+        ) : (
+          <CoreImage
+            {...imageProps}
+            nativeProps={{ ...dimensions }}
+            src={src}
+          />
```
jonathanadler, line 56: Regarding the core Image component, my previous comment was meant to be a question.
I just wanted to understand if we should prefer using the core Image rather than a native img, or vice versa

jonathanadler (commented):
Just left some small comments, sorry 😬

jonathanadler:
Also, maybe it would be good to add a `mediaItem` example for the playground?

src/components/Image/Image.tsx
nitayneeman: I think it's a good idea to use ui-core's Image since it's already implemented - it supports `srcSet`, `errorImage` and `resizeMode` that we might want in the future.

At the moment it doesn't really matter (maybe a bundle size difference), but it does help to implement the driver (otherwise we'll need to implement `isLoaded` on our own)

src/components/Image/Image.driver.ts
nitayneeman: I've planned to support error and image failure in different PR (out of the MVP scope).
Probably we'll want something like `onError` but in general we need to design the failure mode...

jonathanadler (approved):
For now it's good 👍 

I think we might want to add an `onError` prop, and add the corresponding driver method

Issue #468, opened 2020-11-10
Dialog component broken on dark themes (where text is mapped to white)


sarunas:
Dialog component is broken on dark themes cause (where text is mapped to white) cause Dialog component has fixes background as #FFF.

jonathanadler:
Hey @sarunas can you provide a link to a live site with this issue?
It will help us debug the issue and create a fix for it
Thanks!

sarunas:
<img width="1790" alt="Screenshot 2020-11-10 at 17 12 53" src="https://user-images.githubusercontent.com/250523/98692699-1593fd80-2378-11eb-8e00-920477cc832a.png">
<img width="1786" alt="Screenshot 2020-11-10 at 17 13 11" src="https://user-images.githubusercontent.com/250523/98692707-16c52a80-2378-11eb-9a34-d569b4f01e35.png">

sarunas:
@jonathanadler all good or you need more details from me?

jonathanadler:
Yes, thank you @sarunas.

The issue here, is that by the design system's spec, the Dialog should not be wired to the sites palette:
https://zeroheight.com/7sjjzhgo2/p/447d36-06-overlays/t/04f416

So this means that the `Dialog`'s background should always stay white, and the foreground color should stay black.

However, since the `Dialog` accepts any content through the `children` prop, we currently don't have a way to restrict the foreground color of the content.
This means the responsibility for keeping the content with a black foreground, is currently on the TPA's side.

To fix this, we will:
1. First allow you to override the `Dialog`'s background color - so that it will be wired to the site's template
2. Then find a way to restrict all the `Dialog`'s content to be grayscale by default

However, please consult with your designers to understand if your `Dialog` should be wired to the palette, or stay black & white.

Thanks!

PR #475, opened 2020-11-15
Dialog allow style wiring

jonathanadler:
This PR is a preliminary fix for https://github.com/wix/wix-ui-tpa/issues/468

At the moment the Dialog component doesn't allow overriding the background color and allowing it to be wired to the site colors.

This PR allows this behaviour, but still leaves the responsibility for passing black and white components to the Dialog, on the TPA's side.

### DIFF
--- a/src/components/Dialog/Dialog.st.css
+++ b/src/components/Dialog/Dialog.st.css
@@ -5,7 +5,7 @@
 
 :import {
     -st-from: "../IconButton/IconButton.st.css";
-    -st-default: TPAIconButton;
+    -st-named: skin-line;
 }
 
 /*Defaults*/
@@ -54,7 +54,7 @@
 }
 
 .closeIconButton {
-    -st-mixin: TPAIconButton(
+    -st-mixin: skin-line(
             IconColor value(DefaultCloseButtonColor)
     );
 }
@@ -87,7 +87,7 @@
 }
 
 .overrideStyleParams:wired .closeIconButton {
-    -st-mixin: TPAIconButton(
+    -st-mixin: skin-line(
             IconColor value(WiredCloseButtonColor)
     );
 }
--- a/src/components/Dialog/Dialog.st.css
+++ b/src/components/Dialog/Dialog.st.css
@@ -77,17 +77,6 @@
     left: 20px;
 }
 
-/* Deprecated - Old style params overriding for full mixin */
-.root:wired .contentWrapper {
-    background: value(WiredBackgroundColor);
-}
-
-.root:wired .closeIconButton {
-    -st-mixin: TPAIconButton(
-            IconColor value(WiredCloseButtonColor)
-    );
-}
-
 /* New style overrides capabilities */
 .overrideStyleParams {
     -st-extends: root;
--- a/src/components/Dialog/Dialog.st.css
+++ b/src/components/Dialog/Dialog.st.css
@@ -8,14 +8,6 @@
     -st-named: skin-line;
 }
 
-/*Defaults*/
-:vars {
-    DefaultWiredCloseButtonColor: color-5;
-    DefaultCloseButtonColor: #000000;
-    DefaultWiredBackgroundColor: color-1;
-    DefaultBackgroundColor: #ffffff;
-}
-
 /*Overrides*/
 :vars {
     CloseButtonColor: --overridable;
@@ -24,12 +16,14 @@
 
 /*Fallbacks*/
 :vars {
-    WiredCloseButtonColor: color(fallback(value(CloseButtonColor), value(DefaultWiredCloseButtonColor)));
-    WiredBackgroundColor: color(fallback(value(BackgroundColor), value(DefaultWiredBackgroundColor)));
+    FixedCloseButtonColor: color(fallback(value(CloseButtonColor), black));
+    FixedBackgroundColor: color(fallback(value(BackgroundColor), white));
+    WiredCloseButtonColor: color(fallback(value(CloseButtonColor), color-5));
+    WiredBackgroundColor: color(fallback(value(BackgroundColor), color-1));
 }
 
 .root {
-    -st-states: mobile, rtl, wired;
+    -st-states: mobile, rtl;
 }
 
 .contentWrapper {
@@ -43,20 +37,12 @@
     top: 50%;
     left: 50%;
     transform: translate(-50%, -30%);
-
-    background: value(DefaultBackgroundColor);
 }
 
 .closeButtonWrapper {
     position: absolute;
     top: 20px;
     right: 20px;
-}
-
-.closeIconButton {
-    -st-mixin: skin-line(
-            IconColor value(DefaultCloseButtonColor)
-    );
 }
 
 .dialogContent {
@@ -78,16 +64,22 @@
 }
 
 /* New style overrides capabilities */
-.overrideStyleParams {
-    -st-extends: root;
+.skin-fixed .contentWrapper {
+    background: value(FixedBackgroundColor);
 }
 
-.overrideStyleParams:wired .contentWrapper {
+.skin-fixed .closeIconButton {
+    -st-mixin: skin-line(
+        IconColor value(FixedCloseButtonColor)
+    );
+}
+
+.skin-wired .contentWrapper {
     background: value(WiredBackgroundColor);
 }
 
-.overrideStyleParams:wired .closeIconButton {
+.skin-wired .closeIconButton {
     -st-mixin: skin-line(
-            IconColor value(WiredCloseButtonColor)
+        IconColor value(WiredCloseButtonColor)
     );
 }
--- a/src/components/Dialog/Dialog.st.css
+++ b/src/components/Dialog/Dialog.st.css
@@ -50,7 +50,6 @@
 }
 
 .root:mobile .contentWrapper {
-    height: 100%;
     top: 0;
     left: 0;
     width: 100%;
@@ -65,7 +64,7 @@
 
 /* New style overrides capabilities */
 .skin-fixed .contentWrapper {
-    background: value(FixedBackgroundColor);
+    background-color: value(FixedBackgroundColor);
 }
 
 .skin-fixed .closeIconButton {
@@ -75,7 +74,7 @@
 }
 
 .skin-wired .contentWrapper {
-    background: value(WiredBackgroundColor);
+    background-color: value(WiredBackgroundColor);
 }
 
 .skin-wired .closeIconButton {
--- a/src/components/Dialog/Dialog.st.css
+++ b/src/components/Dialog/Dialog.st.css
@@ -10,21 +10,26 @@
 
 /*Defaults*/
 :vars {
-    DefaultCloseButtonColor: color-5;
+    DefaultWiredCloseButtonColor: color-5;
+    DefaultCloseButtonColor: #000000;
+    DefaultWiredBackgroundColor: color-1;
+    DefaultBackgroundColor: #ffffff;
 }
 
 /*Overrides*/
 :vars {
     CloseButtonColor: --overridable;
+    BackgroundColor: --overridable;
 }
 
 /*Fallbacks*/
 :vars {
-    CurrentCloseButtonColor: color(fallback(value(CloseButtonColor), value(DefaultCloseButtonColor)));
+    WiredCloseButtonColor: color(fallback(value(CloseButtonColor), value(DefaultWiredCloseButtonColor)));
+    WiredBackgroundColor: color(fallback(value(BackgroundColor), value(DefaultWiredBackgroundColor)));
 }
 
 .root {
-    -st-states: mobile, rtl;
+    -st-states: mobile, rtl, wired;
 }
 
 .contentWrapper {
@@ -39,7 +44,23 @@
     left: 50%;
     transform: translate(-50%, -30%);
 
-    background: #fff;
+    background: value(DefaultBackgroundColor);
+}
+
+.closeButtonWrapper {
+    position: absolute;
+    top: 20px;
+    right: 20px;
+}
+
+.closeIconButton {
+    -st-mixin: TPAIconButton(
+            IconColor value(DefaultCloseButtonColor)
+    );
+}
+
+.dialogContent {
+    padding: 80px 32px;
 }
 
 .root:mobile .contentWrapper {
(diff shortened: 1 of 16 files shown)


src/components/Dialog/Dialog.visual.tsx
```diff
@@ -3,36 +3,46 @@ import { visualize, story, snap } from 'storybook-snapper';
...
-          </Button>
-        </div>
-      </Dialog>,
-    );
+    snap('Dialog with some content', <DialogWithContent />);
+  });
+
+  story('Wired to palette', () => {
+    snap('Dialog with some content', () => {
+      setDarkPalette();
```
shlomitc, line 43: This will cause a side effect. every test that comes afterward will stay in the dark palette.

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_475/

src/components/Dialog/Dialog.st.css
```diff
@@ -51,23 +72,17 @@
...
-            IconColor value(CurrentCloseButtonColor)
-    );
+.root:wired .contentWrapper {
+    background: value(WiredBackgroundColor);
 }
 
-.dialogContent {
-    padding: 80px 32px;
+.root:wired .closeIconButton {
+    -st-mixin: TPAIconButton(
```
shlomitc, line 85: this will cause an increase in bundle size obviously.

src/components/Dialog/Dialog.st.css
```diff
@@ -39,7 +44,23 @@
...
+}
+
+.closeButtonWrapper {
+    position: absolute;
+    top: 20px;
+    right: 20px;
+}
+
+.closeIconButton {
+    -st-mixin: TPAIconButton(
```
shlomitc, line 57: don't you want to use here the `overrideStyleParams` pattern?

src/components/Dialog/Dialog.tsx
```diff
@@ -31,6 +31,8 @@ export interface DialogProps extends TPAComponentProps {
   closeButtonAriaLabel?: string;
   /** Identifies the element that labels the close button element. Optional. */
   closeButtonAriaLabelledby?: string;
+  /** Whether the Dialog is wired to the site palette, or has a white background */
+  wiredToSiteColors?: boolean;
```
shlomitc, line 35: following our discussion. since the component knows it should not be wired. I think we can keep this logic internal.

wrap the children with

```
<TPAComponentsProvider value ={{mobile, rtl, wiredToSiteColors}}>
...
</TPAComponentsProvider>
```

and then, every component (`IconButton` in this case) will apply a `black-and-white` skin

In the rare case that some would like to opt-out the black and white theme, only then he should pass `wiredToSiteColors=true`

src/components/Dialog/Dialog.st.css
jonathanadler: I thought about it, but we'll need to support both approaches
for backwards compatibility.
Or we can break the api, and talk to whoever's using it

src/components/Dialog/Dialog.st.css
jonathanadler: Unfortunately, yes
But I guess it's a small increase, no?

src/components/Dialog/Dialog.visual.tsx
jonathanadler: I was hoping the tests would run in parallel, so that it won't affect the rest of the tests.
I think the batch runs in the same window 😞

src/components/Dialog/Dialog.tsx
jonathanadler: So I actually thought that we'll have a different provider for this, and the Dialog component will use it.
Since the `TPAComponentsProvider` is in the app level, I'm not sure it's the right place.

So each component will use this context:
```tsx
render() {
  ...
  return (
    <BnWContext.Consumer>
      {
        (isBnW) => (
          ...
        )
      }
    </BnWContext.Consumer>
  )
}
```

and the `Dialog` component will use the provider:
```tsx
render() {
  ...
  return (
    <BnWContext.Provider value={{ isBnW: true }}>
      ...
    </BnWContext.Provider>
  );
}
```

jonathanadler: and also, I preferred working on this in another PR

src/components/Dialog/Dialog.st.css
```diff
@@ -51,23 +72,22 @@
...
 }
 
-.closeIconButton {
-    -st-mixin: TPAIconButton(
-            IconColor value(CurrentCloseButtonColor)
-    );
+/* New style overrides capabilities */
+.overrideStyleParams {
+    -st-extends: root;
 }
```
jonathanadler, line 83: @shlomitc is this better than using the `root`?
Or is it actually the same?

src/components/Dialog/Dialog.st.css
```diff
@@ -5,26 +5,31 @@
...
+    -st-named: skin-line;
 }
 
 /*Defaults*/
 :vars {
-    DefaultCloseButtonColor: color-5;
+    DefaultWiredCloseButtonColor: color-5;
+    DefaultCloseButtonColor: #000000;
+    DefaultWiredBackgroundColor: color-1;
+    DefaultBackgroundColor: #ffffff;
```
shlomitc, line 16: Since these variable are used mostly once, I think it's better to add them directly where required
If I take `DefaultWiredCloseButtonColor` and the combined `WiredCloseButtonColor`, I'd simply do

```diff
- DefaultWiredCloseButtonColor: color-5;
- WiredCloseButtonColor: color(fallback(value(CloseButtonColor), value(DefaultWiredCloseButtonColor)));
+ wiredCloseButtonColor: color(fallback(value(CloseButtonColor), color-5));

- DefaultCloseButtonColor: #000000;
+ defaultCloseButtonColor: #000000;
```

The lower case indicates by convention that it's internal (in oppose to the capital case that is overridable)

I'd also consider making the `default` variation themeable
```diff
+ defaultCloseButtonColor: color(fallback(value(CloseButtonColor), #000000));
```

although this is not a recommended approach, it will make the API consistent. (overriding CloseButtonColor will always work)

src/components/Dialog/Dialog.st.css
```diff
@@ -39,7 +44,23 @@
...
+    position: absolute;
+    top: 20px;
+    right: 20px;
+}
+
+.closeIconButton {
+    -st-mixin: skin-line(
+            IconColor value(DefaultCloseButtonColor)
+    );
+}
```
shlomitc, line 60: this can be transformed to a `skin-fixed-colors`
`-st-mixin: skin-fixed-colors(IconColor value(CloseButtonColor))`

and the `skin-fixed-colors` variation will be an overridable parameter with a fallback to `#000000`

In my head, it's like propagating a prop const `Dialog = ({skin}) => <div> ... <IconButton skin={skin}/> ... </div>`

src/components/Dialog/Dialog.st.css
```diff
@@ -5,26 +5,31 @@
...
 /*Fallbacks*/
 :vars {
-    CurrentCloseButtonColor: color(fallback(value(CloseButtonColor), value(DefaultCloseButtonColor)));
+    WiredCloseButtonColor: color(fallback(value(CloseButtonColor), value(DefaultWiredCloseButtonColor)));
+    WiredBackgroundColor: color(fallback(value(BackgroundColor), value(DefaultWiredBackgroundColor)));
 }
 
 .root {
-    -st-states: mobile, rtl;
+    -st-states: mobile, rtl, wired;
```
shlomitc, line 32: As I see it, `wired` is simply another skin. so I wouldn't add it as a state, but rather as a class.
This will remove the `-st-extends: root` from `overrideStyleParams`

src/components/Dialog/Dialog.tsx
```diff
@@ -60,13 +62,18 @@ export class Dialog extends React.Component<DialogProps> {
...
     } = this.props;
 
     return (
       <TPAComponentsConsumer>
         {({ mobile, rtl }) => (
           <div
-            className={st(classes.root, { mobile, rtl }, className)}
+            className={st(
+              classes.root,
+              { mobile, rtl, wired: wiredToSiteColors },
```
shlomitc, line 74: following the comment above, I'd change `wired` to be `skin-wired`

src/components/Dialog/Dialog.st.css
jonathanadler: So just to see I understand, 
Theoretically `mobile` and `rtl` can also be a "skin".
So maybe we don't want to use stylable states at all?
and just use classes in all our components?

I think that could be an easier solution to maintain.

However, then it immediately raises the question, then why do stylable states exist at all?
I seem to be confused about how stylable works 😕

src/components/Dialog/Dialog.st.css
jonathanadler: 👍

src/components/Dialog/Dialog.visual.tsx
```diff
@@ -3,36 +3,46 @@ import { visualize, story, snap } from 'storybook-snapper';
...
+    snap('Dialog with some content', <DialogWithContent />);
   });
+
+  // story('Wired to palette', () => {
+  //   snap('Dialog with some content', () => {
+  //     // setDarkPalette();
+  //
+  //     return <DialogWithContent wired />;
+  //   });
+  // });
```
nitayneeman, line 47: Redundant?

src/components/IconButton/docs/IconButtonExtendedExample.tsx
```diff
@@ -6,23 +6,36 @@ import { IconButton, Skins } from '../IconButton';
...
         Skin 'full'
       </div>
+
+      <h3>*Deprecated* - Override the style params in the entire css file</h3>
+      <div>
+        <IconButton
+          icon={<ShareIcon />}
+          skin={Skins.Line}
+          className={classes.mixAll}
+        />{' '}
```
nitayneeman, line 36: Redundant?
```suggestion
        />
```

src/test/visualTestUtils.ts
```diff
@@ -0,0 +1,23 @@
...
+export const setDarkPalette = () => {
+  const obj = { ...SettingsChangedEvent };
+  obj.params.siteColors.forEach((colorObj, ind) => {
+    const col = ind % 5;
+    const row = Math.floor(ind / 5);
+
+    if (row < 5 && col < 5) {
+      colorObj.value = DARK_PALETTE[row][col];
+    }
+  });
```
nitayneeman, line 21: I wonder if we want to unit-test this logic?

src/components/Dialog/Dialog.tsx
nitayneeman: Do you think the provider will have more options than just `isBnW`?
Not sure the use-case justifies a completely new provider at the moment

src/components/IconButton/docs/index.story.tsx
```diff
@@ -38,8 +38,12 @@ const iconExamples = [
...
-  baseCode({ components: allComponents, compact: true, ...config });
+const example = (config, extraContext = {}) =>
+  baseExample({
+    components: { ...allComponents, ...extraContext },
+    compact: true,
+    ...config,
+  });
 
 export default {
   category: StoryCategory.COMPONENTS,
```
sivanharel, line 49: If you are already making some changes here maybe you could add a description for that component :)

sivanharel (commented):
Nice work :)
I left some small comments

src/components/Dialog/Dialog.tsx
jonathanadler: We have to use a new provider, since our `TPAComponentsProvider` is used in the application level, passing application level properties.
In this case we need a specific context for components inside the Dialog (or inside this new provider, for that matter).
One provider won't work.
In any case, this will be the next PR

src/components/Dialog/Dialog.visual.tsx
jonathanadler: I wanted to leave this on purpose.
I tried testing the component in a dark palette, but currently changing the palette affects the whole batch of tests, and not only this specific one.
I've opened a feature request for Applitools, I hope they will give us a way to perform a cleanup after a test is run.
This way we can change the palette to the default one after this test is performed

src/test/visualTestUtils.ts
jonathanadler: I don't think we need to.
At the moment this logic isn't used (it's the commented-out code in the visual test).
In addition, it's supposed to be a util for the visual tests, so if something breaks in this logic, then the visual tests will break. So it kind of takes care of itself

src/components/IconButton/docs/IconButtonExtendedExample.tsx
jonathanadler: I think this for having some space between the button and the text

src/components/Dialog/Dialog.st.css
```diff
@@ -5,22 +5,21 @@
...
 :import {
     -st-from: "../IconButton/IconButton.st.css";
-    -st-default: TPAIconButton;
-}
-
-/*Defaults*/
-:vars {
-    DefaultCloseButtonColor: color-5;
+    -st-named: skin-line;
 }
```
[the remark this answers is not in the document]
jonathanadler: see https://github.com/wix/wix-ui-tpa/pull/475#discussion_r524654453 😁

src/components/Dialog/Dialog.st.css
```diff
@@ -5,22 +5,21 @@
...
+    BackgroundColor: --overridable;
 }
 
 /*Fallbacks*/
 :vars {
-    CurrentCloseButtonColor: color(fallback(value(CloseButtonColor), value(DefaultCloseButtonColor)));
+    FixedCloseButtonColor: color(fallback(value(CloseButtonColor), black));
+    FixedBackgroundColor: color(fallback(value(BackgroundColor), white));
+    WiredCloseButtonColor: color(fallback(value(CloseButtonColor), color-5));
+    WiredBackgroundColor: color(fallback(value(BackgroundColor), color-1));
```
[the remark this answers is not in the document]
jonathanadler: `CloseButtonColor` and `BackgroundColor` are the overridable colors.
We want to let the consumers have a chance to override the styles in an extreme case where needed, without wiring the component to the site colors.
So for instance they might need the Dialog to not be wired, but they want to change the close button color to be red (shouldn't really happen, but it's just for edge cases like this)

src/components/IconButton/docs/index.story.tsx
jonathanadler: 💪

src/components/Dialog/Dialog.visual.tsx
sivanharel: So maybe we could add a todo or a comment.

sivanharel (approved):

nitayneeman (approved):

PR #484, opened 2020-11-22
docs: explain how to use RTL

nitayneeman:
This PR explains what's RTL and how to use it as a consumer.

You can see a readable version here:
https://github.com/wix/wix-ui-tpa/blob/84e5d7baf69110a12056cfaa269265fa4233b887/docs/RTL.md

### DIFF
--- a/docs/RTL.md
+++ b/docs/RTL.md
@@ -3,3 +3,28 @@
 Right-to-left (also known as RTL) is a form of writing so that the direction starts from the right and continues to the left.
 
 The components of `wix-ui-tpa` are RTL-compatible and support that built-in, meaning that all the structure of the component changes from right to the left as well.
+
+## How to Use
+
+In order to apply RTL correctly for the library we'll need to use two things:
+
+1. Using `TPAComponentsProvider` while enabling the `rtl` option
+2. Setting the `dir` attribute as `rtl` on the root element
+
+Here's an example for `App.js`:
+
+```jsx
+import React from 'react';
+import { TPAComponentsProvider } from 'wix-ui-tpa/TPAComponentsConfig';
+
+// Enabling RTL
+const isRTL = true;
+
+export const App: React.FC = () => (
+  <TPAComponentsProvider value={{ rtl: isRTL }}>
+    <MyApp />
+  </TPAComponentsProvider>
+);
+```
+
+The example above demonstrates an application that imports `TPAComponentsProvider` and
--- a/docs/RTL.md
+++ b/docs/RTL.md
@@ -11,14 +11,17 @@ In order to apply RTL correctly for the library we'll need to use two things:
 1. Using `TPAComponentsProvider` while enabling the `rtl` option
 2. Setting the `dir` attribute as `rtl` on the root element
 
-Here's an example for `App.js`:
+Here's an example for `App.ts`:
 
 ```jsx
 import React from 'react';
 import { TPAComponentsProvider } from 'wix-ui-tpa/TPAComponentsConfig';
 
-// Enabling RTL
+// Step 1: Enabling RTL
 const isRTL = true;
+
+// Step 2: Setting `dir` as `rtl`
+const MyApp = () => <div dir="rtl">Hello World!</div>;
 
 export const App: React.FC = () => (
   <TPAComponentsProvider value={{ rtl: isRTL }}>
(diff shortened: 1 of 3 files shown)


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_484/

docs/RTL.md
```diff
@@ -0,0 +1,41 @@
...
+Right-to-left (also known as RTL) is a form of writing so that the direction starts from the right and continues to the left.
+
+The components of `wix-ui-tpa` are RTL-compatible and support that built-in, meaning that all the structure of the component changes from right to the left as well.
+
+## How to Use
+
+In order to apply RTL correctly for the library we'll need to use two things:
+
+1. Using `TPAComponentsProvider` while enabling the `rtl` option
+2. Setting the `dir` attribute as `rtl` on the root element
```
jonathanadler, line 12: Also getting the direction from the `body` is ok, so I guess we can say that consumers just need to ensure the components have an ancestor with the correct direction.
I mean, adding the `dir` attribute to the root of the app will do, but they might get it also from the viewer.
What do you think?

docs/RTL.md
```diff
@@ -0,0 +1,41 @@
+# Using RTL
+
+Right-to-left (also known as RTL) is a form of writing so that the direction starts from the right and continues to the left.
+
+The components of `wix-ui-tpa` are RTL-compatible and support that built-in, meaning that all the structure of the component changes from right to the left as well.
```
jonathanadler, line 5:
```suggestion
`wix-ui-tpa` components support RTL out-of-the-box, meaning, they adjust themselves to the correct reading direction of the app.
```

docs/RTL.md
```diff
@@ -0,0 +1,41 @@
...
+
+The example above demonstrates an application that uses `TPAComponentsProvider` to enable `rtl`.
+
+Also it renders `<MyApp/>` which constitutes our application in practice. Notice we set `dir` as `rtl` on the root of `<MyApp/>`.
+
+## How It Works
+
+The components of the library use `TPAComponentsConsumer` to consume the rtl configuration, and so adjust their structure and/or style to this value. The adjustments might be done using rendering different HTML, flexbox and more techniques.
+
+Also, some of the components do relay on the `dir` to allow the compatibility - and that's why we should set this attribute as `rtl` as consumers of the library.
```
jonathanadler, line 41: For this line, I would maybe add that we try to create components that are agnostic to the reading direction, which means they adjust themselves to the natural reading direction of the app, and this is done by adding the `dir` attribute to the document.
This is actually for all components, and not some.
In addition, maybe it would be nice to show how we achieve this.
For instance, flex gives us this capability.
But also, we use `::before` and `::after` as a way of achieving a kind of `padding-start`/`margin-start` (and `padding-end`/`margin-end`)

jonathanadler: So just a small fix:
actually all components require the `dir` attribute, and only some require the context prop

jonathanadler (commented):
Looks great!

I left some comments.

And maybe we should reference this document in `USAGE.md`?

docs/RTL.md
nitayneeman: 👌

docs/RTL.md
nitayneeman: Good point. 

I'll change the second bullet to something like:
"Making sure the `dir` attribute is already set as `rtl`. Notice that the viewer might set the correct direction by itself on the `body` or on an ancestor of the component. If not, we can always set it on the root of the component by ourselves."

I'll keep the example as:
```
const MyApp = () => <div dir="rtl">Hello World!</div>;
```

But will mention in the comment above that the viewer also can set it.

WDYT?

docs/RTL.md
nitayneeman: I can mention `::before` and `::after` but not sure we should dive into details too much... 
Maybe if you have some good examples we can point to them?

Regarding the small fix - I'll correct it

nitayneeman (commented):
We can reference this guide in USAGE at the moment, but I think we should create a "Usage"  directory (or something like that) and place there different guides by subjects

docs/RTL.md
jonathanadler: Excellent 🙌

jonathanadler:
> We can reference this guide in USAGE at the moment, but I think we should create a "Usage" directory (or something like that) and place there different guides by subjects

Cool sound good 👌

docs/RTL.md
nitayneeman: @jonathanadler I pushed the change - take a look

jonathanadler (approved):
Excellent!

PR #487, opened 2020-11-22
docs: simulate focus ring

nitayneeman:
This PR simulates the viewer's focus-ring by adding the relevant style globally.

Basically it does what TB does, but the following selector sounds redundant for the StoryBook environment:
`#SITE_CONTAINER.focus-ring-active :not(.has-custom-focus):not(.ignore-focus):focus, #SITE_CONTAINER.focus-ring-active :not(.has-custom-focus):not(.ignore-focus):focus ~ .wixSdkShowFocusOnSibling': {`

It also affects Storybook elements, but I don't think it's an issue. We can restrict it somehow to the component preview if it bothers...

### DIFF
--- a/.storybook/addons/FocusRingToggle/index.js
+++ b/.storybook/addons/FocusRingToggle/index.js
@@ -1,0 +1,21 @@
+import addons, { makeDecorator } from '@storybook/addons';
+import * as React from 'react';
+import { ADDON_ID, ADDON_TITLE } from './shared';
+import { useState } from 'react';
+import FocusRing from '../../../stories/helperComponents/FocusRing';
+
+const FocusRingToggle = makeDecorator({
+  name: ADDON_TITLE,
+  parameterName: ADDON_TITLE,
+  allowDeprecatedUsage: true,
+  wrapper: (getStory) => {
+    const [active, setActive] = useState(false);
+    const channel = addons.getChannel();
+
+    channel.on(ADDON_ID + '/change', (params) => setActive(params.active));
+
+    return <FocusRing active={active}>{getStory()}</FocusRing>;
+  },
+});
+
+export default FocusRingToggle;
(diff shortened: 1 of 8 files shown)


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_487/

jonathanadler (approved):
Are we sure we don't need the `!important` and `z-index` that the viewer uses?

nitayneeman:
> Are we sure we don't need the `!important` and `z-index` that the viewer uses?

I haven't managed to find the case it's relevant - but we can add them to be completely sure.
I'll add them with a comment above saying we simulate viewer behavior.

.storybook/addons/FocusRingToggle/register.js
```diff
@@ -0,0 +1,35 @@
...
+
+  return (
+    <div
+      style={{
+        margin: 'auto 15px',
+        cursor: 'pointer',
+        userSelect: 'none',
+        padding: '3px 8px',
+        backgroundColor: active ? '#4EB7F5' : '#eee',
+        color: active ? '#444' : '#999',
```
nitayneeman, line 19: I took the same styles from WSR, but we can style differently later if we want

.storybook/addons/FocusRingToggle/index.js
```diff
@@ -0,0 +1,21 @@
...
+
+const FocusRingToggle = makeDecorator({
+  name: ADDON_TITLE,
+  parameterName: ADDON_TITLE,
+  allowDeprecatedUsage: true,
+  wrapper: (getStory) => {
+    const [active, setActive] = useState(false);
+    const channel = addons.getChannel();
+
+    channel.on(ADDON_ID + '/change', (params) => setActive(params.active));
```
nitayneeman, line 15: The problem might be here.

The events are emitted asynchronously and this, I think, makes the component to render the initial active value even after that the event is emitted.

jonathanadler (approved):
Very nice!

PR #488, opened 2020-11-22
Typography documentation page

jonathanadler:
Added a typography documentation, which explains how typography is dealt with in the library.
It has two parts to it:
1. For consumers: what fonts do they get with the components, and how to use the Text component.
2. For contributors, how to deal with fonts when creating new components

Note that the description is for the current implementation of the `Text` component, which we would like to improve.
So this doc will probably change soon.


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_488/

jonathanadler:
you can see the README here: https://bo.wix.com/pages/wix-ui-tpa.pr_488/?path=/story/getting-started--typography

sivanharel (approved):
Wow, it's so informative!
Just one comment, in some of the components the designers define the font size as fixed but the font-family and the font-weight as overridable. 
So maybe we should add an example for this case and explain the correct way to do that.

docs/Typography.md
```diff
@@ -0,0 +1,65 @@
...
+The `<Text/>` component receives a `typography` prop, which its values are mapped to the font according to the following table:
+
+| Typography  | Font theme | Size | Line Height |
+| ----------  | -----      | ----- | ----- |
+| smallTitle  | Page-title | Desktop: 24px <br /> Mobile: 20px | Desktop: 1.33 <br /> Mobile: 1.4 |
+| largeTitle  | Heading-M  | Desktop: 32px <br /> Mobile: 24px | Desktop: 1.25 <br /> Mobile: 1.33 |
+| listText    | Body-M     | Desktop: 16px <br /> Mobile: 14px | Desktop: 2 <br /> Mobile: 1.72 |
+| runningText | Body-M     | Desktop: 16px <br /> Mobile: 14px | Desktop: 1.5 <br /> Mobile: 1.42 |
+
+It is rendered as a `span` by default, but this can be changed using the `tagName` prop
```
shlomitc, line 21: Since these are all APIs of the Text component, I'd simply put it there and link to it.
Otherwise, it will always require maintaining in both docs

docs/Typography.md
```diff
@@ -0,0 +1,65 @@
...
+    </section>  
+  );
+}
+```
+
+
+## When contributing to the library
+When creating a new component that renders text, we should in most cases use the `<Text/>` component with the correct props.  
+In some cases we might need to use a typography that isn't defined in the `<Text/>` component.  
+In this case, we rely on the style processor's syntax, to ensure we get the correct font from the viewer.
```
shlomitc, line 42: everything that is contribution related should go to a contribution file(s).
As a consumer of a library, I'd like to focus on how I consume it and not how to work on it.

docs/Typography.md
```diff
@@ -0,0 +1,65 @@
...
+
+
+### How it works
+We use the style processor's syntax to get our components wired correctly:
```
shlomitc, line 46: add a link to the style processor feature that does it or to the readme of it

docs/Typography.md
```diff
@@ -0,0 +1,65 @@
...
+
+### Mapping design spec to style-processor syntax
+The design system has a different naming convention for typography, as can be seen [here](https://zeroheight.com/7sjjzhgo2/p/00b9e9-typography), but the mapping is pretty simple:
+
+ 
+| Design Name | Style Processor Syntax |
+| ----------- | ---------------------- |
+| H2 | Page-title |
+| P2 | Body-M |
+| P3 | Body-S |
```
shlomitc, line 65: is this contribution or usage? put it in the current place accordingly

docs/Typography.md
jonathanadler: The combination of your comments effectively suggest simply removing this document 😄 
What do you think should appear in this document?

shlomitc: Yes, kind of.
I think that having a well structured `Text` component story page that includes also the [ZeroHeight Specs](https://zeroheight.com/7sjjzhgo2/p/00b9e9-typography) and a section about the style processor mapping to the design name will be enough

jonathanadler: The problem at the moment is that the `Text` component is a bit out-of-date, so we can't rely on it too much.
I would like to update it soon.
But until then, I think we need a general guide.

I can move the contribution content to the contribution guide, and remove this doc.
Although I think it's also good for consumers to know how this works.

shlomitc: ok

docs/COMPONENT_GUIDELINES.md
```diff
@@ -29,3 +29,80 @@ So solutions like `@media-queries` might not work, and `window` shouldn't be dir
 
 * `wix-ui-tpa` are rendered for Wix's users-of-users, and so they must comply with web accessibility
  principles.
+ 
+ ## Typography
+ When creating a new component that renders text, we should in most cases use the `<Text/>` component with the correct props and style.
```
shlomitc, line 34:
```suggestion
 When creating a new component that renders text, simply use the `<Text/>` component. It already contains pre-defined style variants. simply follow it's API in the documentation page
```

docs/COMPONENT_GUIDELINES.md
```diff
@@ -29,3 +29,80 @@ So solutions like `@media-queries` might not work, and `window` shouldn't be dir
 
 * `wix-ui-tpa` are rendered for Wix's users-of-users, and so they must comply with web accessibility
  principles.
+ 
+ ## Typography
+ When creating a new component that renders text, we should in most cases use the `<Text/>` component with the correct props and style.  
+ In some cases we might need to use a typography that isn't defined in the `<Text/>` component.  
+ In this case, we rely on the style processor's syntax, to ensure we get the correct font from the viewer.
```
shlomitc, line 36: Do you mean here for cases that you wish to connect the text to the user's choice? if so, then I'd simply write:
```suggestion
If a specific text needs to be configurable (e.g. by a user, or for a different variation), you are advised to use the `style-processor` syntax for customing fonts (see example below)
```

And then add a simple example of how to extend the component

docs/COMPONENT_GUIDELINES.md
```diff
@@ -29,3 +29,80 @@ So solutions like `@media-queries` might not work, and `window` shouldn't be dir
...
+ The platform (viewer/editor) references the different fonts by a theme name, such as `Body-M` for paragraph text, and `Heading-L` for titles.  
+ To use these fonts we write for instance:
+ ```css
+:import {
+  -st-from: "src/common/formatters.st.js";
+  -st-named: font;
+}
+
+.myClass {
+  font: font({theme: 'Body-M', size: '16px', weight: 'bold'});
```
shlomitc, line 50: don't you need to extend the `<Text/>`  component?

docs/COMPONENT_GUIDELINES.md
```diff
@@ -29,3 +29,80 @@ So solutions like `@media-queries` might not work, and `window` shouldn't be dir
...
 * `wix-ui-tpa` are rendered for Wix's users-of-users, and so they must comply with web accessibility
  principles.
+ 
+ ## Typography
+ When creating a new component that renders text, we should in most cases use the `<Text/>` component with the correct props and style.  
+ In some cases we might need to use a typography that isn't defined in the `<Text/>` component.  
+ In this case, we rely on the style processor's syntax, to ensure we get the correct font from the viewer.
+ 
+ 
+ ### How it works
```
shlomitc, line 39: I find this section a bit confusing.
As a consumer - I'd expect the simple example of how to extend the Text component and right afterward a link to the style processor repo. Also, consider pointing to https://github.com/wix-incubator/tpa-style-webpack-plugin.

If you want to have a section that talks about how the style-processor works, I wouldn't mix it here in the typography but have a different file

docs/COMPONENT_GUIDELINES.md
```diff
@@ -29,3 +29,80 @@ So solutions like `@media-queries` might not work, and `window` shouldn't be dir
...
+ We use the [style processor's syntax](https://github.com/wix/wix-style-processor) to get our components wired correctly:  
+ The platform (viewer/editor) references the different fonts by a theme name, such as `Body-M` for paragraph text, and `Heading-L` for titles.  
+ To use these fonts we write for instance:
+ ```css
+:import {
+  -st-from: "src/common/formatters.st.js";
```
shlomitc, line 45: This is internally in the library. how do we expect consumers to use it? is it even exported?

docs/COMPONENT_GUIDELINES.md
jonathanadler: Unfortunately the `Text` element is old and can't be easily extended.
I prefer to use the current solution, and once we fix the Text element we will update this doc

docs/COMPONENT_GUIDELINES.md
jonathanadler: Not exactly what I meant

docs/COMPONENT_GUIDELINES.md
jonathanadler: I wish, but it doesn't work well at the moment, and it's not optimized for better performance.

docs/COMPONENT_GUIDELINES.md
jonathanadler: This is part of the contribution guide, and not a consumer guide.
Formatters are exported, but this is an example for internal use

docs/COMPONENT_GUIDELINES.md
jonathanadler: This isn't a consumer doc.
I moved the original doc in the PR to here, to the contribution documentation

jonathanadler: the whole purpose of this task was to clarify the way we write fonts when contributing

PR #489, opened 2020-11-23
Make rtl toggle work with the TPAComponentsProvider

jonathanadler:
create storyComponent for supporting the rtl toggle
use it in existing component stories
and add it to the generator

### DIFF
--- a/stories/helperComponents/storyComponent.tsx
+++ b/stories/helperComponents/storyComponent.tsx
@@ -1,0 +1,40 @@
+import * as React from 'react';
+import { TPAComponentsProvider } from '../../src/components/TPAComponentsConfig';
+
+// tslint:disable:react-hooks-nesting
+export const storyComponent = Component => {
+  return props => {
+    const [rtl, setRtl] = React.useState(false);
+    const rootRef = React.useRef<HTMLDivElement>();
+
+    React.useEffect(() => {
+      if (rootRef && rootRef.current) {
+        const observer = new MutationObserver(mutationsList => {
+          mutationsList.map(mutation => {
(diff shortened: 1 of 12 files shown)


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_489/

stories/helperComponents/storyComponent.tsx
```diff
@@ -0,0 +1,40 @@
+import * as React from 'react';
+import { TPAComponentsProvider } from '../../src/components/TPAComponentsConfig';
+
+// tslint:disable:react-hooks-nesting
+export const storyComponent = Component => {
```
nitayneeman, line 5: Just a suggestion for the future - we can also pass `options` prop (besides `Component`) to control the mobile enable/disable. Maybe could be useful

nitayneeman (approved):

stories/helperComponents/storyComponent.tsx
jonathanadler: interesting 🤔 
Maybe it should be part of the `props` received by the component in the line below (line 6)?
So it will look like this:
```jsx
const { mobile, ...rest } = props;
return (
      <div ref={rootRef}>
        <TPAComponentsProvider value={{ mobile, rtl }}>
          <Component {...rest} />
        </TPAComponentsProvider>
      </div>
    );
```
wdyt @nitayneeman ?

nitayneeman: Sounds good as well 👍

Issue #496, opened 2020-11-24
[TextArea] Accessibility - not on full field size


shlaskt:
Issue-  when using `tab` to navigate (Accessibility) the border is on the inner textarea of the `TextArea`, and it looks bad.
![image](https://user-images.githubusercontent.com/45847538/100114761-6debf480-2e7a-11eb-9c4c-a98814222484.png)

Reproduce:
Goto https://gilada.wixsite.com/sunny-falafel/mainpage, add dish and click on the cart from the left.
Navigate with the `tab` to 'Add Order Notes' option- and click on it.

The blue border should cover all the area.
I guess it is related to the padding 8px from top and bottom of the textarea that you add.

shlaskt:
@jonathanadler Is there an ETA for a fix?

jonathanadler:
@shlaskt we don't have an ETA at the moment.
But you're more than welcome to contribute a fix for this.

shlaskt:
@jonathanadler https://github.com/wix/wix-ui-tpa/pull/506

jonathanadler:
> @jonathanadler #506

Shpitz!

PR #506, opened 2020-12-02
[wix-ui-tpa] TextArea: fix accessibility- border all frame

shlaskt:
Issue: https://github.com/wix/wix-ui-tpa/issues/496
Fix- move padding to textarea

### DIFF
--- a/src/components/TextArea/TextArea.st.css
+++ b/src/components/TextArea/TextArea.st.css
@@ -73,7 +73,6 @@
     box-sizing: border-box;
     display: flex;
     position: relative;
-    padding: 8px 0;
     width: 280px;
     height: 92px;
 }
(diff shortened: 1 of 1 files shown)


jonathanadler:
It looks great, just one small thing.
In order for the build and tests to run, you'll need to create a signed commit.
You can read about how to do it here: https://github.com/wix/wix-style-react/blob/master/docs/contribution/CREATE_PR.md
and then just make some dummy commit, push, and it will run.

Thanks for the contribution!

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_506/

jonathanadler:
@shlaskt 
First of all, thanks again!

Second, as you probably noticed, the visual tests are failing, since this is a breaking change.

I'll check with our designers if this change is ok, 
If not, we might need to manually set the focus-ring, which means manually maintain a `focus`-ed state

shlaskt:
@jonathanadler yes I see that it's needed the bottom padding with the current design, and now the text cuts at the last line for long text. Unfortunately, I didnt find any solution to handle it with the focus ring.
The second option sounds good also.

PR #519, opened 2020-12-08
feat(RadioButton, Checkbox) - add disabled mode state for theme box and some bug fixes.

sivanharel:
This PR solves some bugs that @barboaron from the RichContentEditor team found.
In addition, I added some styles for the Box theme on disabled mode according to this spec: https://zeroheight.com/7sjjzhgo2/p/05a744-05-selections/b/285e96.

The bugs that were fixed:
1. In the checkbox component: fix the box border overlapping issue.
2. In the RadioButton component: Override the styling that comes from the wix-ui-core.

### DIFF
--- a/src/components/RadioButton/RadioButton.st.css
+++ b/src/components/RadioButton/RadioButton.st.css
@@ -24,9 +24,25 @@
 
 /*Overrides*/
 :vars {
+    /*
+        The radio button text color
+        @default color-5
+    */
     TextColor: --overridable;
+    /*
+        The border color. FYI, An opacity of 0.6 will be added.
+        @default color-5
+    */
     BorderColor: --overridable;
+    /*
+        The color of the icon. This color would be used as the box's border color as well.
+        @default color-8
+    */
     IconColor:  --overridable;
+    /*
+        The box's background color. FYI, This value won't be affected on disabled mode.
+        @default color-1
+    */
     BackgroundColor:--overridable;
 }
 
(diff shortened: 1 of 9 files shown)


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_519/

src/components/RadioButton/RadioButton.st.css
```diff
@@ -162,6 +178,19 @@
     box-sizing: border-box;
 }
 
+.root:box:disabled .wrapper {
+    /* overriding the styling from wix-ui-core */
+    filter: none;
+    opacity: 1;
+    /*******/
```
nitayneeman, line 185: I wonder if we want to provide a way to disable these properties in ui-core instead of overriding them here? 🤔

nitayneeman (approved):
Looks good!

I added a little question but I think we can merge, unless we do want to add the visual before?

nitayneeman:
> Looks good!
> 
> I added a little question but I think we can merge, unless we do want to add the visual before?

Anyway I think we should visual tests for these cases

src/components/RadioButton/RadioButton.st.css
sivanharel: Actually, I think that these css properties should be removed from the core and everyone should use its own design but it's breaking changes.
I think that it doesn't worth the effort, I already waste too much time on it.

PR #520, opened 2020-12-08
counter  increase buttons click area

mariakwix:
Counter buttons clickable are is currently very small. can we fix this like so?

### DIFF
--- a/src/components/Counter/Counter.st.css
+++ b/src/components/Counter/Counter.st.css
@@ -59,6 +59,10 @@
     padding: 6px;
     border: none;
     cursor: pointer;
+    flex:1;
+    height:100%;
+    display: flex;
+    justify-content: center;
 }
 
 .btn path {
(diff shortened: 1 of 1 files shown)


jonathanadler:
hey @mariakwix 👋 

Please sign your commits in order for the build to run 🙏 

https://github.com/wix/wix-style-react/blob/master/docs/contribution/CREATE_PR.md#sign-commits

Thanks!

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_520/

sivanharel (commented):
Hi @mariakwix,
First of all, thank you for contributing to the library.

There is an issue with your solution, when it's a big number it overlaps the text so the flex-shrink: 1 couldn't be on the button and the icons are now closer to the text

I talked with our designer to ask her about an alternative solution and she said that the current size of the button which is 24px *24px is bigger than the icon itself and should be enough.
If there is a specific issue with that and users complain about it, you could ask your designer to talk to our designer (@michalvardi )

jonathanadler:
Closing this for now

PR #526, opened 2020-12-14
Dialog: fix not wired state

jonathanadler:
I've noticed that my previous work on the Dialog had a bug:

the `wiredToSiteColors` prop only worked when consumers didn't actively pass colors (using st-mixin).
So for instance, if you go to the Dialog settings panel, in our live storybook site, you'll see that once you select colors to override, the toggle doesn't take any effect.

This PR fixes this issue.

### DIFF
--- a/src/components/Dialog/Dialog.st.css
+++ b/src/components/Dialog/Dialog.st.css
@@ -63,29 +63,21 @@
 }
 
 /* New style overrides capabilities */
-.skin-fixed {
-    -st-extends: root;
-}
-
-.skin-wired {
-    -st-extends: root;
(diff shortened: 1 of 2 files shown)


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_526/

nitayneeman (approved):
Looks good to me.

This way is simplifying the API, so I think it's better than 4 variables for only 2 colors (unless we want to let the consumer render different colors in the same state but I don't think this is a use-case)

PR #529, opened 2020-12-14
fix: prevent ios zoom in FloatingDropdown

jjaskulis:
ios auto zooms if clicking on input element which has font-size smaller than 16px, this fixes this.

### DIFF
--- a/src/components/FloatingDropdown/FloatingDropdownBase.st.css
+++ b/src/components/FloatingDropdown/FloatingDropdownBase.st.css
@@ -130,6 +130,7 @@
     width: 100%;
     height: 100%;
     color: transparent;
+    font-size: 16px;
 }
 
 .root:mobile:disabled select {


jonathanadler:
Hey @jjaskulis 
Thanks for the contribution!

First, in order for the build to run, you'll need to sign your commits.
You can follow the guide here: https://github.com/wix/wix-style-react/blob/master/docs/contribution/CREATE_PR.md#sign-commits

Second, can you share a screenshot, or live site, so that we'll be able to understand the issue a bit better?

Third, I believe the font size of the FloatingDropdown should be wired to the site's and the app's settings.
Maybe instead of a specific font size it should be inherited?

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_529/

jjaskulis:
Dec-14-2020 14-43-09

@jonathanadler I think it's pretty clear from this gif, as you can see when clicking on floating dropdown the screen is zoomed-in on iOS. On mobile you show native select, so setting font-size has not affect on how the element is displayed visually. Font-size should be hardcoded, as any value less than 16px will result in this auto zoom behaviour on iOS.
Here is another gif, only this time with the font-size fix:

Dec-14-2020 14-47-54

jonathanadler:
@jjaskulis 
Thanks for the gifs 👌 

So actually I've encountered a similar issue.
I think that what causes this, is the Button component mobile styling, which the FloatingDropdown component is using internally.
The font size should be 16px by default, but it can also be overridden.
However, the Button mobile sets 14px on mobile devices, and this is overriding the FloatingDropdown's definitions.
So in order to fix this we'll need to disable the Button's mobile style, which sets the 14px font size.

I think we'll need to make a bigger and more robust change in order to fix this issue.

I will try to fix this issue (which is also effecting the Dropdown component) soon. 
So we can close this PR for now.

If you'd like you can add a github issue so that you could follow the progress.

jjaskulis:
@jonathanadler 
got it. is there a way to override the select fontsize now until its fixed?

jonathanadler:
> @jonathanadler
> got it. is there a way to override the select fontsize now until its fixed?

Hey, sorry for the late response.

You can use the Stylable variable `MainTextFont` that the component is exposing.
You can see it in use here:
https://github.com/wix/wix-ui-tpa/blob/master/src/components/FloatingDropdown/docs/FloatingDropdownWiringExample.st.css

PR #533, opened 2020-12-16
feat(DatePicker): add to Working in Progress

sivanharel:
# ✨ Pull Request

### 💁 Description

Implement new component - DatePicker
Let's merge it to the WIP section and then create new PRs for each feature that was requested from Booking.

...

### 💭 Motivation

...

### ☝️ TODOs 

...

### 👀  PR is ready for review 

- [ ] Ready for review


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_533/

src/components/DatePickerInput/DatePickerInput.st.css
```diff
@@ -170,268 +124,28 @@
...
-/*        CalendarDaySelectedBGColorHover applyOpacity(value(CurrentSelectedDayBGColor), 0.7),*/
-/*    */
-/*        CalendarDayDisabledColor value(CurrentDisabledDayColor),*/
-/*        CalendarDayOutsideColor value(CurrentDisabledDayColor),*/
-/*    */
-/*        !**********************!*/
-/*    );*/
-/*}*/
+.overrideStyleParams .wsrDatePicker::calendar {
+    -st-mixin: datePickerOverrideStyleParams(
```
sivanharel, line 137: Doesn't work after I changed from:
 -st-mixin: calendar()
to 
-st-mixin: datePickerOverrideStyleParams()

sivanharel: Reverted for now.

sivanharel:
@shlomitc, @jonathanadler.

Hi,

So thanks to @nitayneeman that added the Perfer analysis.
I was able to test the bundle sizes of the DatePicker and the DatePickerInput.

Here are the results:
**DatePickerInput: 208.614kb** while in WSR the DatePicker component is **132kb.**
**DatePicker: 186.388kb** while in WSR the Calendar component is **117kb.**

It seems like the bundle size was increased a lot!
In the DatePickerInput, it might be also related to the fact that the TPA Input component is added to the bundle as well.
In the DatePicker it might be related to the fact that currently, I am not using the overrideStyleParams because it wasn't worked.

shlomitc:
> Maybe it is also related to the fact the TPA is not supporting ESModules and I haven't changed the locale imports here? https://github.com/wix/wix-style-react/blob/master/src/LocaleUtils.js#L2-L22

yep, that's the reason

src/components/DatePicker/DatePicker.st.css
```diff
@@ -0,0 +1,581 @@
...
+        @default color-1
+    */
+    BackgroundColor: --overridable;
+    /*
+        The color of the calendar navigation arrows
+        @default color-5
+    */
+    NavigationArrowsColor: --overridable;
+    /*
+        The font of the calendar month snd year caption. The font size is 16px and could not be overridden.
```
jonathanadler, line 53: Nice!
I like that you added the disclaimer that the font size is fixed.
We can probably add this as a special `@fixed` statement, or something like that

src/components/DatePicker/DatePicker.st.css
```diff
@@ -0,0 +1,581 @@
...
+    /*
+        The font color of the calendar's weekdays.
+        @default color-8
+    */
+    WeekdayTextColor: --overridable;
+    /*
+        The font of a specific date. The font size is 14px and could not be overridden.
+        @default {theme: 'Body-M', lineHeight: '20px'}
+    */
+    DayFont: --overridable;
```
jonathanadler, line 76: From what I understood from Leigh, this should be overridable, no?

src/components/DatePicker/DatePicker.st.css
```diff
@@ -0,0 +1,581 @@
...
+
+.calendar::header {
+    padding: 28px value(sidePadding) 0;
+    background-color: value(CurrentBackgroundColor);
+}
+
+.calendar::header::arrow {
+    color: value(CurrentNavigationArrowsColor);
+    width: 24px;
+    height: 24px;
```
jonathanadler, line 153: If font will be overridable, we might need to have the arrow dimensions be set as `em`'s

src/components/DatePicker/DatePicker.st.css
```diff
@@ -0,0 +1,581 @@
...
+}
+
+.calendar::day:outside, .calendar::day:disabled {
+    color: value(CurrentDisabledDayColor);
+}
+
+/* todo: need to check if it's really override the viewer focus ring and if we should remove it..  */
+.calendar::day:focus::after {
+    border-radius: value(CurrentSelectedDayBorderRadius);
+    box-shadow: 0 0 0 1px #ffffff, 0 0 0 3px #116dff !important;
```
jonathanadler, line 235: You can see the correct implementation (which is very similar) [here](https://github.com/wix/wix-ui-tpa/blob/2eb09078873af6f8164f0123626139b62c7a7bb7/src/components/RadioButton/RadioButton.st.css#L96).

It would be nice to move it to a mixin in a common file such as palette.st.css

src/components/DatePicker/DatePicker.st.css
```diff
@@ -0,0 +1,581 @@
...
+
+.calendar::header::monthDropdown::dropdown::list::selectableOption:selected:hover,
+.calendar::header::yearDropdown::dropdown::list::selectableOption:selected:hover {
+    background-color: value(DropdownOptionSelectedAndHoverBGColor);
+}
+
+.calendar::header::monthDropdown::dropdown::list::selectableOption::title,
+.calendar::header::yearDropdown::dropdown::list::selectableOption::title {
+    color: value(CurrentDayColor);
+    font: font("{theme: 'Body-M', size: '14px', lineHeight: '20px'}");
```
jonathanadler, line 319: Not critical at all, but is there a reason why the font is hard-coded here?
Is this cause it's used only here?
I wish we would have IDE support, so that we could really use consts effectively

src/components/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,122 @@
...
+  | 'th'
+  | 'cs';
+
+export interface DatePickerProps extends TPAComponentProps {
+  /** The selected date */
+  value?: Date | string;
+  /** Callback function called with a Date or a Range whenever the user selects a day in the calendar */
+  onChange(selectedDays: string | Date): void;
+  /** DatePicker instance locale */
+  locale?: LanguageType | { distanceInWords?: {}; format?: {} };
```
jonathanadler, line 36: Does this allow to pass locales from outside?

src/components/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,122 @@
...
+  'aria-labelledby'?: string;
+}
+
+interface DefaultProps {
+  locale: LanguageType | { distanceInWords?: {}; format?: {} };
+  excludePastDates: boolean;
+  filterDate(date: Date): boolean;
+  firstDayOfWeek: 0 | 1 | 2 | 3 | 4 | 5 | 6;
+  showMonthDropdown: boolean;
+  showYearDropdown: boolean;
```
jonathanadler, line 63: Maybe `LanguageType | { distanceInWords?: {}; format?: {} };` and `0 | 1 | 2 | 3 | 4 | 5 | 6` could be extracted to types? that way we won't get repetition

meaning:
```ts
type Locale = LanguageType | { distanceInWords?: {}; format?: {} };
type FirstDayOfWeek = 0 | 1 | 2 | 3 | 4 | 5 | 6;
```

src/components/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,122 @@
...
+}
+
+/** The DatePicker allows a user to select a specific date. */
+export class DatePicker extends React.Component<DatePickerProps> {
+  static displayName = 'DatePicker';
+  static defaultProps: DefaultProps = {
+    locale: 'en',
+    excludePastDates: false,
+    filterDate: () => true,
+    firstDayOfWeek: 1,
```
jonathanadler, line 73: I think `firstDayOfWeek` should be `0` by default, no?

src/components/DatePicker/DatePicker.visual.tsx
```diff
@@ -0,0 +1,54 @@
...
+    snap(
+      'FilterDate - Prior for today dates (includes today)',
+      <DatePicker
+        value={new Date('2019/12/15')}
+        filterDate={date => date <= new Date('2019/12/15')}
+        onChange={() => {}}
+      />,
+    );
+    snap(
+      'FilterDate - feature dates only (includes today)',
```
jonathanadler, line 21:
```suggestion
      'FilterDate - future dates only (includes today)',
```

src/components/DatePicker/DatePicker.visual.tsx
```diff
@@ -0,0 +1,54 @@
...
+    snap(
+      'Months and years dropdown',
+      <DatePicker
+        value={new Date('2019/12/15')}
+        showMonthDropdown
+        showYearDropdown
+        onChange={() => {}}
+      />,
+    );
+  });
```
jonathanadler, line 53: Let's add a visual test for a selected day, and for the focus-ring

src/components/DatePicker/DatePicker.driver.ts
```diff
@@ -0,0 +1,159 @@
...
+    pressRightArrow: async () => WSRCalendarDriver.pressRightArrow(),
+    /**
+     * Get the selected day
+     */
+    getSelectedDay: async () => WSRCalendarDriver.getSelectedDay(),
+    /**
+     * Trigger key down event
+     * @param key
+     */
+    triggerKeyDown: async key => WSRCalendarDriver.triggerKeyDown({ key }),
```
jonathanadler, line 147: I'm not sure I understand what `triggerKeyDown` is for 🤔

src/components/DatePicker/DatePicker.driver.ts
```diff
@@ -0,0 +1,159 @@
...
+     */
+    getSelectedYear: async () => WSRCalendarDriver.getSelectedYear(),
+    /**
+     *  Returns the text of the focused day or `null` if there is no focused day
+     */
+    getFocusedDay: async () => WSRCalendarDriver.getFocusedDay(),
+    /**
+     * Press left arrow
+     */
+    pressLeftArrow: async () => WSRCalendarDriver.pressLeftArrow(),
```
jonathanadler, line 134: What's the difference between `clickOnPrevMonthButton` and this?
We may have talked about this already and I forgot 😄 
Is this because in RTL the sides have different meanings?

Maybe we don't need to expose both types of methods.

In any case, and not critical at all, but maybe we want `click` instead of `press` 🤔 ?
It's just that `press` usually refers to mobile.
Again, not critical

src/components/DatePicker/docs/DatePickerWiringExample.tsx
```diff
@@ -0,0 +1,13 @@
+import * as React from 'react';
+import { DatePicker } from '../';
+import { classes } from './DatePickerWiringExample.st.css';
+
+export const DatePickerWiringExample = () => {
+  return (
+    <DatePicker
+      className={classes.root}
+      value={new Date()}
+      onChange={() => {}}
```
jonathanadler, line 10: Let's add a state to this example, so that the `onChange` will actually change the selected date

src/components/DatePickerInput/DatePickerInput.st.css
```diff
@@ -170,268 +124,28 @@
...
-/*        CalendarDaySelectedTextColor  value(CurrentSelectedDayColor),*/
-/*        CalendarDaySelectedBGColorHover applyOpacity(value(CurrentSelectedDayBGColor), 0.7),*/
-/*    */
-/*        CalendarDayDisabledColor value(CurrentDisabledDayColor),*/
-/*        CalendarDayOutsideColor value(CurrentDisabledDayColor),*/
-/*    */
-/*        !**********************!*/
-/*    );*/
-/*}*/
+.overrideStyleParams .wsrDatePicker {
```
jonathanadler, line 136: I still think it would be nice to try passing the calendar className to the datepicker component inside the datepickerinput, and then see if the `overrideStyleParams` will work

jonathanadler (commented):
Very very nice 👏 👏 👏 
Left some very small comments

src/components/DatePicker/DatePicker.st.css
sivanharel: yes, so we have just added a task that includes it:)

src/components/DatePicker/DatePicker.st.css
```diff
@@ -0,0 +1,581 @@
...
+
+    DefaultWeekdayFont: "{theme: 'Body-M', lineHeight: '20px'}";
+    DefaultWeekdayTextColor: color-8;
+
+    DefaultDayFont: "{theme: 'Body-M', lineHeight: '20px'}";
+    DeafultDayColor: color-5;
+    DefaultSelectedDayColor: color-1;
+    DefaultSelectedDayBGColor: color-8;
+    DefaultSelectedDayBorderRadius: 50%;
+    DefaultDisabledDayColor: color-3;
```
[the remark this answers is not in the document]
sivanharel: yes, maybe it's a good idea. But do you think I should change it now? Cause it will be different from the other components.

src/components/DatePicker/DatePicker.st.css
sivanharel: Yes, it only used once, and there is no option to override it.

src/components/DatePicker/DatePicker.tsx
sivanharel: It should allow it but I haven't checked it yet.

src/components/DatePicker/DatePicker.tsx
sivanharel: The values are from 0 to 6 (Sunday to Saturday).
So since Sunday is a day off in most countries the default is 1 which means that the week starts from Monday.

src/components/DatePicker/DatePicker.driver.ts
sivanharel: It's in order to simulate a keyboard key down event. But I will remove it.
I also wasn't sure we need it..

src/components/DatePicker/DatePicker.driver.ts
sivanharel: The `clickOnPrevMonthButton` function is for clicking on the header's arrow. The header's arrows are responsible for changing months.

The `pressLeftArrow` function is simulating pressing on the left arrow in the keyboard.
It is used in WSR to test a11y.
There are many tests that check this functionality in WSR and I think we don't need to test it again. 

Actually, there are a lot of functions in the WSR driver that are exposed which I wasn't exposed in the TPA because I think that some of them are not relevant for the users. But maybe we should remove some more..

src/components/DatePickerInput/DatePickerInput.st.css
sivanharel: Yes, I agree that we should find a solution. But let's keep it for later :) I can even create a task so we will not forget it.

src/components/DatePicker/DatePicker.visual.tsx
sivanharel: It is already tested:)
The focus is added automatically on the selected day so the focus ring is displayed on all the visual tests.

src/components/DatePicker/DatePicker.driver.ts
jonathanadler: Oh, I see, ok cool 👌 

I wonder if maybe we should change this to `pressLeftArrowKey` 🤔

src/components/DatePicker/DatePicker.st.css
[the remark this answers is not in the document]
jonathanadler: Not critical for now

src/components/DatePicker/DatePicker.tsx
jonathanadler: right right 🤦

jonathanadler (approved):
Awesome 💪

PR #535, opened 2020-12-16
remove default button priority and size

jonathanadler:
### 👀  PR is ready for review 

- [x] Ready for review

### DIFF
--- a/src/components/Button/Button.st.css
+++ b/src/components/Button/Button.st.css
@@ -105,15 +105,15 @@
     BasicBackgroundColor: color(fallback(value(MainBackgroundColor), value(BasicColor)));
 }
 
-.root:priority(basic) {
+.root, .root:priority(basic) {
     color: value(PrimaryTextColor);
     background-color: value(BasicBackgroundColor);
     border-width: value(PrimaryBorderWidth);
 }
 
-.root:priority(basic):active,
-.root:priority(basic):focus,
-.root:priority(basic):hover {
+.root:active, .root:priority(basic):active,
+.root:focus, .root:priority(basic):focus,
+.root:hover, .root:priority(basic):hover {
     background-color: color(fallback(value(HoverBackgroundColor), applyOpacity(value(BasicBackgroundColor), 0.7)));
     color: applyOpacity(value(PrimaryTextColor), 0.7);
     border-color: fallback(value(HoverBorderColor), value(PrimaryBorderColor));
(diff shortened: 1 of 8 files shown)


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_535/

sivanharel (commented):
I understand the issue and that we don't want to use long selectors in order to override the default design of the button component in the dropdowns but it makes the Button.st.css file unreadable and hard to maintain :(
Maybe we should find another way? Maybe we could talk with Tom Raviv or Barak.

jonathanadler:
> I understand the issue and that we don't want to use long selectors in order to override the default design of the button component in the dropdowns but it makes the Button.st.css file unreadable and hard to maintain :(
> Maybe we should find another way? Maybe we could talk with Tom Raviv or Barak.

🤔 
I didn't feel this change caused the file to be unreadable. 
I actually felt the opposite, since I'm removing the use of stylable.
I converted stylable to regular css in this PR

sivanharel:
> > I understand the issue and that we don't want to use long selectors in order to override the default design of the button component in the dropdowns but it makes the Button.st.css file unreadable and hard to maintain :(
> > Maybe we should find another way? Maybe we could talk with Tom Raviv or Barak.
> 
> 🤔
> I didn't feel this change caused the file to be unreadable.
> I actually felt the opposite, since I'm removing the use of stylable.
> I converted stylable to regular css in this PR

Yeah, you are right it's a trade of which component st.css file would be more readable. I mean now the dropdown is more readable but the button is a bit less readable.
I feel that it is not so intuitive for someone from outside why does the css is also added directly on the root without the size and priority states.
But ok maybe it is not so critical.

jonathanadler:
> > > I understand the issue and that we don't want to use long selectors in order to override the default design of the button component in the dropdowns but it makes the Button.st.css file unreadable and hard to maintain :(
> > > Maybe we should find another way? Maybe we could talk with Tom Raviv or Barak.
> > 
> > 
> > 🤔
> > I didn't feel this change caused the file to be unreadable.
> > I actually felt the opposite, since I'm removing the use of stylable.
> > I converted stylable to regular css in this PR
> 
> Yeah, you are right it's a trade of which component st.css file would be more readable. I mean now the dropdown is more readable but the button is a bit less readable.
> I feel that it is not so intuitive for someone from outside why does the css is also added directly on the root without the size and priority states.
> But ok maybe it is not so critical.
> I just feel that we are doing some workarounds because we don't have a better solution from stylable.

But also in the Button component, assigning the styles for the root itself is just regular css.
Isn't it clear that it's the default styling of the component?

sivanharel (approved):

PR #538, opened 2020-12-17
<Tabs />: change span to link if href is provided for better SEO

martyrdomspangle:
# ✨ Pull Request

### 💁 Description
Allow tabs to contain <a> tags for better SEO.

### 💭 Motivation
Improves SEO.

### ☝️ TODOs 
-

### 👀  PR is ready for review 

- [x] Ready for review

### DIFF
--- a/src/components/Tabs/Tab/Tab.tsx
+++ b/src/components/Tabs/Tab/Tab.tsx
@@ -11,6 +11,10 @@ export interface TabItem {
   title?: string;
   /** URL if tab is link */
   href?: string;
+  /** Target for link */
+  target?: string;
+  /** Rel for link */
+  rel?: string;
 }
 
 interface TabProps extends TPAComponentProps {
@@ -23,7 +27,7 @@ interface TabProps extends TPAComponentProps {
 
 export const Tab = React.forwardRef<HTMLLIElement, TabProps>((props, ref) => {
   const { index, item, onClick, isActive, indicateActive, className } = props;
-  const { title, href } = item;
+  const { title, href, target, rel } = item;
   const onSelectTab = () => {
     onClick(index);
   };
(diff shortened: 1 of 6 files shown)


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_538/

src/components/Tabs/Tab/Tab.tsx
```diff
@@ -9,6 +9,8 @@ export interface TabItem {
   id?: string;
   /** Title of the tab */
   title?: string;
+  /** URL if tab is link */
+  href?: string;
 }
```
jonathanadler, line 14: I just noticed that we probably need to add a `target` prop as well, so that consumers can pass `_blank` for instance.

jonathanadler (commented):
Great work!

Just left a small comment, for something we didn't think of.

In addition, can you add an example with anchors in the storybook?
You can add one here: https://github.com/wix/wix-ui-tpa/blob/36f8a066d5a54a7d4a0a34018bbf8d81ec6efb07/src/components/Tabs/docs/index.story.tsx#L109

In addition, please add a "With anchors" example here: https://github.com/wix/wix-ui-tpa/blob/36f8a066d5a54a7d4a0a34018bbf8d81ec6efb07/src/components/Tabs/docs/index.story.tsx#L41

Thanks!

src/components/Tabs/Tab/Tab.tsx
```diff
@@ -47,10 +53,16 @@ export const Tab = React.forwardRef<HTMLLIElement, TabProps>((props, ref) => {
       id={item.id}
       onClick={onSelectTab}
       onKeyDown={onKeyDown}
```
jonathanadler, line 55: In order to fix selecting the anchor using the `space` key, we need to also adjust the `onKeyDown` callback.
We just need to **not** prevent default if tab has an anchor.

jonathanadler (approved):
Great work!

I added a comment to fix the issue with pressing the `space` or `enter` keys in order to select the anchor.

We can also fix this in a different PR.

PR #543, opened 2020-12-20
[EE-27556] manage focus state and add focusRing if needed - wix-ui-tpa/RadioButton component

urig-wix:
[EE-27556] manage focus state and add focusRing if needed - wix-ui-tpa/RadioButton component

### DIFF
--- a/src/components/RadioButton/RadioButton.driver.ts
+++ b/src/components/RadioButton/RadioButton.driver.ts
@@ -1,4 +1,3 @@
-import { Simulate } from 'react-dom/test-utils';
 import {
   BaseUniDriver,
   baseUniDriverFactory,
@@ -11,7 +10,7 @@ export interface RadioButtonDriver extends BaseUniDriver {
   isDisabled(): Promise<boolean>;
   isFocused(): Promise<boolean>;
   clickInput(): Promise<void>;
-  blurInput(): Promise<void>;
+  getInput(): Promise<HTMLInputElement>;
 }
 
 export const radioButtonDriverFactory = (
@@ -19,11 +18,11 @@ export const radioButtonDriverFactory = (
 ): RadioButtonDriver => {
   return {
     ...baseUniDriverFactory(base),
+    async getInput() {
+      return await base.$('input').getNative();
+    },
     async clickInput() {
       return base.$('input').click();
-    },
-    async blurInput() {
-      return Simulate.blur(await base.$('input').getNative());
     },
     async isChecked() {
       return (await base.attr(RADIOBUTTON_DATA_KEYS.Checked)) === 'true';
--- a/src/components/RadioButton/RadioButton.driver.ts
+++ b/src/components/RadioButton/RadioButton.driver.ts
@@ -19,7 +19,6 @@ export interface RadioButtonDriver extends BaseUniDriver {
 export const radioButtonDriverFactory = (
   base: UniDriver,
 ): RadioButtonDriver => {
-
   return {
     ...baseUniDriverFactory(base),
     async clickInput() {
@@ -38,7 +37,7 @@ export const radioButtonDriverFactory = (
       return (await base.attr(RADIOBUTTON_DATA_KEYS.Focused)) === 'true';
     },
     async isContainsFocusRing() {
-      return await base.$$(`.${style.classes.focused}`).count() > 0;
-    }
+      return (await base.$$(`.${style.classes.focused}`).count()) > 0;
+    },
   };
 };
--- a/src/components/RadioButton/RadioButton.driver.ts
+++ b/src/components/RadioButton/RadioButton.driver.ts
@@ -19,7 +19,7 @@ export const radioButtonDriverFactory = (
   return {
     ...baseUniDriverFactory(base),
     async getInput() {
-      return await base.$('input').getNative();
+      return base.$('input').getNative();
     },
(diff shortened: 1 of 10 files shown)


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_543/

src/components/RadioButton/RadioButton.driver.ts
```diff
@@ -1,25 +1,43 @@
...
     },
     async isDisabled() {
       return (await base.attr(RADIOBUTTON_DATA_KEYS.Disabled)) === 'true';
     },
+    async isFocused() {
+      return (await base.attr(RADIOBUTTON_DATA_KEYS.Focused)) === 'true';
+    },
+    async isContainsFocusRing() {
+      return await base.$$(`.${style.classes.focused}`).count() > 0;
+    }
```
jonathanadler, line 41: We prefer to avoid testing css.
In this case it's best to just use a visual test, and see that the focus ring is working.
So this method can be removed

src/components/RadioButton/RadioButton.spec.tsx
```diff
@@ -24,16 +24,40 @@ describe('RadioButton', () => {
...
+    expect(await driver.isContainsFocusRing()).toBeFalsy();
+  });
+
+  it('should not have focus ring after blur', async () => {
+    const driver = createDriver(<RadioButton withFocusRing {...defProps} />);
+    await driver.clickInput();
+    expect(await driver.isContainsFocusRing()).toBeTruthy();
+    await driver.blurInput();
+    expect(await driver.isContainsFocusRing()).toBeFalsy();
+  });
```
jonathanadler, line 59: I would move these tests to an `e2e` file.
Instead of checking that the focus ring exists, we should just take a screenshot.

you can use `eyes.it` for this purpose.
You can see an example in [AddItem.e2e.ts](https://github.com/wix/wix-ui-tpa/blob/master/src/components/AddItem/AddItem.e2e.ts)
For the RadioButton, you'll need to use the data-hook from [here](https://github.com/wix/wix-ui-tpa/blob/36f8a066d5a54a7d4a0a34018bbf8d81ec6efb07/src/components/RadioButton/docs/index.story.tsx#L47)

jonathanadler: you can read more about eyes.it [here](https://github.com/wix/eyes.it)

jonathanadler (commented):
Nice 👌 

Left some small comments, but great work overall!

src/components/RadioButton/RadioButton.driver.ts
```diff
@@ -8,18 +9,30 @@ import { RADIOBUTTON_DATA_KEYS } from './dataHooks';
...
   base: UniDriver,
 ): RadioButtonDriver => {
   return {
     ...baseUniDriverFactory(base),
+    async clickInput() {
+      return base.$('input').click();
+    },
+    async blurInput() {
+      return Simulate.blur(await base.$('input').getNative());
+    },
```
jonathanadler, line 27: Oh sorry for the ping-pong, I just noticed this.
We also can't use Simulate inside the driver.

The driver is a [Unidriver](https://github.com/wix-incubator/unidriver/blob/master/core/src/index.ts), and it should support various different environments, such as `enzyme`, `puppeteer`, `protractor`, etc.

So we can't use libraries that assume something on the tests environment.
In this case SImulate is used in environments where a dom is present (unlike `puppeteer` for instance).

So, we'll need to remove this method from the driver, and simulate blur in the test file itself.
At least until we'll have support for this via Unidriver.

Again, sorry I didn't notice this earlier

src/components/RadioButton/RadioButton.e2e.ts
```diff
@@ -0,0 +1,42 @@
...
+      'Cannot find radioButton',
+    );
+
+    radioButtonElement = element(by.css(`[data-hook="${RADIOBUTTON_DATA_HOOKS.coreRadioButton}"]`));
+    expect(radioButtonElement.isDisplayed()).toBe(true);
+  });
+
+  eyes.it('should show the correct design on focus', async () => {
+    await driver.click();
+    radioButtonElement.click();
```
jonathanadler, line 35: I thought we wanted to use the `clickInput` method for this, no?

src/components/RadioButton/RadioButton.e2e.ts
```diff
@@ -0,0 +1,42 @@
...
+    expect(
+      await (await browser.driver.switchTo().activeElement()).getAttribute(
+        'data-hook',
+      ),
+    ).toBe(RADIOBUTTON_DATA_HOOKS.coreRadioButton);
```
jonathanadler, line 40: This part isn't really needed I guess.
Once the `eyes.it` test ends, it takes a screenshot of the window, and sends it to applitools.
So I guess all we need to do is click.

However, we'll probably want two tests: one for the regular theme, and another for the box theme, since they have different styling

jonathanadler (commented):
Again, great work!

Left some comments

jonathanadler (approved):
Excellent 🚀 

I'm approving, but before I merge this I'll ask for a designer review.
Hopefully tomorrow we can merge, and then we'll create a release.

Thanks again, awesome work!

PR #550, opened 2020-12-26
infra - fail storybook build for incorrect stylable usage

shlomitc:
Currently, consumers will see this warning once they start using the DatePicker component.
The reason is for incorrect style extending.
image

We will need to fix it anyway, but this PR makes sure the storybook build will fail (and not just throw warnings)

@sivanharel - please fix the style extending before merging

### DIFF
--- a/src/components/Badge/Badge.visual.st.css
+++ b/src/components/Badge/Badge.visual.st.css
@@ -6,8 +6,8 @@
 
 .palette {
   -st-mixin: Badge(
-    BadgeBgColor "color(color-11)",
-    BadgeTextColor "color(color-15)",
+    BadgeBgColor '"color(color-11)"',
+    BadgeTextColor '"color(color-15)"',
   );
 }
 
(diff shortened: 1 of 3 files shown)


jonathanadler (approved):
Awesome!

I was wondering how this can be done

sivanharel:
Nice!

Actually, I don't have all these errors, you might see them since you are using an old version of WSR.
But I found another issue thanks to this rule and fixed it :)


> Currently, consumers will see this warning once they start using the DatePicker component.
> The reason is for incorrect style extending.
> image
> 
> We will need to fix it anyway, but this PR makes sure the storybook build will fail (and not just throw warnings)
> 
> @sivanharel - please fix the style extending before merging

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_550/

PR #547, opened 2020-12-23
[EE-27556] radio button component - support children prop

urig-wix:
ZeroHeight Specs: https://zeroheight.com/7sjjzhgo2/p/05a744-05-selections/t/6514c6

PR #552, opened 2020-12-27
Radio Button - adding optional aria-label prop

urig-wix:
At [#547](https://github.com/wix/wix-ui-tpa/pull/547) the ability to use a children prop instead of label prop was added, and there is intention to deprecate the label prop in the future.

Today we use the value of label prop also for aria-label. I've exposed aria-label as an optional prop so it'd be possible to use the children prop and update 'aria-label' and not to be broken in the future when 'label' would be deprecated.


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_552/

jonathanadler (approved):
💪

PR #554, opened 2020-12-27
fix(RadioButton, Checkbox): Fix height

GuyPerkal:
This is most noticeable in the box theme height, it was 58px and now is 50px, same as in the design guide

### DIFF
--- a/src/components/RadioButton/RadioButton.st.css
+++ b/src/components/RadioButton/RadioButton.st.css
@@ -175,7 +175,7 @@
 
 .root:box .wrapper{
     display: flex;
-    padding: 16px;
+    padding: 15px 16px;
     background-color:applyOpacity(value(CurrentBackgroundColor),1);
     border: 1px solid applyOpacity(value(CurrentBorderColor), 0.6);
     cursor: pointer;
--- a/perfer.config.js
+++ b/perfer.config.js
@@ -15,7 +15,7 @@ const files = [
   ['CopyUrlButton.bundle.min.js', 38],
   ['Counter.bundle.min.js', 34],
   ['DatePicker.bundle.min.js', 187],
-  ['DatePickerInput.bundle.min.js', 209],
+  ['DatePickerInput.bundle.min.js', 210],
   ['Dialog.bundle.min.js', 11],
   ['Divider.bundle.min.js', 5],
   ['DotNavigation.bundle.min.js', 8],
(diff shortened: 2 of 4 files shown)


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_554/

src/components/RadioButton/RadioButton.st.css
```diff
@@ -158,6 +158,7 @@
     width: 100%;
     font: font(value(LabelTextFont));
     color: applyOpacity(value(CurrentTextColor),1);
+    line-height: 16px;
```
sivanharel, line 161: The line-height property is defined inside the 'LabelTextFont' var and set to 1.5em.
As I see in the spec in our ZeroHeights the line-height should be 24px which is 1.5em.

`LabelTextFont: "{theme: 'Body-M', size: '16px', lineHeight: '1.5em'}";`

Screen Shot 2020-12-27 at 15 53 39

https://zeroheight.com/7sjjzhgo2/p/05a744-05-selections/b/285e96

sivanharel: Why do you need to change it?
If there is a problem let's talk with @michalvardi and consider changing it :)

GuyPerkal: The issue is that having the 24px line-height causes the total height to be 58px, while in zeroheight it's 50px. If that is not a problem feel free to close this.

jonathanadler: From what I can see in zeroheight, it should actually be `56px`.
So maybe instead of changing the lineheight, we need to change the padding?

jonathanadler: image

GuyPerkal:
Alright, changed the padding so that it's now 56px. I was going by this originally - 
Screen Shot 2020-12-27 at 17 40 12

src/components/RadioButton/RadioButton.st.css
sivanharel: So it seems that there is no need to change anything...
There is padding of 16px top an bottom so it's 16+16+ 24(line height) = 56px

GuyPerkal: You would think but it is actually 58px, I'm not sure why. I guess the border is adding to the height for some reason.

sivanharel: I got it.. The border is 1px so yes we could reduce the padding

sivanharel (approved):

PR #562, opened 2020-12-30
feat(wix-ui-tpa): counter-add xSmall size

urig-wix:
We use the <Counter /> component in our new ooi-cart and we need to use the XS size for our mobile layout, which is not yet developed (but is designed: https://zeroheight.com/7sjjzhgo2/p/157dab-04-inputs/b/591891).


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_562/

src/components/Counter/Counter.st.css
```diff
@@ -46,6 +46,11 @@
     border-color: color(value(counterColor));
 }
 
+.root.xsmall {
+    padding: 0 4px;
+    width: 64px;
```
jonathanadler, line 51: I think the `width` should probably also be overridable, as in line 40

src/components/Counter/Counter.st.css
```diff
@@ -46,6 +46,11 @@
     border-color: color(value(counterColor));
 }
 
+.root.xsmall {
```
jonathanadler, line 49: Also, not critical, but the `.root` isn't needed here.
This can be:
```css
.xsmall {
  ...
}
```

src/components/Counter/Counter.tsx
```diff
@@ -59,6 +70,21 @@ export class Counter extends React.Component<CounterProps> {
...
 
+  _getPlus = () => {
+    const { size } = this.props;
+    return size === CounterSize.xSmall ? <PlusS /> : <Plus />;
+  }
+
+  _getMinus = () => {
+    const { size } = this.props;
+    return size === CounterSize.xSmall ? <MinusS /> : <Minus />;
+  }
```
jonathanadler, line 81: Is it necessary to use different assets for the different sizes?
Or maybe we can just change it via css (which is better)?

jonathanadler (commented):
Excellent!

I left some really small notes.

Also, please just add an xSmall example in the `CounterExtendedExample.tsx` file 🙏 

I'll ask the designers to give a review as well

src/components/Counter/Counter.st.css
urig-wix: tried, the following didn't work, wdim?
`:vars {
    /*Defaults*/
    ...
    DefaultWidthXsmall: 64px;
}

:vars {
    ...
    counterWidthXsmall: fallback(value(MainWidth), value(counterWidthXsmall));
}

.xsmall {
    width: value(counterWidthXsmall);
}`

src/components/Counter/Counter.tsx
urig-wix: done

src/components/Counter/Counter.tsx
```diff
@@ -92,7 +102,7 @@ export class Counter extends React.Component<CounterProps> {
           name="increment"
           disabled={disabled || (max && value + step > max)}
         >
-          <Plus />
+          <Plus className={sizeClass} />
```
jonathanadler, line 105: It's better to avoid using the same class for different elements, since it's harder to target them and make changes.

Instead of passing the `sizeClass` to different elements, you should just pass it to the root element and target the other elements with it:
```css
.xSmall .btnIcon { ... }
.xSmall .error { ... }
```
etc.

jonathanadler (commented):
Looks good 👌 

I left one comment regarding the styling.

Also note that the build is failing because of the name change of the PlusS icon (`PlusS.svg` was changed to `plusS.svg`).

After the build will run again, we can see how the component looks in storybook, and I'll pass it for a designer QA

src/components/Counter/docs/CounterExtendedExample.tsx
```diff
@@ -52,6 +52,17 @@ export class CounterExtendedExample extends React.Component<{}, State> {
...
+      <Counter
+        inputAriaLabel={'amount'}
+        incrementAriaLabel={'increment'}
+        decrementAriaLabel={'decrement'}
+        onChange={this.handleChange}
+        value={this.state.value}
+        size={CounterSize.xSmall}
+        {...this.props}
+        className={classes.root}
+      />
```
jonathanadler, line 65: I'm not sure why, but this isn't rendering as an xSmall Counter.
It might be because of the `classes.root` class.
this is how it looks:
image

jonathanadler (commented):
Great job!

Again, left a small comment, regarding the extended example.

Also, notice that the component doesn't work well with the error icon.
It should somehow stretch:
image

src/components/Counter/docs/CounterExtendedExample.tsx
urig-wix: Was definitely caused because of the 'root' class name. was able to solve by adding '!important' to the overrided properties at xSmall. Is that the right thing to do though?

And one more issue - when the counter value is greater than 9 the input number is cut at the edges due to lack of space. The medium sized counter behaves the same (the number is cut when there is not enough space) but for much bigger numbers. What is the right thing to do in such cases ? Is that a questions for the designers?

jonathanadler: So no, actually we shouldn't use `!important` in the component.
I would just pass a different className to the xSmall example.
Cause the problem is with the example, and not the code in this case

jonathanadler (commented):
Looks great 🏆 

I'll ask Johanna to review and hopefully we can merge it soon.

Oh, and just one last thing I forgot to ask you:
Please add the `size` prop to the component's playground.

You can see how it works in the Tags component [here](https://github.com/wix/wix-ui-tpa/blob/7d19f82ac2f8a68f6ef1d89eeb5697741bbe9fed/src/components/Tags/docs/index.story.tsx#L69) and [here](https://github.com/wix/wix-ui-tpa/blob/7d19f82ac2f8a68f6ef1d89eeb5697741bbe9fed/src/components/Tags/docs/index.story.tsx#L75)

Great job!

jonathanadler (commented):
Looks good!
I'll send it to Johanna for review.

However, I'm not sure why, but it seems the width variable isn't affecting the xSmall counter in the settings panel page.
Can you check why this is happening?

urig-wix:
> Looks good!
> I'll send it to Johanna for review.
> 
> However, I'm not sure why, but it seems the width variable isn't affecting the xSmall counter in the settings panel page.
> Can you check why this is happening?

Solved by changing [this](https://github.com/wix/wix-ui-tpa/commit/0b0c8c77dfb789f2a44ab03620d4752458d01e63#diff-def64fe4eb3858c340f76676fd1c2a7fa6093aa2d80e3134b0ed6256ed38b1f1R54).

In addition, from a reason I don't quite understand, the counters at the settings panel [here](http://localhost:6006/?activeTab=Usage&path=%2Fstory%2Fcomponents-api-components--counter) don't have width initialised with the default value unless they have max & min props (see the first two counters verse the third). Not sure why they behave like that but I've added max min (with the values 999/-999 respectively) to the xSmall one as well. LMK if it does not make sense

jonathanadler (approved):
👑

PR #566, opened 2020-12-30
Settings panel improvements

jonathanadler:
# ✨ Pull Request

### 💁 Description
* Fixed default colors issue (wrong default colors)
* Added font size slider
* Some UI touch-up - margins and such

...

### 💭 Motivation
To have a better experience, especially for designers' QA

...

### 👀  PR is ready for review 

- [x] Ready for review


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_566/

stories/helperComponents/MockSettings.tsx
```diff
@@ -151,7 +145,7 @@ export class MockSettings extends React.PureComponent<
       const tpaIdx = parseInt(item.defaultColor.split('-')[1], 10) - 1;
       const { row, col } = this.getPaletteIndicesFromTPAIdx(tpaIdx);
       obj[item.wixParam] = {
-        value: COLOR_PALETTE[col][row],
+        value: palettes[0][col][row],
```
jonathanadler, line 148: This part was just wrong.
It caused a bug where the initial colors were wrong, because the `COLOR_PALETTE` was the old paletee, before our refactor.

nitayneeman (approved):
Nice!

sivanharel (commented):
Very nice!

Something that we should pay attention to:
There are some components that the font (font-family) could be changed but the font size is fixed. 
After this fix, it adds this font size slider automatically for all the components that have font settings. 
We need the ability to add it only when it is needed or the ability to remove it in cases that the font size is fixed.

jonathanadler:
> Very nice!
> 
> Something that we should pay attention to:
> There are some components that the font (font-family) could be changed but the font size is fixed.
> After this fix, it adds this font size slider automatically for all the components that have font settings.
> We need the ability to add it only when it is needed or the ability to remove it in cases that the font size is fixed.

That's a good point.
I thought about this, and I haven't reached a conclusion yet.

On one hand, we can show the font-size slider only when the `size` prop is passed in `index.story.tsx` file, as part of the settingsPanel props.
That way it will appear only for components which we decide to open it for.

On the other hand:
1. It might be a bit of a hassle adding it for each component (but also maybe it won't)
2. Having it appear for all components will give us a chance to check that the font-size stays fix when it needs to be. So it will allow us to give it some QA

Again, I'm not really sure what's best.

We can of course start with one approach, and then try the second if it seems to be better.

wdyt?

sivanharel:
> > Very nice!
> > Something that we should pay attention to:
> > There are some components that the font (font-family) could be changed but the font size is fixed.
> > After this fix, it adds this font size slider automatically for all the components that have font settings.
> > We need the ability to add it only when it is needed or the ability to remove it in cases that the font size is fixed.
> 
> That's a good point.
> I thought about this, and I haven't reached a conclusion yet.
> 
> On one hand, we can show the font-size slider only when the `size` prop is passed in `index.story.tsx` file, as part of the settingsPanel props.
> That way it will appear only for components which we decide to open it for.
> 
> On the other hand:
> 
> 1. It might be a bit of a hassle adding it for each component (but also maybe it won't)
> 2. Having it appear for all components will give us a chance to check that the font-size stays fix when it needs to be. So it will allow us to give it some QA
> 
> Again, I'm not really sure what's best.
> 
> We can of course start with one approach, and then try the second if it seems to be better.
> 
> wdyt?

We could also add a prop something like "hideSizePannel", but maybe your first option is the best solution.
I really think we should hide it from the irrelevant places in the storybook because it would confused peoples and we will probably get questions on the channel.

jonathanadler:
> > > Very nice!
> > > Something that we should pay attention to:
> > > There are some components that the font (font-family) could be changed but the font size is fixed.
> > > After this fix, it adds this font size slider automatically for all the components that have font settings.
> > > We need the ability to add it only when it is needed or the ability to remove it in cases that the font size is fixed.
> > 
> > 
> > That's a good point.
> > I thought about this, and I haven't reached a conclusion yet.
> > On one hand, we can show the font-size slider only when the `size` prop is passed in `index.story.tsx` file, as part of the settingsPanel props.
> > That way it will appear only for components which we decide to open it for.
> > On the other hand:
> > 
> > 1. It might be a bit of a hassle adding it for each component (but also maybe it won't)
> > 2. Having it appear for all components will give us a chance to check that the font-size stays fix when it needs to be. So it will allow us to give it some QA
> > 
> > Again, I'm not really sure what's best.
> > We can of course start with one approach, and then try the second if it seems to be better.
> > wdyt?
> 
> We could also add a prop something like "hideSizePannel", but maybe your first option is the best solution.
> I really think we should hide it from the irrelevant places in the storybook because it would confused peoples and we will probably get questions on the channel.

OK, you convinced me.

So i've added a `fixedSize` prop, so that if it isn't needed, we can just use this flag.
In order for it to be discoverable, I've added it to the generator.

PR #567, opened 2020-12-31
Image - support blurry loading

nitayneeman:
This PR adds a blurry progressive loading for `<Image/>` inspired by [Restaurants](https://github.com/wix-private/restaurants/blob/master/restaurants-olo-client/src/components/MainPage/components/PreviewedImage/PreviewedImage.tsx#L100).

Notice it's not the exact implementation, but the concept is pretty similar.


src/components/Image/Image.tsx
```diff
@@ -17,37 +17,106 @@ export interface ImageProps extends TPAComponentProps {
...
   /** A callback to be called if error occurs while loading */
   onError?: React.EventHandler<React.SyntheticEvent>;
+  /** An experience to set while the image is fetched and loaded  */
+  loadingBehavior?: 'none' | 'blur';
 }
 
+interface Dimensions {
+  width: ImageProps['width'];
+  height: ImageProps['height'];
+}
```
nitayneeman, line 27: I tried to use `Pick` but the problem is that `width` & `height` are optional at the moment in `ImageProps`. Either way, the dimensions logic is probably going to change...

src/components/Image/Image.visual.tsx
```diff
@@ -48,6 +48,15 @@ visualize('Image', () => {
       snap('with onError', (done) => (
         <ImageWrapper src={invalidSrc} onDone={done} />
       ));
+      snap('with blurry loading', (done) => (
+        <Image
+          src={src}
+          width={200}
+          height={200}
```
nitayneeman, line 55: Currently we have to pass the dimensions to render a media image, so we cannot take the screenshot without them (even though they are not related to the test)

nitayneeman:
I'm reviewing the animation with @michalvardi

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_567/

src/components/Image/Image.st.css
```diff
@@ -1,2 +1,32 @@
...
+  }
+}
+
 .root {
+  -st-states: preload, loaded;
+
+  display: grid;
+}
+
+.root img {
```
shlomitc, line 20: why don't you use the `MediaImage` style API?

src/components/Image/docs/examples.ts
```diff
@@ -17,3 +17,13 @@ export const relativeUriExample = `
   alt="Garfield smiles and puts his hand over chest"
 />
 `;
+
+export const blurLoadingExample = `
```
shlomitc, line 21: consider adding a button to retrigger the rendering since it's very quick.

src/components/Image/Image.st.css
nitayneeman: I'm not sure there is a style API...

These are the props:
https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/media-image/media-image.tsx#L96-L102

I thought about exposing `className` prop - WDYT?

src/components/Image/docs/examples.ts
nitayneeman: Basically you can just change any value in the props to see it again, but I'll add a button to make it easier 👌

src/components/Image/Image.st.css
jonathanadler: Yeah I think a `className` prop is definitely needed no matter the approach.
That way you won't need to target any internal elements.

I would even add it to wix-ui-core before merging this PR.

src/components/Image/Image.st.css
```diff
@@ -1,2 +1,32 @@
...
+
+.root:preload img:last-child {
+  filter: blur(value(BlurRadius));
+}
+
+.root:loaded img:last-child {
+  animation-name: blurry-loading;
+  animation-duration: 0.3s;
+  animation-timing-function: easeout;
 }
```
jonathanadler, line 32: After adding the `className` prop to the MediaImage component, I would give different class names to the two images, just so not to rely on the dom ordering, and for better readability.

src/components/Image/Image.tsx
```diff
@@ -17,37 +17,106 @@ export interface ImageProps extends TPAComponentProps {
...
+  />
+);
+
 /** Image is a component to literally display an image - whether an absolute with full URL or a media platform item with relative URI */
 export class Image extends React.Component<ImageProps> {
   static displayName = 'Image';
 
+  state = { isLoaded: false };
+
+  _onLoad(event) {
```
jonathanadler, line 58:
```suggestion
  _onLoad = (event) => {
```

src/components/Image/Image.tsx
```diff
@@ -17,37 +17,106 @@ export interface ImageProps extends TPAComponentProps {
...
+          className,
+        )}
         data-hook={this.props['data-hook']}
       >
         {isAbsoluteUrl ? (
           <CoreImage
             {...imageProps}
             nativeProps={{ ...dimensions }}
             src={src}
+            onLoad={(event) => this._onLoad(event)}
```
jonathanadler, line 100:
```suggestion
            onLoad={this._onLoad}
```

jonathanadler (commented):
Nice!

Left some small comments.

src/components/Image/Image.st.css
nitayneeman: I'll add a `className` for the wrapper div in `Image` & `MediaImage`(since I use both, depends on `src`), but I'll still need to target the native `img` inside.

I can do that in two ways:
1. Expose a prop like `nativeElementClass` on the native `img`
2. Make the native `img` extensible using Stylable

Any preference?

nitayneeman:
@jonathanadler I added `className` props in ui-core:
https://github.com/wix/wix-ui/pull/2000

And also changed the component implementation to use them, I agree it's much readable than before.

jonathanadler (approved):
The only thing I'm not sure about, is how to reflect the fact that using the blur load with an absolute src is merely visual, and doesn't have an affect on performance.

nitayneeman:
> The only thing I'm not sure about, is how to reflect the fact that using the blur load with an absolute src is merely visual, and doesn't have an affect on performance.

I'll mention it on the story page.

PR #578, opened 2021-01-04
build: configure Prettier

nitayneeman:
This PR adds a basic Prettier configuration to avoid conflicts after committing - inspired by [WSR solution](https://github.com/wix/wix-style-react/blob/master/.prettierrc.js#L1-L5).

I didn't add `arrowParens` since I don't see the source code is formatted like that in wix-ui-tpa after committing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -173,5 +173,9 @@
   "yoshi": {
     "exports": "wix-ui-tpa",
     "tpaStyle": true
+  },
+  "prettier": {
+    "singleQuote": true,
+    "trailingComma": "all"
   }
 }


package.json
```diff
@@ -173,5 +173,9 @@
   "yoshi": {
     "exports": "wix-ui-tpa",
     "tpaStyle": true
+  },
+  "prettier": {
+    "singleQuote": true,
+    "trailingComma": "all"
```
jonathanadler, line 179: I think we already get these from yoshi, no?

nitayneeman: Perhaps, but it seems that Husky runs with different defaults (or doesn't take from Yoshi) - which probably leads to the conflicts...

For VSCode I can say this fixes the problem, but for WebStrom it might be needed for each contributor to configure manually (not sure if Prettier plugin is already supported well in WebStorm?).

We also can take the complete file instead:
https://github.com/wix/wix-style-react/blob/master/.prettierrc.js

And then WebStorm might find the file (not sure if it detects the config when its inside the package.json)

jonathanadler:
@nitayneeman Can I close this for now?
Have you checked with Yoshi maybe about this?

nitayneeman:
> @nitayneeman Can I close this for now?
> Have you checked with Yoshi maybe about this?

I haven't checked yet but let's close it for now.

PR #584, opened 2021-01-05
use one version back of chrome in eyes-storybook

jonathanadler:
# ✨ Pull Request

### 💁 Description
Trying to control browser version in applitools
...

### 💭 Motivation
https://wixeyes.applitools.com/app/test-results/00000251792465730209/?accountId=mE99eP8CUn9UTSXSv98L3OSaJ0NZmAcY6Wndx6WSHODdY110

...

### 👀  PR is ready for review 

- [x] Ready for review

### DIFF
--- a/applitools.config.js
+++ b/applitools.config.js
@@ -9,6 +9,6 @@ try {
 module.exports = applitoolsConfig({
   config: merge(privateConfig, {
     concurrency: 50,
-    browser: { name: 'chrome-one-version-back' }
+    browser: { name: 'chrome-one-version-back', width: 800, height: 600 }
   }),
 });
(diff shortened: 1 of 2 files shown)


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_584/

nitayneeman (commented):
Interesting.
Is it going to always use the older chrome version or only when having failures?
And it solves the problems because the problems happen due to Chrome version changes?

jonathanadler:
> Interesting.
> Is it going to always use the older chrome version or only when having failures?
> And it solves the problems because the problems happen due to Chrome version changes?

Until we bring it back, it will always use chrome two version back.

For now, this will allow us to apply all the differences in a PR and not in master

PR #586, opened 2021-01-05
Popover component

mbutsykin:
# ✨ Pull Request

### 💁 Description
[Popover component](https://docs.google.com/document/d/1SUCddmcECr3n2hoTAryQa5fd3Am7Tp_EOaRgV4Nff4I/edit#) implementation

### 👀  PR is ready for review 

- [X] Ready for review

### DIFF
--- a/src/components/Popover/Popover.spec.tsx
+++ b/src/components/Popover/Popover.spec.tsx
@@ -8,7 +8,8 @@ import { popoverTestkitFactory } from '../../testkit';
 import { popoverTestkitFactory as enzymePopoverTestkitFactory } from '../../testkit/enzyme';
 
 import { popoverDriverFactory } from './Popover.driver';
-import { Popover, PopoverProps } from './';
+import { Popover, PopoverProps, TriggerAction } from './';
+import { Simulate } from 'react-dom/test-utils';
 
 describe('Popover', () => {
   const createDriver = createUniDriverFactory(popoverDriverFactory);
@@ -34,12 +35,29 @@ describe('Popover', () => {
     expect(await driver.exists()).toBe(true);
   });
 
-  describe('behaviour', () => {
-    it('should open on element click', async () => {
+  describe('uncontrolled', () => {
+    it("shouldn't open on click", async () => {
       const driver = bootstrap();
-
-      expect(await driver.isContentElementExists()).toBeFalsy();
       await driver.click();
+      expect(await driver.isContentElementExists()).toBeFalsy();
+    });
+
+    it("shouldn't open on click", async () => {
+      const driver = bootstrap();
+      expect(await driver.isContentElementExists()).toBeFalsy();
+      Simulate.mouseEnter(await driver.element());
+      expect(await driver.isContentElementExists()).toBeTruthy();
+    });
+  });
+
+  describe('controlled', () => {
+    it('should be open when shown is false', async () => {
+      const driver = bootstrap({ shown: false });
+      expect(await driver.isContentElementExists()).toBeFalsy();
+    });
+
+    it('should be open when shown is true', async () => {
+      const driver = bootstrap({ shown: true });
       expect(await driver.isContentElementExists()).toBeTruthy();
     });
   });
--- a/src/components/Popover/Popover.spec.tsx
+++ b/src/components/Popover/Popover.spec.tsx
@@ -46,8 +46,8 @@ describe('Popover', () => {
 
   describe('testkit', () => {
     it('should exist', async () => {
-      await expect(
-        isUniTestkitExists(
+      expect(
+        await isUniTestkitExists(
           <Popover placement="top">
             <Popover.Element>
               <span>element</span>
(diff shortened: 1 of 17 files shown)


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_586/

src/components/Popover/Popover.st.css
```diff
@@ -0,0 +1,87 @@
...
+
+.root:wired .popover[data-placement*="bottom"] .arrow,
+.root.overrideStyleParams .popover[data-placement*="bottom"] .arrow {
+  border-bottom-color: value(backgroundColor);
+}
+
+.root:wired .popover[data-placement*="top"] .arrow,
+.root.overrideStyleParams .popover[data-placement*="top"] .arrow {
+  border-top-color: value(backgroundColor);
+}
```
jonathanadler, line 87: I suggest you take a look at how we implemented `wiredToSiteColors` in the Dialog component.
https://github.com/wix/wix-ui-tpa/blob/master/src/components/Dialog/Dialog.st.css

src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,117 @@
...
+      {React.cloneElement(
+        Element,
+        {},
+        React.cloneElement(Element.props.children, getTriggerProps()),
+      )}
+      {Content}
+    </CorePopover>
+  );
+
+  function getTriggerProps() {
```
jonathanadler, line 57: We prefer to use the `_` prefix for our internal methods:
```suggestion
  function _getTriggerProps() {
```
Please adjust the rest as well 🙏

src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,117 @@
...
+} = props => {
+  const {
+    triggerAction,
+    wiredToSiteColors,
+    children,
+    className,
+    ...rest
+  } = props;
+  const [isShown, setIsShown] = React.useState<boolean>(false);
+  const [id] = React.useState(generateKey('popover'));
```
jonathanadler, line 33: In this case it's better to use `useRef`
```suggestion
  const id = React.useRef(generateKey('popover'));
```
But are hooks allowed in the platform?

src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  const [Element, Content] = React.Children.toArray(
+    children,
+  ) as React.ReactElement[];
+
+  return (
+    <CorePopover
+      id={id}
```
jonathanadler, line 41: So the `id` should actually be added to the tooltip's content container, in order for it to really describe the element

jonathanadler: Meaning this probably should move to the Popover.Content part

src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,117 @@
...
+    children,
+    className,
+    ...rest
+  } = props;
+  const [isShown, setIsShown] = React.useState<boolean>(false);
+  const [id] = React.useState(generateKey('popover'));
+
+  const [Element, Content] = React.Children.toArray(
+    children,
+  ) as React.ReactElement[];
```
jonathanadler, line 37: This means the `Element` child **has** to be written above the `Content` child, no?
Maybe it would be better to iterate over the children and check their type.
You can see how it was done in the AvatarGroup component:
https://github.com/wix/wix-ui-tpa/blob/3a76e4002d818cd59d0a00ba2589b900bd4b6c24/src/components/AvatarGroup/AvatarGroup.tsx#L47

src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,117 @@
...
+    setIsShown(false);
+  }
+
+  function handleKeyDown(e: React.KeyboardEvent<HTMLElement>) {
+    if (e.key === KEYS.Spacebar || e.key === KEYS.Enter) {
+      e.preventDefault();
+      handleOpen();
+    }
+  }
+};
```
jonathanadler, line 106: You can use our util function from here:
https://github.com/wix/wix-ui-tpa/blob/master/src/common/keyCodes.ts#L48

src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,117 @@
...
+          onKeyDown: handleKeyDown,
+          role: 'dialog',
+          appendTo: 'parent',
+        };
+      case TriggerAction.hover:
+      default:
+        return {
+          onMouseEnter: handleOpen,
+          onMouseLeave: handleClose,
+          role: 'tooltip',
```
jonathanadler, line 87: So the id should be passed here

jonathanadler (commented):
Nice!

I left some comments but overall it looks great.
I'll also send this to @michalvardi for design review.

Great work as always 💪

src/components/Popover/Popover.spec.tsx
```diff
@@ -0,0 +1,90 @@
...
+  });
+
+  describe('behaviour', () => {
+    it('should open on element click', async () => {
+      const driver = bootstrap();
+
+      expect(driver.isContentElementExists()).toBeFalsy();
+      await driver.click();
+      expect(driver.isContentElementExists()).toBeTruthy();
+    });
```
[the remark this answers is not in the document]
mbutsykin: after like an hour of trying - i wasnt able to successfully trigger mouseover event =/

PR #588, opened 2021-01-06
Image - support resizing

nitayneeman:
# ✨ Pull Request

### 💁 Description

This PR implements resizing options for `Image`.

### ☝️ TODOs 

Depends on https://github.com/wix/wix-ui/pull/2004, so we need to merge that one first.

### 👀  PR is ready for review 

- [X] Ready for review

### DIFF
--- a/src/components/Image/Image.st.css
+++ b/src/components/Image/Image.st.css
@@ -12,7 +12,7 @@
 }
 
 .root {
-  -st-states: resize(enum(cover, contain)), preload, loaded;
+  -st-states: resize(enum(contain, cover)), preload, loaded;
 
   display: grid;
 }
--- a/src/components/Image/Image.st.css
+++ b/src/components/Image/Image.st.css
@@ -12,13 +12,17 @@
 }
 
 .root {
-  -st-states: contain, preload, loaded;
+  -st-states: resize(enum(cover, contain)), preload, loaded;
 
   display: grid;
 }
 
-.root:contain .absoluteImage, .root:contain .relativeImage, .root:contain .placeholder {
+.root:resize(contain) .absoluteImage, .root:resize(contain) .relativeImage, .root:resize(contain) .placeholder {
   object-fit: contain;
+}
+
+.root:resize(cover) .absoluteImage, .root:resize(cover) .relativeImage, .root:resize(cover) .placeholder {
+  object-fit: cover;
 }
 
 .placeholder, .relativeImage {
--- a/src/components/Image/Image.st.css
+++ b/src/components/Image/Image.st.css
@@ -12,16 +12,16 @@
 }
 
 .root {
-  -st-states: resize(enum(contain, cover)), preload, loaded;
+  -st-states: preload, loaded;
 
   display: grid;
 }
 
-.root:resize(contain) .absoluteImage, .root:resize(contain) .relativeImage, .root:resize(contain) .placeholder {
+.root.contain .absoluteImage, .root.contain .relativeImage, .root.contain .placeholder {
   object-fit: contain;
 }
 
-.root:resize(cover) .absoluteImage, .root:resize(cover) .relativeImage, .root:resize(cover) .placeholder {
+.root.cover .absoluteImage, .root.cover .relativeImage, .root.cover .placeholder {
   object-fit: cover;
 }
 
--- a/src/components/Image/Image.st.css
+++ b/src/components/Image/Image.st.css
@@ -12,9 +12,13 @@
 }
 
 .root {
-  -st-states: preload, loaded;
+  -st-states: contain, preload, loaded;
 
   display: grid;
+}
+
+.root:contain .absoluteImage, .root:contain .relativeImage, .root:contain .placeholder {
+  object-fit: contain;
 }
 
 .placeholder, .relativeImage {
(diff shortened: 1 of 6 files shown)


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_588/

src/components/Image/Image.tsx
```diff
@@ -112,6 +121,7 @@ export class Image extends React.Component<ImageProps> {
             )}
             <MediaImage
               {...imageProps}
+              {...dimensions}
```
nitayneeman, line 124: Notice that at the moment we support only 1:1 aspect ratio, which means we actually cannot reflect the `contain` option for a relative media item. In the next task we'll support different aspect ratios and then `contain` will be also reflected for media items. 

However, for pre-made (absolute) URLs with specified dimensions that are different from the provided `width` & `height` it will work starting this PR.

src/components/Image/Image.tsx
```diff
@@ -73,20 +79,23 @@ export class Image extends React.Component<ImageProps> {
...
     const isAbsoluteUrl = src && src.match('^https?://');
+    const isContained = resize === 'contain';
     const hasLoadingBehavior = loadingBehavior === 'blur';
 
     return (
       <div
         className={st(
           classes.root,
           {
+            contain: isContained,
```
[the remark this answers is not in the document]
nitayneeman: Okay, my bad, I thought for some reason the cover is the default value of `object-fit` - and that's not true. So basically we'll need to style it as well so `resize` probably should be passed as state.

src/components/Image/docs/examples.ts
```diff
@@ -3,21 +3,40 @@ export const importExample = `import { Image } from 'wix-ui-tpa/Image';`;
...
 />
 `;
 
 export const relativeUriExample = `
 <Image
   src="c5f754_dd75514d14fa4057b4f4a6cc8ce7add3~mv2.jpg"
-  width="300"
-  height="250"
+  width="480"
+  height="360"
```
nitayneeman, line 16: Just alining the image to original dimensions to emphasize the diff in the resizing example

nitayneeman:
After talking to @michalvardi, the default will be `contain` because:

- `cover` & `none` might crop the image, and we want to full image by default
- `fill` stretches/shrinks the image, which doesn't look good

@jonathanadler @sivanharel  Now it's ready for review

src/components/Image/Image.tsx
```diff
@@ -17,10 +17,14 @@ export interface ImageProps extends TPAComponentProps {
   onLoad?: React.EventHandler<React.SyntheticEvent>;
   /** A callback to be called if error occurs while loading */
   onError?: React.EventHandler<React.SyntheticEvent>;
+  /** Specifies how the image is resized to fit its container */
+  resize?: 'contain' | 'cover';
```
nitayneeman, line 21: Actually we can rename this prop merely `cover`, since it's only two values and `contain` is the default anyway, but I think it's more clear like that and also there still has a potential we'll add more resize options in the future so I wouldn't do that to avoid breaking the API...

src/components/Image/Image.st.css
```diff
@@ -12,11 +12,19 @@
...
 
 .root {
-  -st-states: preload, loaded;
+  -st-states: resize(enum(contain, cover)), preload, loaded;
 
   display: grid;
 }
 
+.root:resize(contain) .absoluteImage, .root:resize(contain) .relativeImage, .root:resize(contain) .placeholder {
+  object-fit: contain;
```
jonathanadler, line 21: This isn't supported by older browsers.
I guess it's fine, as long as it doesn't break in older browsers

src/components/Image/Image.st.css
```diff
@@ -12,11 +12,19 @@
 }
 
 .root {
-  -st-states: preload, loaded;
+  -st-states: resize(enum(contain, cover)), preload, loaded;
```
jonathanadler, line 15: I think it might be better to use classes instead of stylable states.
In general, in our library.

jonathanadler (approved):

src/components/Image/Image.st.css
nitayneeman: Yeah, IE11 isn't supported, but it's pretty much the case with CSS variables support as well...
AFAIR the viewer will never support IE11, so I think it's okay.

src/components/Image/Image.st.css
nitayneeman: I saw a lot of usages (about 30+- components) in our library, maybe worth mentioning it in a guideline doc?

src/components/Image/Image.st.css
jonathanadler: Not exactly.
Currently TPA's also run on bolt, especially in the editor where TB isn't supported yet.
That's why we don't use css variables yet as well.

In general we should support ie11, but it's also ok if ie11 will have an inferior experience, without breakage though.

src/components/Image/Image.st.css
jonathanadler: yeah, we used states up until now.
It's something that we're just starting to realize and understand

src/components/Image/Image.st.css
nitayneeman: I tested using Browserling on IE9 - it seems the browser just displays it as a regular image and ignores the property:
image

In a normal browser, it should look like this:
image

PR #591, opened 2021-01-10
Image - support aspectRatio prop

nitayneeman:
# ✨ Pull Request

### 💁 Description

This PR implements the aspectRatio prop for `Image`.

We should decide how to manage and infer the dimensions (it's not in the scope of this task but it might change the implementation of `aspectRatio` later).

### 👀  PR is ready for review 

- [X] Ready for review

### DIFF
--- a/src/components/Image/Image.tsx
+++ b/src/components/Image/Image.tsx
@@ -4,16 +4,16 @@ import { MediaImage } from 'wix-ui-core/media-image';
 import { TPAComponentProps } from '../../types';
 import { classes, st } from './Image.st.css';
 
-const enum AspectRatioOptions {
-  square = 1,
-  cinema = 16 / 9,
-  landscape = 4 / 3,
-}
-
 const enum ResizeOptions {
   contain = 'contain',
   cover = 'cover',
 }
+
+const AspectRatioOptions = {
+  square: 1,
+  cinema: 16 / 9,
+  landscape: 4 / 3,
+};
 
 export interface ImageProps extends TPAComponentProps {
   /** The source could be any absolute full URL or a relative URI of a media platform item */
@@ -31,7 +31,7 @@ export interface ImageProps extends TPAComponentProps {
   /** Specifies how the image is resized to fit its container */
   resize?: 'contain' | 'cover';
   // Specifies the proportional relationship between width and height
-  aspectRatio?: '1:1' | '16:9' | '4:3' | number;
+  aspectRatio?: 'square' | 'cinema' | 'landscape' | number;
   /** An experience to set while the image is fetched and loaded  */
   loadingBehavior?: 'none' | 'blur';
 }
@@ -107,7 +107,9 @@ export class Image extends React.Component<ImageProps> {
     const isAbsoluteUrl = src && src.match('^https?://');
 
     const aspectRatioAsNumber =
-      typeof aspectRatio === 'number' ? aspectRatio : 1;
+      typeof aspectRatio === 'number'
+        ? aspectRatio
+        : AspectRatioOptions[aspectRatio];
     const sourceDimensions = { width, height };
     const containerDimensions = {
       width:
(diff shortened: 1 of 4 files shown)


src/components/Image/Image.tsx
```diff
@@ -84,14 +97,27 @@ export class Image extends React.Component<ImageProps> {
...
+      typeof aspectRatio === 'number'
+        ? aspectRatio
+        : AspectRatioOptions[aspectRatio];
+    const sourceDimensions = { width, height };
+    const containerDimensions = {
+      width:
+        !width && aspectRatioAsNumber ? height * aspectRatioAsNumber : width,
+      height:
+        !height && aspectRatioAsNumber ? width / aspectRatioAsNumber : height,
+    };
```
nitayneeman, line 119: We might improve this logic a little later

src/components/Image/Image.visual.tsx
```diff
@@ -60,6 +60,21 @@ visualize('Image', () => {
         />
       ));
 
+      story('with aspectRatio', () => {
```
nitayneeman, line 63: After we decide on the valid variations I'll add additional tests representing more combinations (aspectRatio with/without dimensions)

src/components/Image/Image.tsx
```diff
@@ -84,14 +97,27 @@ export class Image extends React.Component<ImageProps> {
...
       resize,
       loadingBehavior,
       ...imageProps
     } = this.props;
     const { isLoaded } = this.state;
-    const dimensions = { width, height };
 
     const isAbsoluteUrl = src && src.match('^https?://');
+
+    const aspectRatioAsNumber =
```
nitayneeman, line 109: It's a bit odd to name this variable like that since aspectRatio is typically a number. But in our case it also supports the predefined values 😄

On the other hand, not sure I would rename the prop...

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_591/

src/components/Image/Image.tsx
sivanharel: I think that it is good to keep the prop name as is even it could also be a string :)

src/components/Image/Image.tsx
sivanharel: Is the case that both width and height are not defined is possible?

sivanharel (approved):
Awesome!!
I just have 2 questions :)

src/components/Image/Image.tsx
nitayneeman: Currently we don't support it since MediaImage of ui-core must get width & height... But we're planning to support this use case by inferring the values from the parent in DOM

src/components/Image/Image.tsx
```diff
@@ -24,11 +30,13 @@ export interface ImageProps extends TPAComponentProps {
...
   /** Specifies how the image is resized to fit its container */
   resize?: 'contain' | 'cover';
+  // Specifies the proportional relationship between width and height
+  aspectRatio?: 'square' | 'cinema' | 'landscape' | number;
   /** An experience to set while the image is fetched and loaded  */
   loadingBehavior?: 'none' | 'blur';
 }
 
-type DefaultProps = Pick<ImageProps, 'resize'>;
+type DefaultProps = Pick<ImageProps, 'resize' | 'aspectRatio'>;
```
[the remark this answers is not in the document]
nitayneeman: I meant `Pick`, which is util of TypeScript to pick only specific fields from an interface (in this case - ImageProps)  :)

Otherwise I would have to duplicate the code and declared them again (not a big deal BTW, but the current one is just cleaner)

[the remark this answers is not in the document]
sivanharel: NIce, good to know:)

PR #599, opened 2021-01-12
add extended examples to perfer check

jonathanadler:


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_599/

jonathanadler:
@deanshub 
I've added the "extended" components test, I even didn't need to create an additional build process.

However, we need to notice that the "extended" components I'm testing in this case, are our storybook examples, so sometimes it's more than a single component.
This is the current solution I created now.

We might need to consider replacing what I did in this PR, to two new components instead:
1. A component that uses our component without overriding the styles
2. A component that uses our component with overriding the styles
And then compare the two.
This is in addition to just testing our component's bundle (but maybe this is redundant?)

@nitayneeman @sivanharel @shlomitc will be happy to hear what you think as well.

nitayneeman:
> However, we need to notice that the "extended" components I'm testing in this case, are our storybook examples, so sometimes it's more than a single component.

I think we'd like to generate them regardless of the storybook examples because of this 👆
And slim as much as possible, but that's a pretty good start.

Maybe we can be content with only testing the two cases you mentioned, since degradations would be reflected in them anyway? Basically, it's for us to understand the impact of overriding the styles, right? So as long as we compare them we theoretically can simply manage only these two cases, but on the other hand, once we decide probably we'd like to bring back the regular bundle testing... 

So I don't know if to remove them at first, depends on the build time

perfer.config.js
```diff
@@ -1,11 +1,18 @@
...
   ['Autocomplete.bundle.min.js', 37],
   ['Avatar.bundle.min.js', 16],
   ['AvatarGroup.bundle.min.js', 19],
   ['Badge.bundle.min.js', 5],
+  ['BadgePerfBasic.bundle.min.js', 5],
+  ['BadgePerfExtendedBad.bundle.min.js', 5],
+  ['BadgePerfExtendedGood.bundle.min.js', 5],
   ['Button.bundle.min.js', 9],
+  ['ButtonPerfBasic.bundle.min.js', 9],
+  ['ButtonPerfExtended.bundle.min.js', 10],
```
jonathanadler, line 15: @deanshub 
You can see for instance that I've added two bundles:
1. is a basic component using our component
2. the same basic component that in addition overrides the default stylings (using stylable)

This will give us a relatively accurate indication on how overriding styles affects the bundle.
I'm not sure about the naming, I wanted it to be something like: `Button.perf.basic.bundle.min.js` and `Button.perf.extended.bundle.min.js`, but this creates a name collision for stylable's compiler.

nitayneeman (approved):
Looks good to me.

Personally I like the current naming - `Button.perf.extended.bundle.min.js` and the other seem  threatening 😯

perfer.config.js
deanshub: looks good from my standpoint as is

PR #609, opened 2021-01-17
feat(DatePicker): Adaptive to font change

sivanharel:
# ✨ Pull Request

### 💁 Description

...

### 💭 Motivation

...

### ☝️ TODOs 

...

### 👀  PR is ready for review 

- [ ] Ready for review

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "lodash": "^4.17.11",
     "normalize-scroll-left": "^0.2.1",
     "react-resize-detector": "^4.1.3",
-    "wix-style-react": "^9.47.0",
+    "wix-style-react": "http://cdn.dev.wixpress.com/wix-style-react/0.0.0-d17a7488344806f2a9ebd8458bb03873",
     "wix-ui-core": "^3.0.150",
     "wix-ui-icons-common": "^2.0.0",
     "yoshi-stylable-dependencies": "^4.86.0",
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "lodash": "^4.17.11",
     "normalize-scroll-left": "^0.2.1",
     "react-resize-detector": "^4.1.3",
-    "wix-style-react": "http://cdn.dev.wixpress.com/wix-style-react/0.0.0-d17a7488344806f2a9ebd8458bb03873",
+    "wix-style-react": "^9.69.0",
     "wix-ui-core": "^3.0.150",
     "wix-ui-icons-common": "^2.0.0",
     "yoshi-stylable-dependencies": "^4.86.0",
--- a/src/components/DatePicker/DatePicker.st.css
+++ b/src/components/DatePicker/DatePicker.st.css
@@ -78,7 +78,8 @@
     WeekdayTextColor: --overridable;
     /*
         The font of a specific date.
-        @default {theme: 'Body-M', size: '14px', lineHeight: '20px'}
+        @default {theme: 'Body-M', size: '14px', lineHeight: 1 }
+        @fixed The line height is 1 and could not be overridden.
     */
     DayFont: --overridable;
     /*
--- a/src/components/DatePicker/DatePicker.st.css
+++ b/src/components/DatePicker/DatePicker.st.css
@@ -22,7 +22,7 @@
 /* Defaults - default params */
 :vars {
     calendarWidth: 100%;
-    calendarHeight: 332px;
+    calendarHeight: 100%;
     sidePadding: 20px;
 
     DefaultBackgroundColor: color-1;
@@ -34,7 +34,7 @@
     DefaultWeekdayFont: "{theme: 'Body-M', lineHeight: '20px'}";
     DefaultWeekdayTextColor: color-8;
 
-    DefaultDayFont: "{theme: 'Body-M', lineHeight: '20px'}";
+    DefaultDayFont: "{theme: 'Body-M', size: '14px', lineHeight: '20px'}";
     DeafultDayColor: color-5;
     DefaultSelectedDayColor: color-1;
     DefaultSelectedDayBGColor: color-8;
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -70,7 +70,7 @@
     "lodash": "^4.17.11",
     "normalize-scroll-left": "^0.2.1",
     "react-resize-detector": "^4.1.3",
-    "wix-style-react": "^9.47.0",
+    "wix-style-react": "http://cdn.dev.wixpress.com/wix-style-react/0.0.0-d17a7488344806f2a9ebd8458bb03873",
```
sivanharel, line 73: Just for testing

src/components/DatePicker/DatePicker.st.css
```diff
@@ -203,10 +204,18 @@
...
-    font-size: 14px;
     color: value(CurrentDayColor);
-    width: 34px;
-    height: 34px;
+    width: calc(34em/14);
+    height: calc(34em/14);
+}
+
+.calendar::dayText {
+    line-height: 1;
```
nitayneeman, line 213: Maybe this should also be relative to `em` like the others?

src/components/DatePicker/docs/DatePickerWiringExample.tsx
```diff
@@ -11,11 +11,45 @@ export const DatePickerWiringExample = class DatePickerExample extends React.Com
...
+      <>
+        <DatePicker
+          className={classes.root}
+          value={value}
+          onChange={this._onChange}
+        />
+        <br />
+        <br />
+        <h3>Calendar With Indications:</h3>
+        <br />
```
nitayneeman, line 23: I would try to avoid these `<br />` 😛

A wrapper with padding and margin-bottom might do the work

nitayneeman (approved):
Nice solution!

src/components/DatePicker/docs/DatePickerWiringExample.tsx
sivanharel: You are right.. I will change it :)

nitayneeman:
Also consider adding a visual test (or maybe already exists?)

src/components/DatePicker/DatePicker.st.css
sivanharel: Actually, I also added this `line-height: 1;` for the WSR component. I added it here as well in order to verify that no one is overriding it.
It is required in order to align the dates that have an indication for those who don't.

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_609/

sivanharel:
> Also consider adding a visual test (or maybe already exists?)

There are a lot of tests but not for the case that the font size is not the default.
I agree that it could be nice to test it but I don't know how?
Actually, it is changed according to the font-size override. Usually, we don't test overrides but maybe we should find a solution to do it. If you or @jonathanadler have an idea of how to test it, I would love to add a test, but if it's required a lot of work maybe we could open another task:)

jonathanadler:
> > Also consider adding a visual test (or maybe already exists?)
> 
> There are a lot of tests but not for the case that the font size is not the default.
> I agree that it could be nice to test it but I don't know how?
> Actually, it is changed according to the font-size override. Usually, we don't test overrides but maybe we should find a solution to do it. If you or @jonathanadler have an idea of how to test it, I would love to add a test, but if it's required a lot of work maybe we could open another task:)

We have a task to add "overridden" visual tests.
Unfortunately it's not that straight forward.

nitayneeman:
> > Also consider adding a visual test (or maybe already exists?)
> 
> There are a lot of tests but not for the case that the font size is not the default.
> I agree that it could be nice to test it but I don't know how?
> Actually, it is changed according to the font-size override. Usually, we don't test overrides but maybe we should find a solution to do it. If you or @jonathanadler have an idea of how to test it, I would love to add a test, but if it's required a lot of work maybe we could open another task:)

I think we've a visual test with style overriding in Badge:
https://github.com/wix/wix-ui-tpa/blob/master/src/components/Badge/Badge.visual.tsx#L42-L55

Maybe we can do the same? 
Passing a className to the component (inside the `snap`) which uses `-st-mixin` to override the default font-size value, WDYT?

sivanharel:
> > > Also consider adding a visual test (or maybe already exists?)
> > 
> > 
> > There are a lot of tests but not for the case that the font size is not the default.
> > I agree that it could be nice to test it but I don't know how?
> > Actually, it is changed according to the font-size override. Usually, we don't test overrides but maybe we should find a solution to do it. If you or @jonathanadler have an idea of how to test it, I would love to add a test, but if it's required a lot of work maybe we could open another task:)
> 
> I think we've a visual test with style overriding in Badge:
> https://github.com/wix/wix-ui-tpa/blob/master/src/components/Badge/Badge.visual.tsx#L42-L55
> 
> Maybe we can do the same?
> Passing a className to the component (inside the `snap`) which uses `-st-mixin` to override the default font-size value, WDYT?

Cool, I will try to add tests in the same way:)

nitayneeman:
> > > > Also consider adding a visual test (or maybe already exists?)
> > > 
> > > 
> > > There are a lot of tests but not for the case that the font size is not the default.
> > > I agree that it could be nice to test it but I don't know how?
> > > Actually, it is changed according to the font-size override. Usually, we don't test overrides but maybe we should find a solution to do it. If you or @jonathanadler have an idea of how to test it, I would love to add a test, but if it's required a lot of work maybe we could open another task:)
> > 
> > 
> > I think we've a visual test with style overriding in Badge:
> > https://github.com/wix/wix-ui-tpa/blob/master/src/components/Badge/Badge.visual.tsx#L42-L55
> > Maybe we can do the same?
> > Passing a className to the component (inside the `snap`) which uses `-st-mixin` to override the default font-size value, WDYT?
> 
> Cool, I will try to add tests in the same way:)

You can merge and do it separately or create a task if you think it might take time - but just give it a try as you said 😬

PR #611, opened 2021-01-17
<BoxSelection /> - contribute new component

roni6chabra:
# ✨ Pull Request

### 💁 Description

...

### 💭 Motivation

...

### ☝️ TODOs 

...

### 👀  PR is ready for review 

- [ ] Ready for review

### DIFF
--- a/src/components/BoxSelection/BoxSelection.driver.ts
+++ b/src/components/BoxSelection/BoxSelection.driver.ts
@@ -16,8 +16,8 @@ export const boxSelectionDriverFactory = (
   base: UniDriver,
 ): BoxSelectionDriver => {
   const boxOptionDataHook = `[data-hook=${BOX_SELECTION_DATA_HOOKS.BOX_SELECTION_OPTION}]`;
-  const getOptionById = async (id: string) =>
-    base.$$(`[data-id="${id}"]`).get(0);
+  const getOptionById = async (id: string) => base.$(`[id="${id}"]`);
+  const getOptionByDataId = async (id: string) => base.$(`[data-id="${id}"]`);
   return {
     ...baseUniDriverFactory(base),
     async getOptionsCount() {
@@ -25,21 +25,21 @@ export const boxSelectionDriverFactory = (
     },
     async isChecked(id: string) {
       const option = await getOptionById(id);
-      const isChecked = (await option.attr('data-checked')) === 'true';
+      const isChecked = await option._prop('checked');
       return isChecked;
     },
     async isDisabled(id: string) {
       const option = await getOptionById(id);
-      const isDisabled = (await option.attr('data-disabled')) === 'true';
+      const isDisabled = await option._prop('disabled');
       return isDisabled;
     },
     async isUnavailable(id: string) {
-      const option = await getOptionById(id);
+      const option = await getOptionByDataId(id);
       const isUnavailable = (await option.attr('data-unavailable')) === 'true';
       return isUnavailable;
     },
     async clickOnOption(id: string) {
-      const option = base.$(`[data-id="${id}"] input`);
+      const option = await getOptionById(id);
       if (!(await this.isDisabled(id)) && !(await this.isUnavailable(id))) {
         await option.click();
       }
--- a/src/components/BoxSelection/BoxSelection.driver.ts
+++ b/src/components/BoxSelection/BoxSelection.driver.ts
@@ -10,7 +10,6 @@ export interface BoxSelectionDriver extends BaseUniDriver {
   isDisabled(id: string): Promise<boolean>;
   isUnavailable(id: string): Promise<boolean>;
   getOptionsCount(): Promise<number>;
-  clickOnOption(id: string): Promise<void>;
 }
 export const boxSelectionDriverFactory = (
   base: UniDriver,
@@ -37,12 +36,6 @@ export const boxSelectionDriverFactory = (
       const option = await getOptionByDataId(id);
       const isUnavailable = (await option.attr('data-unavailable')) === 'true';
       return isUnavailable;
-    },
-    async clickOnOption(id: string) {
-      const option = await getOptionById(id);
-      if (!(await this.isDisabled(id)) && !(await this.isUnavailable(id))) {
-        await option.click();
-      }
     },
   };
 };
--- a/src/components/BoxSelection/BoxSelection.driver.ts
+++ b/src/components/BoxSelection/BoxSelection.driver.ts
@@ -3,46 +3,44 @@ import {
   baseUniDriverFactory,
 } from 'wix-ui-test-utils/base-driver';
 import { UniDriver } from 'wix-ui-test-utils/unidriver';
-import {
-  BOX_SELECTION_DATA_ATTRIBUTES,
-  BOX_SELECTION_DATA_HOOKS,
-} from './dataHooks';
+import { BOX_SELECTION_DATA_HOOKS } from './dataHooks';
 
 export interface BoxSelectionDriver extends BaseUniDriver {
-  isChecked(): Promise<boolean>;
-  isDisabled(): Promise<boolean>;
-  isUnavailable(): Promise<boolean>;
+  isChecked(id: string): Promise<boolean>;
+  isDisabled(id: string): Promise<boolean>;
+  isUnavailable(id: string): Promise<boolean>;
   getOptionsCount(): Promise<number>;
-  click(): Promise<void>;
+  clickOnOption(id: string): Promise<void>;
 }
 export const boxSelectionDriverFactory = (
   base: UniDriver,
 ): BoxSelectionDriver => {
   const boxOptionDataHook = `[data-hook=${BOX_SELECTION_DATA_HOOKS.BOX_SELECTION_OPTION}]`;
-  const boxOptionWrapperDataHook = `[data-hook=${BOX_SELECTION_DATA_HOOKS.BOX_SELECTION_OPTION_WRAPPER}]`;
-  const getOption = async () => base.$('input').getNative();
-  const getOptionWrapper = async () => base.$(boxOptionWrapperDataHook);
+  const getOptionById = async (id: string) =>
+    base.$$(`[data-id="${id}"]`).get(0);
   return {
     ...baseUniDriverFactory(base),
     async getOptionsCount() {
       return base.$$(boxOptionDataHook).count();
     },
-    async isChecked() {
-      const option = await getOption();
-      return option.checked;
+    async isChecked(id: string) {
+      const option = await getOptionById(id);
+      const isChecked = (await option.attr('data-checked')) === 'true';
+      return isChecked;
     },
-    async isDisabled() {
-      const option = await getOption();
-      return option.disabled;
+    async isDisabled(id: string) {
+      const option = await getOptionById(id);
+      const isDisabled = (await option.attr('data-disabled')) === 'true';
+      return isDisabled;
     },
-    async isUnavailable() {
-      const optionWrapper = await getOptionWrapper();
-      const unavailableAttr = await optionWrapper.attr('data-unavailable');
-      return unavailableAttr === 'true';
+    async isUnavailable(id: string) {
+      const option = await getOptionById(id);
+      const isUnavailable = (await option.attr('data-unavailable')) === 'true';
+      return isUnavailable;
     },
-    async click() {
-      const option = await getOption();
-      if (!option.disabled) {
+    async clickOnOption(id: string) {
+      const option = base.$(`[data-id="${id}"] input`);
+      if (!(await this.isDisabled(id)) && !(await this.isUnavailable(id))) {
         await option.click();
       }
     },
--- a/src/components/BoxSelection/BoxSelection.driver.ts
+++ b/src/components/BoxSelection/BoxSelection.driver.ts
@@ -1,0 +1,50 @@
+import {
+  BaseUniDriver,
+  baseUniDriverFactory,
+} from 'wix-ui-test-utils/base-driver';
+import { UniDriver } from 'wix-ui-test-utils/unidriver';
+import {
+  BOX_SELECTION_DATA_ATTRIBUTES,
+  BOX_SELECTION_DATA_HOOKS,
+} from './dataHooks';
+
+export interface BoxSelectionDriver extends BaseUniDriver {
+  isChecked(): Promise<boolean>;
+  isDisabled(): Promise<boolean>;
+  isUnavailable(): Promise<boolean>;
+  getOptionsCount(): Promise<number>;
+  click(): Promise<void>;
+}
+export const boxSelectionDriverFactory = (
+  base: UniDriver,
+): BoxSelectionDriver => {
+  const boxOptionDataHook = `[data-hook=${BOX_SELECTION_DATA_HOOKS.BOX_SELECTION_OPTION}]`;
+  const boxOptionWrapperDataHook = `[data-hook=${BOX_SELECTION_DATA_HOOKS.BOX_SELECTION_OPTION_WRAPPER}]`;
+  const getOption = async () => base.$('input').getNative();
+  const getOptionWrapper = async () => base.$(boxOptionWrapperDataHook);
+  return {
+    ...baseUniDriverFactory(base),
+    async getOptionsCount() {
+      return base.$$(boxOptionDataHook).count();
+    },
+    async isChecked() {
+      const option = await getOption();
+      return option.checked;
+    },
+    async isDisabled() {
+      const option = await getOption();
+      return option.disabled;
+    },
+    async isUnavailable() {
+      const optionWrapper = await getOptionWrapper();
+      const unavailableAttr = await optionWrapper.attr('data-unavailable');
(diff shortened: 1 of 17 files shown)


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_611/

src/components/BoxSelection/BoxSelection.tsx
```diff
@@ -0,0 +1,117 @@
...
+   */
+  'aria-labelledby'?: string;
+  /**
+   * The onChange callback
+   */
+  onChange?(id: string): void;
+  /**
+   * The Box selected option value
+   */
+  value?: string;
```
jonathanadler, line 36: So this prop isn't needed in the `BoxSelection` component

src/components/BoxSelection/BoxSelection.tsx
```diff
@@ -0,0 +1,117 @@
...
+      size,
+      name,
+      ['aria-label']: ariaLabel,
+      ['aria-labelledby']: ariaLabelledBy,
+    } = this.props;
+
+    //const { value } = this.state;
+    return (
+      <TPAComponentsConsumer>
+        {() => (
```
jonathanadler, line 82: Since we're not using anything from the context, we can remove these lines

src/components/BoxSelection/BoxSelection.tsx
```diff
@@ -0,0 +1,117 @@
...
+                return (
+                  React.isValidElement(child) && (
+                    <>
+                      {React.cloneElement(child, {
+                        key,
+                        id: `"${key}"`,
+                        name,
+                        size,
+                        checked: child.props.id === this.props.value,
+                        className: 'boxSelectionOption',
```
jonathanadler, line 101: This passes a global `boxSelectionOption` class.
I don't think we want that.
Also I didn't see a reference to this class anywhere

src/components/BoxSelection/BoxSelection.tsx
```diff
@@ -0,0 +1,117 @@
...
+              (child: BoxSelectionOption, key: number) => {
+                return (
+                  React.isValidElement(child) && (
+                    <>
+                      {React.cloneElement(child, {
+                        key,
+                        id: `"${key}"`,
+                        name,
+                        size,
+                        checked: child.props.id === this.props.value,
```
jonathanadler, line 100: So this part isn't needed.
It's enough to pass `checked` directly to the Option component.

So for instance, when using the component, I would write something like this:

```tsx
function myApplication () {
  const [selectedId, setSelectedId] = React.useState('1');

  return (
    <BoxSelection onChange={({ id }) => setSelectedId(id)} name="myApp">
      <BoxSelection.Option checked={selectedId === '1'}>Item 1</BoxSelection.Option>
      <BoxSelection.Option checked={selectedId === '2'}>Item 2</BoxSelection.Option>
      <BoxSelection.Option checked={selectedId === '3'}>Item 3</BoxSelection.Option>
    </BoxSelection>
  );
}

File: src/components/BoxSelection/BoxSelection.tsx [hunk 5]
```diff
@@ -0,0 +1,117 @@
...
+                        className: 'boxSelectionOption',
+                        onChange: (e) => {
+                          this.props.onChange(e.value);
+                          //this._onChange(e.value);
+                        },
```
jonathanadler, line 105: So actually, instead of passing the `onChange` method to the children, we can use event delegation, and add the `onChange` callback directly on the wrapping `div`.

So it would look like this:
```tsx
render() {
  ...
  return (
    <div className={st(classes.root, {}, className)}
            aria-label={ariaLabel}
            aria-labelledby={ariaLabelledBy}
            onChange={e => this.props.onChange({id: e.target.getAttribute('data-id')})}
            >
      {React.Children.map(
        ...
      )}
    <div>
  );
}
```
I used the `data-id` attribute to save the `id` on the element itself

src/components/BoxSelection/Option.tsx
```diff
@@ -0,0 +1,146 @@
...
+              },
+              focused ? classes.focused : '',
+              className,
+            )}
+            data-hook={BOX_SELECTION_DATA_HOOKS.BOX_SELECTION_OPTION_WRAPPER}
+            {...this._getDataAttributes()}
+          >
+            <CoreRadioButton
+              data-hook={BOX_SELECTION_DATA_HOOKS.BOX_SELECTION_OPTION}
+              tabIndex={0}
```
jonathanadler, line 129: tabIndex isn't needed in this case.
The radioButton is tabbable by default

src/components/BoxSelection/Option.tsx
```diff
@@ -0,0 +1,146 @@
...
+              name={name}
+              label={children}
+              checked={checked}
+              value={id}
+              onChange={onChange}
```
jonathanadler, line 134: So the `onChange` isn't needed here

jonathanadler (commented):
First of all, the PR looks really great!

I haven't gone over everything yet, but I've already added some comments about the `checked` logic, so that you can start going over them while I continue going over the rest of the PR.

Greta job!

src/components/BoxSelection/BoxSelection.tsx
roni6chabra: So I change it as you described and changed it to '**onClick**' event since the div that wraps the children is not triggered on the 'Change' event.

There is still some problem (that you can see on the Settings panel) when the user clicks on the button, it works fine, but when he clicks on the children button (the text), it doesn't receive the data-id, and therefore, there is no change.

So I'm trying to figure out what can we do here. 
If you tackled this before please let me know.

jonathanadler (commented):
Ok, so I've added some comments.
It looks like a lot, but they're mostly tiny things.

In general it's a very good PR, nice!

src/components/BoxSelection/BoxSelection.driver.ts
```diff
@@ -0,0 +1,50 @@
...
+    async getOptionsCount() {
+      return base.$$(boxOptionDataHook).count();
+    },
+    async isChecked() {
+      const option = await getOption();
+      return option.checked;
+    },
+    async isDisabled() {
+      const option = await getOption();
+      return option.disabled;
```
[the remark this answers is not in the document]
roni6chabra: As you can see, I used '**attr**' and not '**prop**'
Is it ok? 
for some reason, I received undefined on the checked prop that I was passing through the test.

src/components/BoxSelection/Option.st.css
```diff
@@ -0,0 +1,159 @@
...
+    CurrentBorderColor: color(fallback(value(BorderColor), value(DefaultBorderColor)));
+    CurrentBoxFillOpacity: fallback(value(BoxFillOpacity), value(DefaultBoxFillOpacity));
+    CurrentBoxBorderWidth: fallback(value(BoxBorderWidth), value(DefaultBoxBorderWidth));
+    CurrentBoxBorderRadius: fallback(value(BoxBorderRadius), value(DefaultBoxBorderRadius));
+    CurrentBoxPadding: fallback(value(BoxPadding), value(DefaultBoxPadding));
+    CurrentSelectedBoxColor: color(fallback(value(SelectedBoxColor), value(DefaultSelectedBoxColor)));
+}
+
+.focused {
+    /* -st-mixin: focusRing(); */
```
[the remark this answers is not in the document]
roni6chabra: I removed the box-shadow,
the focused box is shown when the user navigates through the TAB key (you can see it on the playground with the example of 10 slots)
is it ok?

src/components/BoxSelection/Option.st.css
```diff
@@ -0,0 +1,159 @@
...
+.root:unavailable .wrapper{
+    pointer-events: none;
+    cursor: default;
+    border-color: color(value(DisabledColor));
+    background: linear-gradient(to top left,
+        color(value(DefaultBoxFillColor)) 0%,
+        color(value(DefaultBoxFillColor)) calc(50% - 0.8px),
+        rgba(0,0,0,1) 50%,
+        color(value(DefaultBoxFillColor)) calc(50% + 0.8px),
+        color(value(DefaultBoxFillColor)) 100%)
```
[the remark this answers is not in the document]
roni6chabra: Thanks,
I checked this line and it seems that this line is above the text and not beneath.
I checked this one on the EDGE browser on windows and it works fine.
When I tried to open this link on explorer - it immediately opened my EDGE browser again.
Do you have access to ie11 to check it out with the link that you provided?

src/components/BoxSelection/docs/BoxSelectionWiringExample.tsx
```diff
@@ -0,0 +1,27 @@
...
+    <BoxSelection
+      name="hours"
+      className={classes.root}
+      onChange={({ id }) => {
+        return setSelectedId(id);
+      }}
+    >
+      {[1, 2, 3].map((_n, i) => (
+        <BoxSelection.Option
+          key={i}
```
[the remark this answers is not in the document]
roni6chabra: I didn't remove it since React yells at me that I must have a unique key prop on the map functionality.
if you still want me to remove it please write again.

src/components/BoxSelection/BoxSelection.tsx
roni6chabra: So I added to the child to have the 'onChange' event as you suggested in a different comment.

src/components/BoxSelection/BoxSelection.driver.ts
```diff
@@ -0,0 +1,50 @@
...
+  isUnavailable(): Promise<boolean>;
+  getOptionsCount(): Promise<number>;
+  click(): Promise<void>;
+}
+export const boxSelectionDriverFactory = (
+  base: UniDriver,
+): BoxSelectionDriver => {
+  const boxOptionDataHook = `[data-hook=${BOX_SELECTION_DATA_HOOKS.BOX_SELECTION_OPTION}]`;
+  const boxOptionWrapperDataHook = `[data-hook=${BOX_SELECTION_DATA_HOOKS.BOX_SELECTION_OPTION_WRAPPER}]`;
+  const getOption = async () => base.$('input').getNative();
```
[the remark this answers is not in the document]
roni6chabra: As you can see now I'm getting the element using 
  `const getOptionById = async (id: string) => base.$$(`[data-id="${id}"]`).get(0);`
is it ok?
I couldn't find any other way to get the element itself.

src/components/BoxSelection/BoxSelection.driver.ts
```diff
@@ -0,0 +1,50 @@
...
+  const getOptionWrapper = async () => base.$(boxOptionWrapperDataHook);
+  return {
+    ...baseUniDriverFactory(base),
+    async getOptionsCount() {
+      return base.$$(boxOptionDataHook).count();
+    },
+    async isChecked() {
+      const option = await getOption();
+      return option.checked;
```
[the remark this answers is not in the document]
roni6chabra: As you can see, I used 'attr' and not 'prop' like so;
`const isChecked = (await option.attr('data-checked')) === 'true';`
Is it ok?
for some reason, I received undefined on the checked prop that I was passing through the test.

src/components/BoxSelection/BoxSelection.st.css
```diff
@@ -0,0 +1,29 @@
...
+/* Overrides */
+:vars {
+  boxSelectionGap: applyUnit(fallback(value(ItemGap), value(defaultItemGap)), px);
+}
+
+.root {
+  display: flex;
+  flex-wrap: wrap;
+  gap: value(boxSelectionGap);
+}
```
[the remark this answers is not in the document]
roni6chabra: Changed!
But can you please tell me the reason to have multiple codes of the GAP CSS style? (on root and on overrideStyleParams)
the _boxSelectionGap_ param is already having a fallback on the **:vars** overrides

src/components/BoxSelection/Option.tsx
```diff
@@ -0,0 +1,156 @@
...
+  _onChange = () => {
+    const { id, onChange } = this.props;
+
+    if (onChange) {
+      onChange({ id });
+    }
+  };
+  _getAriaLabel = () => {
+    const { children } = this.props;
+    return this.props['aria-label'] ? this.props['aria-label'] : children;
```
jonathanadler, line 100: In this case I think we should use children only if it's a string
So something like this:
```js
    return this.props['aria-label'] ? this.props['aria-label'] : (typeof children === 'string' ? children : null);
```

jonathanadler: If we really want, we can create a kind of "hack" in the future, where we create a ref for the radio button wrapping element, and then take the element's [textContent](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent) but I'm not sure it's necessary now

src/components/BoxSelection/Option.tsx
```diff
@@ -0,0 +1,156 @@
...
+            disabled,
+            unavailable,
+          },
+          classnames({
+            [classes.focused]: focused,
+          }),
+        )}
+        data-id={id}
+        data-hook={BOX_SELECTION_DATA_HOOKS.BOX_SELECTION_OPTION_WRAPPER}
+        tabIndex={0}
```
jonathanadler, line 131: This tabIndex should be removed, since the core radio button already handles focus

src/components/BoxSelection/Option.st.css
```diff
@@ -0,0 +1,211 @@
...
+    /*
+        Text color of the box
+        @default color-5
+    */
+    TextColor: --overridable;
+    /*
+        Text color of the box when disabled or Unavailable
+        @default color-3
+    */
+    DisabledTextColor: --overridable;
```
jonathanadler, line 71: I'm not sure Johanna meant we need to open a variable for the disabled color, but it's ok to leave it now.
I'm checking with her

jonathanadler (commented):
Some very minor comments, but really, the PR looks great! 
Nice and clean coding 👏 👏 👏

src/components/BoxSelection/Option.st.css
jonathanadler: so I checked with @johannaasseraf, and she said don't need this variable.
Her comment was about having the correct default color (color-3), but it doesn't need to be overridable by a variable

src/components/BoxSelection/BoxSelection.driver.ts
```diff
@@ -0,0 +1,48 @@
...
+  isUnavailable(id: string): Promise<boolean>;
+  getOptionsCount(): Promise<number>;
+  clickOnOption(id: string): Promise<void>;
+}
+export const boxSelectionDriverFactory = (
+  base: UniDriver,
+): BoxSelectionDriver => {
+  const boxOptionDataHook = `[data-hook=${BOX_SELECTION_DATA_HOOKS.BOX_SELECTION_OPTION}]`;
+  const getOptionById = async (id: string) => base.$(`[id="${id}"]`);
+  const getOptionByDataId = async (id: string) => base.$(`[data-id="${id}"]`);
```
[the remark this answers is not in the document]
roni6chabra: yes because **id** is for the native radio button and **data-id** is for the div that wraps the radio.
I needed the **data-id** because I need to get the '**unavailable**' attribute.
since it does not exist on the native.

src/components/BoxSelection/BoxSelection.driver.ts
```diff
@@ -0,0 +1,48 @@
...
+      return isDisabled;
+    },
+    async isUnavailable(id: string) {
+      const option = await getOptionByDataId(id);
+      const isUnavailable = (await option.attr('data-unavailable')) === 'true';
+      return isUnavailable;
+    },
+    async clickOnOption(id: string) {
+      const option = await getOptionById(id);
+      if (!(await this.isDisabled(id)) && !(await this.isUnavailable(id))) {
```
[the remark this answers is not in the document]
roni6chabra: removed

src/components/BoxSelection/Option.st.css
roni6chabra: removed

jonathanadler (commented):
@roni6chabra Hey, so the tests are finally passing (we had some issues with applitools).

I just left some minor comments that need some attention, but after that we can merge (I'll also add the needed changes we talked about to wix-ui-core).

Great work and thanks!

src/components/BoxSelection/Option.st.css
```diff
@@ -0,0 +1,241 @@
...
+    CurrentBoxBorderRadius: fallback(value(BoxBorderRadius), value(DefaultBoxBorderRadius));
+    CurrentBoxPadding: fallback(value(BoxPadding), value(DefaultBoxPadding));
+    CurrentSelectedBoxColor: color(fallback(value(SelectedBoxColor), value(DefaultSelectedBoxColor)));
+    CurrentDisabledColor: color(fallback(value(DisabledColor), value(DefaultDisabledColor)));
+}
+
+input[type='radio'] {
+    opacity: 1;
+    clip: rect(1px, 1px, 1px, 1px);
+}
```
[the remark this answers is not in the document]
roni6chabra: you are definitely right!

jonathanadler (approved):

PR #612, opened 2021-01-17
Radio button remove focus ring on mobile

urig-wix:
# ✨ Pull Request

### 💁 Description

We'd like to use the radioButton component but we get a focus ring on mobile after a regular tapping which looks odd. 
More details available here -> https://jira.wixpress.com/browse/EE-28048.

...

### 💭 Motivation

Using TPAComponentsContext was suggested as a relatively immediate and simple solution. We'd be happy to unblock this problem so we'd be able to use the component.

...

### ☝️ TODOs 

...

### 👀  PR is ready for review 

- [ ] Ready for review

### DIFF
--- a/src/components/RadioButton/docs/RadioButtonTestStory.tsx
+++ b/src/components/RadioButton/docs/RadioButtonTestStory.tsx
@@ -1,6 +1,6 @@
 import { storiesOf } from '@storybook/react';
 import * as React from 'react';
-import { TPAComponentsWrapper } from 'src/test/utils';
+import { TPAComponentsWrapper } from '../../../test/utils';
 import { RadioButton, RadioButtonTheme } from '..';
 
 const kind = 'Tests';
--- a/src/components/RadioButton/docs/RadioButtonTestStory.tsx
+++ b/src/components/RadioButton/docs/RadioButtonTestStory.tsx
@@ -38,7 +38,7 @@ function renderTest(props?: any) {
             withFocusRing
             onChange={(val) => console.log(val)}
             theme={RadioButtonTheme.Default}
-            label="default theme"
+            label="no focus-ring on mobile"
             data-hook="radio-button-default-mobile"
           />,
         )}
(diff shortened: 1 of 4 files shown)


urig-wix:
@jonathanadler hopefully I got the idea of using 'TPAComponentsContext' right.
w questions regarding it - 
1. what's the right way to test that?
2. I ran it (using npm start) and when I simulated a mobile and debugged seems like isMobile was set to false. Am I doing something wrong?

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_612/

jonathanadler:
> @jonathanadler hopefully I got the idea of using 'TPAComponentsContext' right.
> w questions regarding it -
> 
> 1. what's the right way to test that?
> 2. I ran it (using npm start) and when I simulated a mobile and debugged seems like isMobile was set to false. Am I doing something wrong?

The PR looks great 👍 

Another option is using the `TPAComponentsConsumer` to consume the `mobile` value.
You can see it used in many other components, but using `contextType` is also great.

In order to test the component, you'll need to wrap it with the `TPAComponentsProvider` with `{mobile: true}` so that it will be passed down to the component.
A visual test would sufice.

Thanks for the contribution!

jonathanadler (approved):

src/components/RadioButton/docs/RadioButtonTestStory.tsx
```diff
@@ -29,6 +30,19 @@ function renderTest(props?: any) {
...
       </div>
+      <div style={{ marginTop: '50px' }}>
+        {TPAComponentsWrapper({ mobile: true })(
+          <RadioButton
+            value={'Checked'}
+            checked
+            withFocusRing
+            onChange={(val) => console.log(val)}
+            theme={RadioButtonTheme.Default}
+            label="default theme"
```
jonathanadler, line 41: Maybe it would be better to use a more descriptive label here:
```suggestion
            label="no focus-ring on mobile"
```

PR #616, opened 2021-01-19
Image - calculate dimensions

nitayneeman:
# ✨ Pull Request

### 💁 Description

This PR implements the dimension calculations based on this logic:

1. Given fixed dimensions ⇒ use them
2. Given fixed dimension and aspectRatio ⇒ use them to calculate the missing one
3. Missing dimensions ⇒ take from parent 
4. Missing dimensions but given aspect ratio ⇒ take from parent the width and calculate using the aspect ratio

Also, the PR adjusts the docs & tests and improves the code readability.

### 👀  PR is ready for review 

- [X] Ready for review

### DIFF
--- a/src/components/Image/Image.spec.tsx
+++ b/src/components/Image/Image.spec.tsx
@@ -16,19 +16,14 @@ describe('Image', () => {
       'https://m.media-amazon.com/images/M/MV5BZGMwOGIwZjUtOWQ1OS00YWRjLWJmZGMtN2Y1OWQ3ZDYwYTM3XkEyXkFqcGdeQXVyNzU1NzE3NTg@._V1_.jpg',
     relative: 'c5f754_dd75514d14fa4057b4f4a6cc8ce7add3~mv2.jpg',
   };
+  const sampleDimensions = { width: 480, height: 360 };
   const sampleAlt = 'Garfield smiles and puts his hand over chest';
-
-  it('should render', async () => {
-    const driver = createDriver(<Image src={sampleSources.absolute} />);
-
-    expect(await driver.exists()).toBe(true);
-  });
 
   describe('absolute', () => {
     const src = sampleSources.absolute;
 
     it('should return `src` as is', async () => {
-      const driver = createDriver(<Image src={src} />);
+      const driver = createDriver(<Image src={src} {...sampleDimensions} />);
 
       expect(await driver.getSrc()).toBe(src);
     });
@@ -48,7 +43,7 @@ describe('Image', () => {
     const src = sampleSources.relative;
 
     it('should return `src` as full media item URL containing the provided URI', async () => {
-      const driver = createDriver(<Image src={src} />);
+      const driver = createDriver(<Image src={src} {...sampleDimensions} />);
 
       const mediaItemUrl = await driver.getSrc();
 
@@ -60,8 +55,16 @@ describe('Image', () => {
   describe.each(Object.keys(sampleSources))('%s', (sourceType) => {
     const src = sampleSources[sourceType];
 
+    it('should render', async () => {
+      const driver = createDriver(<Image src={src} {...sampleDimensions} />);
+
+      expect(await driver.exists()).toBe(true);
+    });
+
     it('should return `alt`', async () => {
-      const driver = createDriver(<Image src={src} alt={sampleAlt} />);
+      const driver = createDriver(
+        <Image src={src} {...sampleDimensions} alt={sampleAlt} />,
+      );
 
       expect(await driver.getAlt()).toBe(sampleAlt);
     });
--- a/src/components/Image/Image.st.css
+++ b/src/components/Image/Image.st.css
@@ -12,8 +12,6 @@
 }
 
 .root {
-  -st-states: preload, loaded;
-
   display: grid;
   min-height: 100%;
 }
@@ -26,11 +24,11 @@
   object-fit: cover;
 }
 
-.root:preload .absoluteImage {
+.absoluteImage.preload {
   filter: blur(value(BlurRadius));
 }
 
-.root:loaded .absoluteImage {
+.absoluteImage.loaded {
   animation-name: blurry-loading;
   animation-duration: 0.3s;
   animation-timing-function: easeout;
--- a/src/components/Image/Image.st.css
+++ b/src/components/Image/Image.st.css
@@ -24,12 +24,20 @@
   object-fit: cover;
 }
 
-.absoluteImage.preload {
+.preload {
   filter: blur(value(BlurRadius));
 }
 
-.absoluteImage.loaded {
+.loaded {
   animation-name: blurry-loading;
   animation-duration: 0.3s;
   animation-timing-function: easeout;
+}
+
+.absoluteImage.preload {
+  -st-mixin: preload;
+}
+
+.absoluteImage.loaded {
+  -st-mixin: loaded;
 }
--- a/src/components/Image/Image.st.css
+++ b/src/components/Image/Image.st.css
@@ -15,6 +15,7 @@
   -st-states: preload, loaded;
 
   display: grid;
+  min-height: 100%;
 }
 
 .root.contain .absoluteImage, .root.contain .relativeImage, .root.contain .placeholder {
@@ -25,15 +26,11 @@
   object-fit: cover;
 }
 
-.placeholder, .relativeImage {
-  grid-area: 1 / 1;
-}
-
-.root:preload .absoluteImage, .root:preload .placeholder {
+.root:preload .absoluteImage {
   filter: blur(value(BlurRadius));
 }
 
-.root:loaded .absoluteImage, .root:loaded .relativeImage {
+.root:loaded .absoluteImage {
   animation-name: blurry-loading;
   animation-duration: 0.3s;
   animation-timing-function: easeout;
--- a/src/components/Image/Image.st.css
+++ b/src/components/Image/Image.st.css
@@ -18,11 +18,11 @@
   min-height: 100%;
 }
 
-.root.contain .absoluteImage, .root.contain .relativeImage, .root.contain .placeholder {
+.root.contain .absoluteImage, .root.contain .relativeImage {
   object-fit: contain;
 }
 
-.root.cover .absoluteImage, .root.cover .relativeImage, .root.cover .placeholder {
+.root.cover .absoluteImage, .root.cover .relativeImage {
   object-fit: cover;
 }
 
--- a/src/components/Image/Image.st.css
+++ b/src/components/Image/Image.st.css
@@ -16,11 +16,13 @@
   min-height: 100%;
 }
 
-.root.contain .absoluteImage, .root.contain .relativeImage {
+.root.contain .absoluteImage,
+.root.contain .relativeImage {
   object-fit: contain;
 }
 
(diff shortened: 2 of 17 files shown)


src/components/Image/Image.spec.tsx
```diff
@@ -16,19 +16,14 @@ describe('Image', () => {
       'https://m.media-amazon.com/images/M/MV5BZGMwOGIwZjUtOWQ1OS00YWRjLWJmZGMtN2Y1OWQ3ZDYwYTM3XkEyXkFqcGdeQXVyNzU1NzE3NTg@._V1_.jpg',
     relative: 'c5f754_dd75514d14fa4057b4f4a6cc8ce7add3~mv2.jpg',
   };
+  const sampleDimensions = { width: 480, height: 360 };
```
nitayneeman, line 19: We need some dimensions (fixed/one+aspectRatio) to render the component properly

src/components/Image/Image.st.css
```diff
@@ -12,29 +12,34 @@
 }
 
 .root {
-  -st-states: preload, loaded;
```
nitayneeman, line 15: I removed the st-states because we don't want to use it anymore... So now I'm passing simply classes

src/components/Image/Image.tsx
```diff
@@ -1,84 +1,22 @@
...
-  />
-);
+type DefaultProps = Pick<ImageProps, 'resize'>;
 
 /** Image is a component to literally display an image - whether an absolute with full URL or a media platform item with relative URI */
 export class Image extends React.Component<ImageProps> {
   static displayName = 'Image';
   static defaultProps: DefaultProps = {
     resize: ResizeOptions.contain,
-    aspectRatio: 1,
```
nitayneeman, line 78: This was a mistake 😅

src/components/Image/Image.tsx
```diff
@@ -102,71 +59,63 @@ export class Image extends React.Component<ImageProps> {
...
 
     return (
       <div
+        ref={this.containerRef}
         className={st(
           classes.root,
-          {
-            resize,
-            ...(hasLoadingBehavior && { preload: !isLoaded, loaded: isLoaded }),
-          },
```
nitayneeman, line 135: As I said... I removed the st-states

src/components/Image/Image.tsx
```diff
@@ -102,71 +59,63 @@ export class Image extends React.Component<ImageProps> {
...
-            {...imageProps}
-            nativeProps={{ ...containerDimensions }}
             src={src}
-            className={classes.absoluteImage}
+            className={st(
+              classes.absoluteImage,
+              hasLoadingBehavior &&
+                (isLoaded ? classes.loaded : classes.preload),
+            )}
+            {...(calculatedDimensions && {
```
nitayneeman, line 103: For an absolute image it might work without dimensions at all, but I decided to calculate them as well in this case to make the image look consistently

src/components/Image/RelativeMediaImage/RelativeMediaImage.tsx
```diff
@@ -0,0 +1,73 @@
...
+      className,
+      sourceDimensions,
+      containerDimensions,
+      isPlaceholderDisplayed,
+      onLoad,
+      ...imageProps
+    } = this.props;
+
+    return (
+      containerDimensions && (
```
nitayneeman, line 48: Actually without these dimensions we cannot render at all...

But we make sure that the component does get them (from fixed props/bounding rect). This condition is important because in case of the bounding rect, it happens in the didMount so we need to block the render until we have them (otherwise ui-core's MediaImage will fail to render).

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_616/

src/components/Image/Image.tsx
```diff
@@ -102,71 +59,66 @@ export class Image extends React.Component<ImageProps> {
...
           <CoreImage
-            {...imageProps}
-            nativeProps={{ ...containerDimensions }}
             src={src}
-            className={classes.absoluteImage}
+            className={st(
+              classes.absoluteImage,
+              hasLoadingBehavior &&
+                (isLoaded ? classes.loaded : classes.preload),
+            )}
```
jonathanadler, line 102: `st` isn't needed here, but only for the root (or when using states).
You can use `classnames` for this:
```suggestion
            className={classnames(
              classes.absoluteImage,
              { 
                [classes.loaded]: hasLoadingBehavior && isLoaded,
                [classes.preload]: hasLoadingBehavior && !isLoaded, 
              })}
```

src/components/Image/Image.tsx
```diff
@@ -102,71 +59,66 @@ export class Image extends React.Component<ImageProps> {
...
-                src={src}
-                sourceDimensions={sourceDimensions}
-                containerDimensions={containerDimensions}
-              />
+          <RelativeMediaImage
+            src={src}
+            className={st(
+              classes.relativeImage,
+              hasLoadingBehavior && isLoaded && classes.loaded,
             )}
```
jonathanadler, line 115: Same here

src/components/Image/types.ts
```diff
@@ -0,0 +1,38 @@
...
+  /** The intrinsic height of the image in pixels */
+  height?: number;
+  /** The alternative text description of the image. Allowing better SEO when the image has meaning of content */
+  alt?: string;
+  /** A callback to be called when the image is loaded */
+  onLoad?: React.EventHandler<React.SyntheticEvent>;
+  /** A callback to be called if error occurs while loading */
+  onError?: React.EventHandler<React.SyntheticEvent>;
+  /** Specifies how the image is resized to fit its container */
+  resize?: 'contain' | 'cover';
```
jonathanadler, line 17:
```suggestion
  resize?: ResizeOptions;
```

src/components/Image/types.ts
```diff
@@ -0,0 +1,38 @@
...
+  // Specifies the proportional relationship between width and height
+  aspectRatio?: 'square' | 'cinema' | 'landscape' | number;
+  /** An experience to set while the image is fetched and loaded  */
+  loadingBehavior?: 'none' | 'blur';
```
jonathanadler, line 21: I know this isn't part of the PR itself, but the convention in the library is to use enums, so I'm wondering if it's better to align this component to the convention.
Personally I prefer using enums, cause I think it's a great tool for consumers.
In this case this would be:
```jsx
  loadingBehavior?: LoadingBehavior;
  ...
const enum LoadingBehavior {
  none = 'none',
  blur = 'blur'
}
```

src/components/Image/types.ts
```diff
@@ -0,0 +1,38 @@
...
+  /** The alternative text description of the image. Allowing better SEO when the image has meaning of content */
+  alt?: string;
+  /** A callback to be called when the image is loaded */
+  onLoad?: React.EventHandler<React.SyntheticEvent>;
+  /** A callback to be called if error occurs while loading */
+  onError?: React.EventHandler<React.SyntheticEvent>;
+  /** Specifies how the image is resized to fit its container */
+  resize?: 'contain' | 'cover';
+  // Specifies the proportional relationship between width and height
+  aspectRatio?: 'square' | 'cinema' | 'landscape' | number;
```
jonathanadler, line 19: I think in this case what I would do is just accept a number, and pass the AspectRatios as "preset".
That way, consumer could use:
```jsx
aspectRatio={AspectRatios.square}
```

Alternatively we can write the AspectRatios as an enum , and have:
```tsx
const enum AspectRatios {
  square = 1,
  cinema = 16 / 9,
  landscape = 4 / 3
}
...
aspectRatio?: AspectRatios | number
```

jonathanadler (commented):
Great work!

I've added some comments, and I have some general comments:
1. I see that you use `st(...)` in all the classNames, however it's not necessary if you're not using a stylable state, you can just pass the classname
2. I know it's not related to the PR itself, but can you add options to the `aspectRatio`, `resize` and `loadingBehavior` props for the playground page? Currently they're in the Misc. section and they need to be entered manually (this can be automatically fixed when using enums as proptypes)
3. I've noticed that when I have square dimensions (width & height equal 200px), and aspectRatio of 1, but the resize option is `contain` then the aspect ratio is "broken". Should that be the case? `cover` crops the image of course and gives me a square image, with the image itself cropped. Image attached:
image

src/components/Image/types.ts
nitayneeman: I wanted to do so but then the story page doesn't show the actual options, meaning only the enum name:
image

Also, it enforces the consumer to use the enum:
image

Not sure how comfortable it is...

src/components/Image/types.ts
nitayneeman: Same dilemma like `resize`. If you think it still better I'll change

src/components/Image/types.ts
nitayneeman: Same concerns as I mentioned above

src/components/Image/Image.spec.tsx
```diff
@@ -60,8 +55,16 @@ describe('Image', () => {
...
+    });
+
     it('should return `alt`', async () => {
-      const driver = createDriver(<Image src={src} alt={sampleAlt} />);
+      const driver = createDriver(
+        <Image src={src} {...sampleDimensions} alt={sampleAlt} />,
+      );
 
       expect(await driver.getAlt()).toBe(sampleAlt);
     });
```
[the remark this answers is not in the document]
nitayneeman: We already cover the cases by visual, the partial dimensions don't really affect the drivers (maybe only the on the `src` tests).
I can add them for extra coverage - WDYT?

nitayneeman:
Thanks for the review.

> I see that you use `st(...)` in all the classNames, however it's not necessary if you're not using a stylable state, you can just pass the classname

Your'e right, will be done.

>I know it's not related to the PR itself, but can you add options to the `aspectRatio`, `resize` and `loadingBehavior` props for the playground page? Currently they're in the Misc. section and they need to be entered manually (this can be automatically fixed when using enums as proptypes)

If we'll go with the enums I'll do it as part of this PR, otherwise I would prefer to open another separate PR...

>  I've noticed that when I have square dimensions (width & height equal 200px), and aspectRatio of 1, but the resize option is `contain` then the aspect ratio is "broken". Should that be the case? `cover` crops the image of course and gives me a square image, with the image itself cropped.

That's a good point, I also encountered this issue yesterday. The problem is that using `MediaImage` we actually "crop" the image by the calculated dimensions by the image-client-api, meaning the image is returned cropped already (depends on local pixel density and other factors). So if we use `aspectRatio` and `contain` it might lead to this effect. A solution I thought is detecting the case of aspectRatio and `contain` and not calculate the container dimensions (and pass the source dimensions instead) so that the image wouldn't be cropped by the image-client-api. WDYT? 🤔

nitayneeman:
@jonathanadler I finished with the feedbacks:

- Using `classnames` instead of `st`
- Created enums for `resize` & `loadingBehavior` whereas `aspectRatio` still remains as const. These three types are used by the propTypes, examples and tests. Also they are exported for the consumers
- Added to the playground those variables as `exampleProps` to see them under "Primary Props" and for easy simulation

Can you approve? 🙏

jonathanadler (approved):

PR #622, opened 2021-01-21
feat(wix-ui-tpa): dropdown-expose-styles #1 commit

urig-wix:
# ✨ Pull Request

### 💁 Description

Exposing styles from the dropdown component so we'd be able to override them at our project without being broken in the future  
https://jira.wixpress.com/browse/EE-28610

...

### 💭 Motivation

We're in a process of re-writing a component and migrating from wix-style-react to wix-ui-tpa.
We'd like to keep the previous visibility in some places and therefore need a way to override styles at the dropdown component.

...

### ☝️ TODOs 

...

### 👀  PR is ready for review 

- [x] Ready for review


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_622/

src/components/Dropdown/Dropdown.st.css
```diff
@@ -115,14 +241,15 @@
     stroke: none;
 }
 
-.dropdown::popoverContent {
-    border-width: 0;
+.root.overrideStyleParams .dropdown::popoverContent {
+    border-width: 1px;
+    border-color: rgba(0, 0, 0, 0.4);
```
jonathanadler, line 246: Why was the border added here?

src/components/Dropdown/Dropdown.st.css
```diff
@@ -153,6 +280,8 @@
 
 .divider {
     -st-extends: Divider;
+    opacity: 30%;
+    background-color: value(CurrentItemTextColor);
```
jonathanadler, line 284: I think you'll need to use the st-mixin approach here as well 😬 
https://github.com/wix/wix-ui-tpa/blob/master/src/components/Divider/Divider.st.css#L18

jonathanadler: Also I think the -st-extends isn't necessary here

src/components/Dropdown/Dropdown.tsx
```diff
@@ -223,7 +223,7 @@ export class Dropdown extends React.Component<DropdownProps, State> {
         : OptionFactory.create({
             ...option,
             render: () => (
-              <DropdownOption className={classes.dropdownOption} {...option} />
+              <DropdownOption className={`${classes.dropdownOption}`} {...option} />
```
jonathanadler, line 226: This seems to be redundant

src/components/Dropdown/DropdownOption.st.css
```diff
@@ -31,20 +64,38 @@
...
     -st-mixin: TPAText(
-        MainTextColor "value(colorPlaceholder)",
-        MainTextFont "{theme: 'Body-M', size: '14px', lineHeight: '20px'}"
+        MainTextColor value(CurrentItemTextDisabledColor),
+        MainTextFont value(CurrentTitleTextFont)
     );
+}
+
+.subtitle {
+    font-size: 14px !important;
```
jonathanadler, line 97: Will this still work without the `!important`?

src/components/Dropdown/docs/DropdownExtendedExample.st.css
```diff
@@ -0,0 +1,22 @@
+:import {
+    -st-from: "../Dropdown.st.css";
+    -st-default: TPADropdaown;
```
jonathanadler, line 3: this line can be removed I guess

jonathanadler (commented):
Great work!!

I've left some small comments, but overall it looks really great!

Can you send a link for Johanna to give a review?

src/components/Dropdown/Dropdown.st.css
urig-wix: because of this requirement - "Menu border should be constant: 1px color-5 with 40% opacity". have i done it wrong?

src/components/Dropdown/DropdownOption.st.css
urig-wix: does not work without the `!important` =\

jonathanadler:
Note that you need to update the bundle sizes in the `perfer.config` file
image

src/components/Dropdown/Dropdown.st.css
jonathanadler: Not at all, it's just changing the behaviour we have now.
But I guess that's ok.
But maybe this should be under:
```css
.dropdown::popoverContent
```

and not 
```css
.root.overriderStyleParams .dropdown::popoverContent
```

Meaning, this is a change that should affect the default dropdown design, right?
Not only when overriding the colors

jonathanadler (commented):
There's just one important comment I left.

Other than that looks great, and if Johanna approves then after you address my comment we can merge 💪 

Great work!

src/components/Dropdown/Dropdown.st.css
```diff
@@ -116,13 +253,17 @@
...
 .dropdown::popoverContent {
-    border-width: 0;
+    border-width: 1px;
+    border-color: rgba(0, 0, 0, 0.4);
     box-sizing: border-box;
     /* Popover fix. For some reason popover has inline style with transform: translate3d(-1px, 42px, 0px); */
     margin-left: 1px;
     max-width: 100%;
     box-shadow: 0 4px 8px 0 rgba(0, 0, 0, 0.1), 0 0 4px 0 rgba(0, 0, 0, 0.1);
-    background-color: "color(color-1)";
```
jonathanadler, line 125: Sorry for the ping-pong review 🙈  but I just noticed 2 things:
1. I think the box-shadow defined above is the border you've added. In that case I think there's no need for the extra border.
2. I see you removed the `background-color` here, is there a reason? I think it should remain here for the default styling, no?

urig-wix: will merge the box shadow & the border.
regarding the `background-color` - removed it per Johanna's request (to change the hover color)

jonathanadler: removing the background color might cause your teams apps to break 🙈 
We already had this issue a few weeks ago, where I changed the background color and it broke the product gallery page or something like that.

Also I see the background color can be overriden (a few lines below - 265), so why shouldn't we have a default background color?

Also I'm not sure I understand, when overriding we do set the

jonathanadler: The same goes the border color.
Also I think we should check with Johanna if the box shadow is necessary then.

jonathanadler: OK, so first, we need to bring back the `background-color`
I don't know why it's written like this (using quotation marks).
It can just be: 
```css
background-color: color(color-1);
``` 

Also, we need to add the border, which isn't overridable.

It should probably be something like this:
```css
border: 1px solid applyOpacity(value(CurrentItemTextColor), 0.4);
```

src/components/Dropdown/Dropdown.st.css
```diff
@@ -116,13 +282,18 @@
 }
 
 .dropdown::popoverContent {
-    border-width: 0;
+    border-width: 1px;
```
jonathanadler, line 285: No biggy, but this can be removed

jonathanadler (approved):
Cool, looks good 👍 

Let's get @johannaasseraf to approve as well 😊

jonathanadler:
Note that you need to update `perfer.config` again

PR #624, opened 2021-01-24
Image - support style wiring

nitayneeman:
# ✨ Pull Request

### 💁 Description

This PR adds overridable styles for wiring using `overrideStyleParams`:
Screen Shot 2021-01-24 at 8 58 47

BTW notice the container width - basically that happens because the wrapper inside the component, which is `<div>`. I wonder if in case of fixed width & height we also want to set them on the wrapper div. Anyway it's not related to this PR.

### 👀  PR is ready for review 

- [X] Ready for review


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_624/

src/components/Image/Image.st.css
```diff
@@ -1,10 +1,61 @@
...
+    An opacity between 0 to 100 for the image itself
+    @default 100%
+  */
+  ImageOpacity: --overridable;
+}
+
 :vars {
-  BlurRadius: 3px;
+  backgroundColor: color(value(ContainerBackgroundColor));
+  borderColor: color(value(ContainerBorderColor));
```
sivanharel, line 48: I think we should have fallback colors here as well:)

src/components/Image/Image.st.css
```diff
@@ -16,6 +67,18 @@
...
+  border: value(borderWidth) solid value(borderColor);
+  border-radius: value(borderRadius);
+  opacity: value(containerOpacity);
+}
+
+.root.overrideStyleParams .absoluteImage, 
+.root.overrideStyleParams .relativeImage {
+  opacity: value(imageOpacity);
+}
+
```
sivanharel, line 81: In all the other components we put the overrideStyleParams styling at the bottom of the st.css file. 
Do we want to keep this convention?

src/components/Image/docs/index.story.tsx
```diff
@@ -11,15 +14,17 @@ import {
...
   LoadingBehaviorOptions,
+  ResizeOptions,
 } from '../';
+import { StoryCategory } from '../../../../stories/storyHierarchy';
 import { allComponents } from '../../../../stories/utils/allComponents';
 import { settingsApi } from '../../../../stories/utils/SettingsApi';
+import { settingsPanel } from '../../../../stories/utils/SettingsPanel';
 import * as examples from './examples';
-import { StoryCategory } from '../../../../stories/storyHierarchy';
+import { ImageWiringExample } from './ImageWiringExample';
```
sivanharel, line 28: Jonathan created an OptimizedStylesBanner for all the components that use the overrideStylesParams.
Since this component was created before he updated it in the generator we should add it manually. 

`import { OptimizedStylesBanner } from '../../../../stories/OptimizedStylesBanner';`

` header({
    component: <OptimizedStylesBanner />,
}),`

src/components/Image/Image.st.css
```diff
@@ -1,10 +1,61 @@
...
+  /*
+    A border radius in pixels for the container
+    @default 0px
+  */
+  ContainerBorderRadius: --overridable;
+  /*
+    An opacity between 0 to 100 for the container
+    @default 100%
+  */
+  ContainerOpacity: --overridable;
```
sivanharel, line 38: Should the ContainerOpacity add opacity for the image as well?
If yes maybe we should change the name?

sivanharel (approved):
Nice:)
I add some minor comments:)

src/components/Image/Image.st.css
nitayneeman: According to the spec we should have both variables (one for the container, another for image)

src/components/Image/Image.st.css
nitayneeman: We can set a default color for the border (because the default width is 0px), but in case of the image it will appear by default if the image doesn't fully cover the container - not sure if we want this case?

sivanharel: So since you said that it works fine now you don't need to change anything:)

src/components/Image/Image.st.css
nitayneeman: I asked @michalvardi regarding the opacity, maybe we should use alpha color for the background of the container instead `opacity`

PR #628, opened 2021-01-25
Fix stylable warnings

jonathanadler:
### 👀  PR is ready for review 

- [x] Ready for review

### DIFF
--- a/src/components/Image/Image.st.css
+++ b/src/components/Image/Image.st.css
@@ -60,32 +60,22 @@
   min-height: 100%;
 }
 
-.root.contain .absoluteImage,
-.root.contain .relativeImage {
+.root.contain .image {
   object-fit: contain;
 }
 
-.root.cover .absoluteImage,
-.root.cover .relativeImage {
+.root.cover .image {
   object-fit: cover;
 }
 
-.preload {
+.root.preload .image {
   filter: blur(value(blurRadius));
 }
 
-.loaded {
+.root.loaded .image {
   animation-name: blurry-loading;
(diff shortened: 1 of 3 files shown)


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_628/

src/components/Image/Image.st.css
```diff
@@ -60,32 +60,22 @@
   min-height: 100%;
 }
 
-.root.contain .absoluteImage,
-.root.contain .relativeImage {
+.root.contain .image {
```
jonathanadler, line 63: Since there weren't any differences between the two classes, I thought it would be easier to use just one class

src/components/Image/Image.st.css
```diff
@@ -60,32 +60,22 @@
...
-.loaded {
+.root.loaded .image {
   animation-name: blurry-loading;
   animation-duration: 0.3s;
-  animation-timing-function: easeout;
-}
-
-.absoluteImage.preload {
-  -st-mixin: preload;
-}
```
jonathanadler, line 85: This part caused a recursive loop in stylable, since preload is using preload

src/components/Image/Image.st.css
nitayneeman: Before having RelativeMediaImage it was necessary but now I agree this is much better

nitayneeman (approved):

sivanharel (approved):

PR #632, opened 2021-01-27
Image - implement <HeroImage> & <ThumbnailImage>

nitayneeman:
# ✨ Pull Request

### 💁 Description

This PR implements two new components using the original Image - `<HeroImage>` & `<ThumbnailImage>`.

Also, it adds visual tests for wiring of the Image itself.

### 👀  PR is ready for review 

- [X] Ready for review


src/components/HeroImage/docs/index.story.tsx
```diff
@@ -0,0 +1,124 @@
...
+            text: [
+              <p style={{ marginBottom: '20px' }} key={`${dataHook}-link`}>
+                ⚠️ This component inherits the props API from{' '}
+                <LinkTo kind={'Working In Progress'} story="Image">
+                  {'<Image/>'}
+                </LinkTo>
+                , so take a look at that component for additional examples.
+              </p>,
+            ],
+          }),
```
nitayneeman, line 68: It tried to merge them into one `description` but then the problem is that we lose the "bold design" of the description, and I would like it to be consistent with the description of other components.

But anyway I think it's more readable like that actually

src/components/Image/Image.tsx
```diff
@@ -99,6 +99,11 @@ export class Image extends React.Component<ImageProps> {
             [classes.loaded]: hasLoadingBehavior && isLoaded,
           }),
         )}
+        style={{
+          // If fixed dimensions were passed, we set the calculated values to fit the container with the fixed image
+          width: width && calculatedDimensions.width,
+          height: height && calculatedDimensions.height,
+        }}
```
nitayneeman, line 106: This logic is a fix for a bug I missed out. It's relevant to the PR in order to display the components in Settings Panel correctly when passing fixed dimensions.

Now it's covered by the visuals I added

src/components/Image/Image.visual.tsx
```diff
@@ -131,6 +132,45 @@ visualize('Image', () => {
           />
         ));
       });
+
+      story('with wiring', () => {
```
nitayneeman, line 136: I really liked this approach of testing the wiring, maybe we would like to make it a convention?

src/components/Image/docs/examples.ts
```diff
@@ -6,6 +6,37 @@ const getKeyByValue = (object, value) =>
...
+  resize: Object.keys(ResizeOptions),
+  aspectRatio: [
+    ...Object.keys(AspectRatioPresets),
+    { label: 'example of custom number (2.33)', value: 2.33 },
+    { label: 'none', value: null },
+  ],
+  loadingBehavior: Object.keys(LoadingBehaviorOptions),
+};
+
+export const commonWiringNumberParams = [
```
nitayneeman, line 19: Just to reuse between the components and place the logic once

WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_632/

src/components/HeroImage/HeroImage.st.css
```diff
@@ -0,0 +1,45 @@
...
+  ImageOpacity: --overridable;
+}
+
+:vars {
+  backgroundColor: color(fallback(value(BackgroundColor), value(defaultBackgroundColor)));
+  imageOpacity: fallback(value(ImageOpacity), value(defaultImageOpacity));
+}
+
+.root {
+  -st-extends: Image;
```
sivanharel, line 37: Since you are not extending this Image and you only use the -st-mixin, you don't need the st-extends here.
Moreover, when adding this st-extends you give the ability for our users to target inner classes from the Image component and to override the styling.

For example:
.myHeroImage::anyClassFromTheImageComponent {
   // some styling
}

src/components/Image/Image.visual.tsx
sivanharel: Yes, I also think that we should add this kind of tests for all the components :)

src/components/Image/docs/index.story.tsx
```diff
@@ -72,6 +66,27 @@ export default {
...
+                using{' '}
+                <LinkTo kind={'Working In Progress'} story="HeroImage">
+                  {'<HeroImage/>'}
+                </LinkTo>{' '}
+                and{' '}
+                <LinkTo kind={'Working In Progress'} story="ThumbnailImage">
+                  {'<ThumbnailImage/>'}
+                </LinkTo>{' '}
+                instead, which are meant for common TPAs use-cases.
+              </p>,
```
sivanharel, line 86: If we want that our users will use only the HeroImage and the ThumbnailImage so maybe we shouldn't expose the Image component. I mean maybe we could remove this story page.
WDYT?

src/components/ThumbnailImage/ThumbnailImage.st.css
```diff
@@ -0,0 +1,67 @@
...
+:vars {
+  backgroundColor: color(fallback(value(BackgroundColor), value(defaultBackgroundColor)));
+  borderColor: color(fallback(value(BorderColor), value(defaultBorderColor)));
+  borderWidth: fallback(value(BorderWidth), value(defaultBorderWidth));
+  borderRadius: fallback(value(BorderRadius), value(defaultBorderRadius));
+  imageOpacity: fallback(value(ImageOpacity), value(defaultImageOpacity));
+}
+
+.root {
+  -st-extends: Image;
```
sivanharel, line 56: Here also. I think that we can remove the `-st-extends: Image;`

src/components/ThumbnailImage/docs/ThumbnailImageWiringExample.st.css
```diff
@@ -0,0 +1,15 @@
+:import {
+  -st-from: '../ThumbnailImage.st.css';
+  -st-default: ThumbnailImage;
```
sivanharel, line 3: You can remove the `-st-default: ThumbnailImage;` here :)

sivanharel (commented):
Very nice!!!

src/components/Image/docs/index.story.tsx
nitayneeman: We thought about this concern if to make it internal, but at the moment we don't really know. In some cases they might want to use the image for different specific use cases so we decided to open it in the meantime

sivanharel (approved):

nitayneeman:
@jonathanadler are you planning to go over it or may I merge? 🥴

jonathanadler:
> @jonathanadler are you planning to go over it or may I merge? 🥴

Sorry, didn't get a chance to go over it, but you can merge 💪

PR #638, opened 2021-02-01
Radio Button - box border should change from 60% to 100% both on hover and focus

urig-wix:
# ✨ Pull Request

### 💁 Description
The box border should change from 60% to 100% both on hover and focus.
See comment on [EE-28048](https://jira.wixpress.com/browse/EE-28048?focusedCommentId=8601528&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel)

...

### 💭 Motivation

Would like to migrate to radio button at our project.
...

### ☝️ TODOs 

...

### 👀  PR is ready for review 

- [x] Ready for review

### DIFF
--- a/src/components/RadioButton/RadioButton.st.css
+++ b/src/components/RadioButton/RadioButton.st.css
@@ -198,16 +198,9 @@
 
 /* Create the indicator (the dot/circle - hidden when not checked) */
 
-.root:box:hover:checked .wrapper {
-    border-color: applyOpacity(value(CurrentIconColor), 1);
-}
 
 .root:box:hover:not(:checked) .wrapper {
     border-color: applyOpacity(value(CurrentBorderColor), 1);
-}
-
-.root:box:focus-within .wrapper {
-    border-color: applyOpacity(value(CurrentIconColor), 1);
 }
 
 .root:hover .radioIcon {
--- a/src/components/RadioButton/RadioButton.st.css
+++ b/src/components/RadioButton/RadioButton.st.css
@@ -198,7 +198,23 @@
 
 /* Create the indicator (the dot/circle - hidden when not checked) */
 
+.root:box:hover:checked .wrapper {
+    border-color: applyOpacity(value(CurrentIconColor), 1);
+}
+
+.root:box:hover:not(:checked) .wrapper {
+    border-color: applyOpacity(value(CurrentBorderColor), 1);
+}
+
+.root:box:focus-within .wrapper {
+    border-color: applyOpacity(value(CurrentIconColor), 1);
+}
+
 .root:hover .radioIcon {
+    border-color: applyOpacity(value(CurrentBorderColor), 1);
+}
+
+.root:focus-within .radioIcon {
     border-color: applyOpacity(value(CurrentBorderColor), 1);
 }
 
--- a/src/components/RadioButton/RadioButton.st.css
+++ b/src/components/RadioButton/RadioButton.st.css
@@ -201,14 +201,17 @@
 
 .root:box:hover:not(:checked) .wrapper {
     border-color: applyOpacity(value(CurrentBorderColor), 1);
+    transition: border-color .2s ease-in-out;
 }
 
 .root:hover .radioIcon {
     border-color: applyOpacity(value(CurrentBorderColor), 1);
+    transition: border-color .2s ease-in-out;
 }
 
 .root:focus-within .radioIcon {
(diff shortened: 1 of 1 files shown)


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_638/

src/components/RadioButton/RadioButton.st.css
```diff
@@ -198,10 +198,26 @@
 
 /* Create the indicator (the dot/circle - hidden when not checked) */
 
+.root:box:hover:checked .wrapper {
+    border-color: applyOpacity(value(CurrentIconColor), 1);
+}
```
jonathanadler, line 203: I'm not sure this is needed, I think that it's always defined without opacity when it's `checked`

src/components/RadioButton/RadioButton.st.css
```diff
@@ -198,10 +198,26 @@
...
+
+.root:box:hover:not(:checked) .wrapper {
+    border-color: applyOpacity(value(CurrentBorderColor), 1);
+}
```
jonathanadler, line 207: And then maybe we don't need the `:not(:checked)` her.
Maybe

src/components/RadioButton/RadioButton.st.css
```diff
@@ -198,10 +198,26 @@
...
+
+.root:box:focus-within .wrapper {
+    border-color: applyOpacity(value(CurrentIconColor), 1);
```
jonathanadler, line 210: should this be:
```suggestion
    border-color: applyOpacity(value(CurrentBorderColor), 1);
```
?

jonathanadler (commented):
Looks good as always 👍 
left some comments and questions.

Also unfortunately we don't have a good way to test this at the moment :( .

src/components/RadioButton/RadioButton.st.css
urig-wix: without the :not(:checked), the color of the border would change when hovering (gets 'CurrentIconColor' instead of 'CurrentBorderColor')

src/components/RadioButton/RadioButton.st.css
urig-wix: deleted ^ as its included in 
.root:hover .radioIcon {
    border-color: applyOpacity(value(CurrentBorderColor), 1);
}
(and the box does not have opacity at all when it's checked)

jonathanadler (approved):
Awesome 😎

Can you send the storybook link for johanna to review?

I'm approving, but we'll just wait for her approval before merging

urig-wix:
> Awesome 😎
> 
> Can you send the storybook link for johanna to review?
> 
> I'm approving, but we'll just wait for her approval before merging

Sorry for the newbies question but how do I send a link of the branches storybook ?

jonathanadler:
> > Awesome 😎
> > Can you send the storybook link for johanna to review?
> > I'm approving, but we'll just wait for her approval before merging
> 
> @jonathanadler Sorry for the newbies question but how do I send a link of the branches storybook ?

No need to apologise 😊 

You can find it if you scroll up a bit
image

src/components/RadioButton/RadioButton.st.css
```diff
@@ -198,8 +198,20 @@
...
+
 .root:hover .radioIcon {
     border-color: applyOpacity(value(CurrentBorderColor), 1);
+    transition: border-color .2s ease-in-out;
+}
+
+.root:focus-within .radioIcon {
+    border-color: applyOpacity(value(CurrentBorderColor), 1);
+    transition: border-color .2s ease-in-out;
 }
```
jonathanadler, line 215: Note that this will add the transition only on hover, but not when removing the cursor/focus from the element.
Is that ok?

urig-wix: that's a good question. 
should I ask Johanna? what would you do different?

jonathanadler: Well, if we add it directly to the `radioIcon`, like this:
```css
.radioIcon {
  transition: border-color .2s ease-in-out;
}
```
It will affect both the enter and the leave.
However, this will also affect the border in all the radio-button's states: when the radio-button is being checked, disabled, etc.
so I guess it's best to leave it like this.
I would check with Johanna in any case, though I'm not sure achieving this is worth the effort.
(and I raised the issue in the first place!)

PR #646, opened 2021-02-03
<Text /> - add priority prop

jjaskulis:
# ✨ Pull Request

### 💁 Description
	Adding ability to specify a lower hierarchy color for <Text/> component;
	https://jira.wixpress.com/browse/DSM-1186

...

### 💭 Motivation

...

### ☝️ TODOs 

...

### 👀  PR is ready for review 

- [x] Ready for review


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_646/

src/components/Text/Text.st.css
```diff
@@ -23,6 +24,11 @@
       @default color-5
     */
     MainTextColor: --overridable;
+    /*
+      The color to use instead of default color for secondary text
+      @default color-4
+    */
+    SecondaryTextColor: --overridable;
```
sivanharel, line 31: We don't need 2 different vars here, we could just use the `MainTextColor` to override the primary and secondary priority.

So you could just update the comments: 
/*
      The color of the text instead of the default color **set by priority**
      @default for primary priority: color-5,  for secondary priority: color-4
*/
 MainTextColor: --overridable;
 
 and use it as follow:

 `
.root:priority(primary){
    color: color(fallback(value(MainTextColor), value(DefaultTextColor)));
}

.root:priority(secondary){
    color: color(fallback(value(MainTextColor), value(DefaultSecondaryTextColor)));
}`

src/components/Text/Text.tsx
```diff
@@ -1,14 +1,24 @@
 import * as React from 'react';
 import { st, classes } from './Text.st.css';
-import { DEFAULT_TAG_NAME, TYPOGRAPHY } from './constants';
+import { DEFAULT_TAG_NAME, TYPOGRAPHY, TEXT_PRIORITY } from './constants';
```
sivanharel, line 3: There is no need for the `Text` prefix since it is known that it is a priority for the text.
Let's change it for PRIORITY.

sivanharel (commented):
Very Nice work!
I wrote some comments.

Let's add the secondary priority example  for the settings panel tab as well:
Screen Shot 2021-02-03 at 14 32 02

In order to do it, you would add another instance of the component with `priority=PRIORITY.secondary` here:
https://github.com/wix/wix-ui-tpa/blob/86f0937284ad76ce140c8cdc27a30e6d7e9e9556/src/components/Text/docs/TextWiringExample.tsx#L6

<>
    <Text className={classes.root}>Primary Priority</Text>
    <br />
    <Text className={classes.root} priority=PRIORITY.secondary>Secondary Priority</Text>
</>

sivanharel:
In addition, the Prefer tests fail since the bundle of the Text component was increase by +0.103kb.
Since it is not a lot :) we just need to add +1 for the sizes of the following components:
1. `CalendarPopover`: https://github.com/wix/wix-ui-tpa/blob/86f0937284ad76ce140c8cdc27a30e6d7e9e9556/perfer.config.js#L17
2. `Dropdown`: https://github.com/wix/wix-ui-tpa/blob/86f0937284ad76ce140c8cdc27a30e6d7e9e9556/perfer.config.js#L29
3. `Event`: https://github.com/wix/wix-ui-tpa/blob/86f0937284ad76ce140c8cdc27a30e6d7e9e9556/perfer.config.js#L30
4. `FloatingDropdown`: https://github.com/wix/wix-ui-tpa/blob/86f0937284ad76ce140c8cdc27a30e6d7e9e9556/perfer.config.js#L31

src/components/Text/Text.tsx
jjaskulis: there is already a PRIORITY enum used in <Button /> component, so I had to name this TEXT_PRIORITY to avoid collision as they're both exported in components/index.ts

sivanharel: Nice catch! Ok so keep this name:)

src/components/Text/Text.st.css
jjaskulis: updated

jjaskulis:
> In addition, the Prefer tests fail since the bundle of the Text component was increase by +0.103kb.
> Since it is not a lot :) we just need to add +1 for the sizes of the following components:
> 
> 1. `CalendarPopover`: https://github.com/wix/wix-ui-tpa/blob/86f0937284ad76ce140c8cdc27a30e6d7e9e9556/perfer.config.js#L17
> 2. `Dropdown`: https://github.com/wix/wix-ui-tpa/blob/86f0937284ad76ce140c8cdc27a30e6d7e9e9556/perfer.config.js#L29
> 3. `Event`: https://github.com/wix/wix-ui-tpa/blob/86f0937284ad76ce140c8cdc27a30e6d7e9e9556/perfer.config.js#L30
> 4. `FloatingDropdown`: https://github.com/wix/wix-ui-tpa/blob/86f0937284ad76ce140c8cdc27a30e6d7e9e9556/perfer.config.js#L31

sivanharel (approved):
nice!!

PR #655, opened 2021-02-04
<ActionsMenuLayout /> - fix focus handling

jjaskulis:
# ✨ Pull Request

### 💁 Description
	Fixes https://jira.wixpress.com/browse/DSM-1138
	
	Currently component always focuses the first element or the one provided with focusedIndex prop on initial load, which is problematic in some use-cases.
Update behaviour will only autofocus if focusedIndex prop is provided. Manual focus is affected by this change - tab focus, arrow key navigation and focus on click are not affected.
Also fixed bug where clicking on list item would not always update internal state, so trying to use arrow keys would focus on the wrong list item.

...

### 💭 Motivation

...

### ☝️ TODOs 

...

### 👀  PR is ready for review 

- [x] Ready for review

### DIFF
--- a/src/components/ActionsMenuLayout/ActionsMenuLayout.driver.ts
+++ b/src/components/ActionsMenuLayout/ActionsMenuLayout.driver.ts
@@ -13,6 +13,7 @@ export interface ActionsMenuLayoutDriver extends BaseUniDriver {
   clickItem(content): Promise<any>;
   item(content): UniDriver<HTMLLIElement>;
   layout(): UniDriver<HTMLUListElement>;
+  isItemFocused(content): Promise<boolean>;
 }
 
 function actionsMenuLayoutItem(base: UniDriver, content) {
@@ -30,6 +31,9 @@ export const actionsMenuLayoutDriverFactory = (
       (await base.attr(ACTIONS_MENU_DATA_KEYS.mobile)) === 'true',
     item: (content) => actionsMenuLayoutItem(base, content),
     clickItem: (content) => actionsMenuLayoutItem(base, content).click(),
+    isItemFocused: async (content) =>
+      (await actionsMenuLayoutItem(base, content).getNative()) ===
+      document.activeElement,
     layout: () => base,
   };
 };
--- a/src/components/ActionsMenuLayout/ActionsMenuLayout.driver.ts
+++ b/src/components/ActionsMenuLayout/ActionsMenuLayout.driver.ts
@@ -13,7 +13,6 @@ export interface ActionsMenuLayoutDriver extends BaseUniDriver {
   clickItem(content): Promise<any>;
   item(content): UniDriver<HTMLLIElement>;
   layout(): UniDriver<HTMLUListElement>;
-  isItemFocused(content): Promise<boolean>;
 }
 
 function actionsMenuLayoutItem(base: UniDriver, content) {
@@ -31,9 +30,6 @@ export const actionsMenuLayoutDriverFactory = (
       (await base.attr(ACTIONS_MENU_DATA_KEYS.mobile)) === 'true',
     item: (content) => actionsMenuLayoutItem(base, content),
     clickItem: (content) => actionsMenuLayoutItem(base, content).click(),
-    isItemFocused: async (content) =>
-      (await actionsMenuLayoutItem(base, content).getNative()) ===
-      document.activeElement,
     layout: () => base,
   };
 };
--- a/src/components/ActionsMenuLayout/ActionsMenuLayout.tsx
+++ b/src/components/ActionsMenuLayout/ActionsMenuLayout.tsx
@@ -9,6 +9,10 @@ import { KEYS } from '../../common/keyCodes';
 
 export interface ActionsMenuLayoutProps extends TPAComponentProps {
   alignment?: Alignment;
+  /**
+   * Provide item index which will be focused on initial render.
+   * If prop is not provided, no item will be autofocused
+   */
   focusedIndex?: number;
   /** a11y */
   'aria-labeledby'?: string;
@@ -23,7 +27,7 @@ export class ActionsMenuLayout extends React.Component<ActionsMenuLayoutProps> {
   static displayName = 'ActionsMenuLayout';
 
   state = {
-    focusedIdx: 0,
+    focusedIdx: this.props.focusedIndex ?? 0,
   };
 
   liRefs: {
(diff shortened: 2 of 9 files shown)


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_655/

stories/index.ts
```diff
@@ -3,6 +3,7 @@ require('./Playground/Playground.story');
 require('./Icons.story');
 
 require('../src/components/ActionsMenuLayout/docs/index.story');
+require('../src/components/ActionsMenuLayout/docs/ActionsMenuLayoutTestStory');
```
sivanharel, line 6: This should be added on https://github.com/wix/wix-ui-tpa/blob/master/stories/testStories.ts instead of here.

sivanharel (commented):
Wow very nice!!

I think that we need to remove the focus onKeyDown event when the key is 'enter' or 'esc'.
@jonathanadler - What do you think?

src/components/ActionsMenuLayout/ActionsMenuLayout.driver.ts
```diff
@@ -30,6 +31,9 @@ export const actionsMenuLayoutDriverFactory = (
       (await base.attr(ACTIONS_MENU_DATA_KEYS.mobile)) === 'true',
     item: (content) => actionsMenuLayoutItem(base, content),
     clickItem: (content) => actionsMenuLayoutItem(base, content).click(),
+    isItemFocused: async (content) =>
+      (await actionsMenuLayoutItem(base, content).getNative()) ===
+      document.activeElement,
```
jonathanadler, line 36: Note that this won't work in e2e tests, since the `document` doesn't exist.
Unfortunately this type of logic needs to be moved to the test itself.

src/components/ActionsMenuLayout/ActionsMenuLayout.tsx
```diff
@@ -75,6 +100,16 @@ export class ActionsMenuLayout extends React.Component<ActionsMenuLayoutProps> {
...
+  _onListItemClick = (e: MouseEvent) => {
+    const indexValue = Object.values(this.liRefs).findIndex(
+      (i) => i.liRef?.current === e.currentTarget,
+    );
+
+    if (indexValue >= 0) {
+      this._focusItem(indexValue);
+    }
+  };
+
```
jonathanadler, line 112: Theoretically the `ActionsMenu` component is a list of action items, so that each one of the list item is clickable, and sends an action. 
After that, the `ActionsMenu` will usually disappear (it should be inside of a popover).
So I'm not sure we should control the focus.
I think the focus should work the same as hover, meaning they should probably share the same state.

jonathanadler (commented):
Left some comments, but really nice work

stories/index.ts
jjaskulis: oops, updated

src/components/ActionsMenuLayout/ActionsMenuLayout.driver.ts
jjaskulis: ah, i guess thats why it was like this in the first place. will update

src/components/ActionsMenuLayout/ActionsMenuLayout.tsx
jjaskulis: @jonathanadler this is fixing the behaviour that was already present. I agree on your point. In our product, we use this component as you described, and item focus state in this scenario doesn't make sense. But I'm not aware of all the use-cases, perhaps the element is used in some scenarios where it's needed. Let me know how to proceed, I'll remove this functionality if it's the right way for this component.

stories/index.ts
jjaskulis: > Wow very nice!!
> 
> I think that we need to remove the focus onKeyDown event when the key is 'enter' or 'esc'.
> @jonathanadler - What do you think?

imho esc should remove focus and enter should work same as click

src/components/ActionsMenuLayout/ActionsMenuLayout.tsx
jonathanadler: 🤔  I understand what you're saying.
Although I think it's best to avoid adding this behaviour at the moment.
We can maybe try to add it in a different PR.

In addition, we're also planning on developing an `ActionMenu` component, which will use this `ActionsMenuLayout` component as its content.
It will also already contain a popover (for desktop) and a mobile drawer (for mobile of course).
The `ActionMenu` will handle all the focus and a11y issues, so it will be much easier for consumers to use the component.

src/components/ActionsMenuLayout/ActionsMenuLayout.tsx
```diff
@@ -87,42 +89,35 @@ export class ActionsMenuLayout extends React.Component<ActionsMenuLayoutProps> {
...
   render() {
     const { alignment, children, className } = this.props;
     return (
       <TPAComponentsConsumer>
         {({ mobile }) => (
           <ul
             onFocus={this._onFocus}
-            onClick={this._onClick}
+            onMouseDown={this._preventFocus}
```
jonathanadler, line 112: Is this still needed here?

src/components/ActionsMenuLayout/ActionsMenuLayout.tsx
```diff
@@ -87,42 +89,35 @@ export class ActionsMenuLayout extends React.Component<ActionsMenuLayoutProps> {
...
             onKeyDown={this._onKeyDown}
             className={st(classes.root, { mobile }, className)}
             {...this.getDataAttributes(mobile)}
             data-hook={this.props['data-hook']}
             role="menu"
             aria-labelledby={this.props['aria-labeledby']}
             aria-label={this.props['aria-label']}
-            tabIndex={-1}
+            tabIndex={0}
```
jonathanadler, line 120: Also, I'm not sure the tabindex here is needed 🤔

jonathanadler (commented):
Just two tiny comments
But looks great in general!

src/components/ActionsMenuLayout/ActionsMenuLayout.tsx
jjaskulis: @jonathanadler  depends, these lines you commented on are related. 
so now you can tab-focus the actions menu component and the first item in the list will focused immediately. to achieve this needed to make container tab-able (with tabIndex), but then we need to prevent the focus event from firing, as they will be fired when clicking on list item or the action menu container.

now, if the way forward is to have `ActionMenu` component which will control the focus, ability to tab into this component is not actually needed. what do you think?

jonathanadler: oh, I see cool 👍

jonathanadler (approved):

jonathanadler:
Great job @jjaskulis!!

PR #651, opened 2021-02-04
Image - implement hover effects

nitayneeman:
# ✨ Pull Request

### 💁 Description

This PR adds a new prop called `hoverEffect` supporting two vales (`zoom` & `darken`) according to the spec.

### 👀  PR is ready for review 

- [X] Ready for review

PR #656, opened 2021-02-04
Image - fix flaky tests

nitayneeman:
# ✨ Pull Request

### 💁 Description

This PR should fix the flaky tests in Image.

Those are the problematic tests:
- Image/Relative Media URI/with wiring: override background color
- Image/Absolute URL/with wiring: override background color
- Image/Relative Media URI: with onError
- Image/Relative Media URI: with onError

The common for all of them is the usage of `invalidSrc` which triggers the native broken image icon of the browser. The thing is ui-core's Image [replaces](https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/image/image.tsx#L113-L116) the source in such cases with an image with dimensions of 1x1 pixels, in order to hide the native icon. Probably sometimes Applitools is fast enough to take the screenshot before that the icon is replaced (I also managed to notice it when running the visuals locally).

I believe the 500ms with `await` should fix the problem because `onError` is invoked in ui-core's Image after setting the state (inside ui-core's Image). In case of the "override background color" I changed it to a non-broken source and simply hide the image to show the wired background.

### 👀  PR is ready for review 

- [X] Ready for review

### DIFF
--- a/src/components/Image/Image.visual.tsx
+++ b/src/components/Image/Image.visual.tsx
@@ -31,8 +31,10 @@ class ImageWithWrapper extends React.Component<ImageWithWrapperProps> {
   state = { hasError: false };
 
   async onError(onError) {
-    this.setState({ hasError: true }, () => () => onError());
-    await delay(500);
+    this.setState({ hasError: true }, async () => {
+      await delay(500);
+      onError();
+    });
   }
 
   render() {
--- a/src/components/Image/Image.visual.tsx
+++ b/src/components/Image/Image.visual.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import { delay } from '../../test/utils';
 import { snap, story, visualize } from 'storybook-snapper';
 import { Image, ImageProps } from './';
 import { classes } from './Image.visual.st.css';
@@ -29,8 +30,9 @@ type ImageWithWrapperProps = ImageProps & {
 class ImageWithWrapper extends React.Component<ImageWithWrapperProps> {
   state = { hasError: false };
 
-  onError(onError) {
-    this.setState({ hasError: true }, () => setTimeout(() => onError(), 500));
+  async onError(onError) {
+    this.setState({ hasError: true }, () => () => onError());
+    await delay(500);
   }
 
   render() {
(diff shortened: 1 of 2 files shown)


WixBuildServer:
### 🤖 Hi, I'm #Janet!

  * https://bo.wix.com/pages/wix-ui-tpa.pr_656/

nitayneeman:
Both visuals of "override background color" should break because now we're not using the `invalidSrc` so we don't have the red border anymore.

src/components/Image/Image.visual.tsx
```diff
@@ -29,8 +30,11 @@ type ImageWithWrapperProps = ImageProps & {
 class ImageWithWrapper extends React.Component<ImageWithWrapperProps> {
   state = { hasError: false };
 
-  onError(onError) {
-    this.setState({ hasError: true }, () => setTimeout(() => onError(), 500));
+  async onError(onError) {
+    this.setState({ hasError: true }, async () => {
+      await delay(500);
+      onError();
+    });
```
jonathanadler, line 37: Isn't it problematic that `onError` is defined twice?
```suggestion
  async onError(callback) {
    this.setState({ hasError: true }, async () => {
      await delay(500);
      callback();
    });
```

src/components/Image/Image.visual.tsx
```diff
@@ -164,7 +168,7 @@ visualize('Image', () => {
       story('with wiring', () => {
         snap('override background color', (done) => (
           <ImageWithWrapper
-            src={invalidSrc}
+            src={src}
```
jonathanadler, line 171: oh I thought the invalidSrc was to test the error state, no?

nitayneeman: Yep it is and it's still being used for the error state. But I also used it to test the background color wiring.

src/components/Image/Image.visual.tsx
nitayneeman: Oops! I did the fix in a different PR instead 🤦🏻‍♂️:
https://github.com/wix/wix-ui-tpa/pull/651/files#diff-59cb9bbb8734f702d7df647b858b4173e292160bcc9615bbb390a3eecd5a7aabR32

I just pushed a fix to this branch.

jonathanadler (approved):

